PR #1, opened 2020-05-01
Unittest

mdbooth:
An experimental unit test for CreateOSPHostsEntries, and a couple of small quality-of-life improvements. To run:

go test ./pkg/util


.gitignore
```diff
@@ -0,0 +1,2 @@
+/go.sum
```
mrKisaoLamb, line 1: We shoudn't ignore go.sum https://github.com/golang/go/wiki/Modules#releasing-modules-all-versions

mdbooth: Good spot, thanks. Had assumed generated == don't commit, but it seems not.

We'll need to be careful to update and commit this along with the changes which caused it. Some kind of automation would be helpful, or I suspect this will be frequently overlooked.

mrKisaoLamb: I found something like this https://github.com/renovatebot/renovate or we can try to do something with git-hooks and 'go mod tidy'

pkg/util/hosts.go
```diff
@@ -1,52 +1,40 @@
...
-        "strings"
+	"fmt"
+	"regexp"
+	"strings"
 
-        corev1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 )
 
+var ansibleBlockRE = regexp.MustCompile(`(?s).*BEGIN ANSIBLE MANAGED BLOCK\n(.*)# END ANSIBLE MANAGED BLOCK.*`)
```
bogdando, line 11: I wonder if this can be a constant as well

stuggi: in the end this is going to be removed as we get the input from the controller deploy ansible vars where we don't have to use regex. This was only initially required during the POC when adding a generated hosts file as input from an existing external control plane.

mpryc:
@mdbooth could your rebase this change? it should work as CI'ed in travis already

PR #5, opened 2020-05-06
Add a pre-push git hook to run travis tests locally

mdbooth:
To enable:
  $ git config core.hooksPath .githooks

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+# Run all ci scripts locally before pushing. To enable this hook, run:
+# $ git config core.hooksPath .githooks
+
+BASE_DIR="$(dirname $0)"
+tests="${BASE_DIR}"/../ci_scripts/travis.*.sh
+for t in $tests; do
+    $t
+done


.githooks/pre-push
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+# Run all ci scripts locally before pushing. To enable this hook, run:
+# $ git config core.hooksPath .githooks
+
+BASE_DIR="$(dirname $0)"
+tests="${BASE_DIR}"/../ci_scripts/travis.*.sh
```
mpryc, line 7: If we decide to have common repo to store all the test scripts then one additional step to clone that repo will be required here.

mrKisaoLamb:
I like idea of runing test all the time **but** I'm afraid that it will bite us hard at least in other repos. Whatif  we will want to push something in WIP state? What if we will want to push code on the end of the day and we don't have to much time etc? Also what in case where we will have more tests, definitly we don't want wait for all of them with all pushe especially for 'nit' changes and cosmetic one.

mpryc:
@mrKisaoLamb I think when the need comes we will find solution :)

For now you can always add [ci skip] to the commit topic and Travis will skip tests.

Also it's possible to have branches / tags / multiple imports for the configuration pipeline and having that we can decide what to run on which repos or what tests to run on specific file changes. e.g. changing README will skip all tests automagically.

mpryc:
Ah the [ci skip] will not work as it's pre-hook. Misunderstood. You have option to remove that hook using git. Also I think  it's not stored in the repo.

PR #10, opened 2020-09-24
Add method to create liveness/readiness probes

stuggi:


pkg/util/probe.go
```diff
@@ -0,0 +1,104 @@
...
+	defaultLivenessPeriodSeconds       int32  = 60
+	defaultLivenessTimeoutSeconds      int32  = 3
+	defaultLivenessFailureThreshold    int32  = 5
+	defaultCommand                     string = "/openstack/healthcheck"
+)
+
+// Probe details
+type Probe struct {
+	// ProbeType, either readiness, or liveness
+	ProbeType           string
```
mdbooth, line 48: Define a ProbeType type here and use that instead of string. Define "readiness" and "liveness" as constants of this type.

stuggi: donw

pkg/util/probe.go
```diff
@@ -0,0 +1,111 @@
...
+	defaultLivenessFailureThreshold    int32  = 5
+	defaultCommand                     string = "/openstack/healthcheck"
+)
+
+const (
+	readiness string = "readiness"
+	liveness  string = "liveness"
+)
+
+type probeType string
```
mdbooth, line 50: type ProbeType string

pkg/util/probe.go
```diff
@@ -0,0 +1,111 @@
...
+
+// Probe details
+type Probe struct {
+	// ProbeType, either readiness, or liveness
+	ProbeType           probeType
```
mdbooth, line 55: ProbeType ProbeType

pkg/util/probe.go
```diff
@@ -0,0 +1,111 @@
...
+	// default Liveness values
+	defaultLivenessInitialDelaySeconds int32  = 30
+	defaultLivenessPeriodSeconds       int32  = 60
+	defaultLivenessTimeoutSeconds      int32  = 3
+	defaultLivenessFailureThreshold    int32  = 5
+	defaultCommand                     string = "/openstack/healthcheck"
+)
+
+const (
+	readiness string = "readiness"
```
mdbooth, line 46: ProbeTypeReadiness ProbeType = "readiness"

pkg/util/probe.go
```diff
@@ -0,0 +1,111 @@
...
+	liveness  string = "liveness"
```
mdbooth, line 47: ProbeTypeLiveness ProbeType = "liveness"

pkg/util/probe.go
```diff
@@ -0,0 +1,111 @@
...
+	PeriodSeconds       int32 // min value 1
+	TimeoutSeconds      int32 // min value 1
+	FailureThreshold    int32 // min value 1
+}
+
+// GetProbe -
+func (p *Probe) GetProbe() *corev1.Probe {
+
+	switch p.ProbeType {
+	case "readiness":
```
mdbooth, line 67: case ProbeTypeReadiness:

pkg/util/probe.go
```diff
@@ -0,0 +1,111 @@
...
+		if p.PeriodSeconds == 0 {
+			p.PeriodSeconds = defaultReadinessPeriodSeconds
+		}
+		if p.TimeoutSeconds == 0 {
+			p.TimeoutSeconds = defaultReadinessTimeoutSeconds
+		}
+		if p.FailureThreshold == 0 {
+			p.FailureThreshold = defaultReadinessFailureThreshold
+		}
+	case "liveness":
```
mdbooth, line 80: case ProbeTypeLiveness:

pkg/util/probe.go
```diff
@@ -0,0 +1,111 @@
...
+	ProbeType           probeType
+	Command             string
+	InitialDelaySeconds int32 // min value 1
+	PeriodSeconds       int32 // min value 1
+	TimeoutSeconds      int32 // min value 1
+	FailureThreshold    int32 // min value 1
+}
+
+// GetProbe -
+func (p *Probe) GetProbe() *corev1.Probe {
```
mdbooth, line 64: If you changed the signature to:

`func (p *Probe) ApplyProbe(probe *corev1.Probe) {`

and the modified the passed-in probe rather than creating a new one, you would also be able to use this util function in apply methods such as CreateOrUpdate()

stuggi: yes that sounds good. I'll add an ApplyProbe this in a follow up.

PR #15, opened 2022-06-09
Create common functions to create typical resources used by operators

stuggi:
Adds functions to make it easy to create database (using mariadb-operator api),
deployment, statefulset, service, routes, conditions and other tools like a generic
helper which holds information of a current instance during reconciliation.


go.mod
```diff
@@ -5,12 +5,33 @@ go 1.17
 require (
 	github.com/go-logr/logr v1.2.0
 	github.com/onsi/gomega v1.17.0
-	golang.org/x/crypto v0.0.0-20210921155107-089bfa567519
+	github.com/openshift/api v3.9.0+incompatible
+	github.com/openstack-k8s-operators/keystone-operator v0.0.0-20220516151700-a6967a7f8a1b
+	github.com/openstack-k8s-operators/mariadb-operator v0.0.0-20220516121356-119f8d825a71
```
stuggi, line 10: right now used in unit tests, should be changed that we do not have a dependency to operators

pkg/database/funcs_test.go
```diff
@@ -0,0 +1,163 @@
...
+			t.Fatalf("NewHelper error: (%v)", err)
+		}
+
+		// createDB
+		_, _, err = d.CreateOrPatchDB(
+			context.TODO(),
+			h,
+		)
+		if err != nil {
+			t.Fatalf("CreateOrPatchDB error: (%v)", err)
```
stuggi, line 110: this unit tests fails right now with 
    funcs_test.go:110: CreateOrPatchDB error: (Error listing services for labels: map[app:mariadb] - Get "http://localhost/api/v1/namespaces/openstack/services?labelSelector=app%3Dmariadb": dial tcp [::1]:80: connect: connection refused)

need to check on that

stuggi: likely need to use envtest for this

go.mod
dprince: ugh, yeah this is really bad I think. A cyclical dependency shouldn't happen here

stuggi: I removed the keystone dependency. as discussed on slack we might move the database package to the mariadb operator in a follow up

pkg/database/funcs_test.go
stuggi: disabled the tests for now

PR #17, opened 2022-06-24
Add shared const used by operators

stuggi:
Moves AppSelector, InputHashName and DebugCommend to shared const
in lib-common to be used by all operators, instead of specify it
in the operator itself.


abays (approved):
/lgtm

olliewalsh (approved):

PR #19, opened 2022-06-28
Update deployment + adds GetDataFromSecret and update GetSecret/s to use helper

stuggi:
* Add method to get deployment object
* Add GetDataFromSecret and update GetSecret/s to use helper

### DIFF
--- a/pkg/common/deployment.go
+++ b/pkg/common/deployment.go
@@ -25,6 +25,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


pkg/common/deployment.go
```diff
@@ -97,3 +104,28 @@ func (d *Deployment) Delete(
...
+	namespace string,
+) (*appsv1.Deployment, error) {
+
+	depl := &appsv1.Deployment{}
+	err := h.GetClient().Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, depl)
+	if err != nil {
+		if k8s_errors.IsNotFound(err) {
+			return depl, err
+		}
+		return depl, err
```
abays, line 127: Do we need to check for `IsNotFound` here if both lines 125 and 127 end up returning the same thing?

stuggi: no, will update. thx

abays (approved):
/lgtm

PR #22, opened 2022-07-06
Add openstack module

stuggi:
Adds openstack module with functions to manage
- client
- endpoints
- project
- role
- service
- user

E.g. used in KeystoneService to register users for service
components.

Usage is to use NewOpenStack() to get a new instance of OpenStack
with an authenticated client. Then can do o.CreateEndpoint(...)
to create a new service endpoint.

Implemented as a new module to not have dependency to gopher
on remaining lib-common content, where these are not used.


pkg/openstack/role.go
```diff
@@ -0,0 +1,119 @@
...
+	role, err := o.GetRole(log, roleName)
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return err
+	}
+
+	log.Info(fmt.Sprintf("Assigning userID %s to role %s - %s", userID, role.Name, role.ID))
+
+	err = roles.Assign(o.osclient, role.ID, roles.AssignOpts{
+		UserID:    userID,
+		ProjectID: projectID}).ExtractErr()
```
abays, line 113: What happens if we get here and the role was not found?

pkg/openstack/service.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	allServices, err := services.ExtractServices(allPages)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(allServices) == 0 {
+		return nil, k8s_errors.NewNotFound(
+			appsv1.Resource("Services"),
+			fmt.Sprintf("%s service not found in keystone", serviceName),
```
abays, line 87: Same question here as with roles

pkg/openstack/role.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	allRoles, err := roles.ExtractRoles(allPages)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(allRoles) == 0 {
+		return nil, k8s_errors.NewNotFound(
+			appsv1.Resource("Roles"),
+			fmt.Sprintf("%s role not found in keystone", roleName),
```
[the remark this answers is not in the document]
stuggi: yeah maybe, I just wanted to provide an easy way to catch not existing roles in keystone and handle it nice via NotFound error handling. Should we instead just return a nil ptr and the calling func has to check on that?

pkg/openstack/service.go
stuggi: same as above, could return nil instead and we need to catch it here?

pkg/openstack/role.go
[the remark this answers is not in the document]
stuggi: or return 3 values, a bool true when exist, false when not exist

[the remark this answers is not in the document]
stuggi: updated to remove NewNotFound

pkg/openstack/go.mod
```diff
@@ -0,0 +1,61 @@
+module github.com/openstack-k8s-operators/lib-common/pkg/openstack
```
bogdando, line 1: this creates a nested module and gopls complains about nested modules in vscode. Would it be possible to not introduce nested modules (directories with go.mod in it)?

stuggi: like its mentioned there, you can fix/workaround this buy open each module in its own workspace folder (File -> Add Folder to Workspace). There is currently the plan to introduce sub modules in the operators themselves to have dedicated module for the apis as this will reduce the go.mod dependencies on operators which just use other operators types.

bogdando: that workaround only helps to have a second instance of that directory in the workspace tree, and browse it without gopls errors. But when you go into the main repo directory (lib-common) with subpackages in it, it explodes with "problems". I would suggest to not use submodules because of that.

stuggi: with https://github.com/openstack-k8s-operators/lib-common/pull/29 we restructure lib-common so that it is easier to work with and only have modules in /modules. you'd then still need to open each module, but the main repo dir can basically ignored, if there is no need to add a new module.

abays (approved):
/lgtm

PR #26, opened 2022-07-11
Add envtest CRD helper

gibizer:
In envtest we need to be able to install the all the CRDs the operator
under test depends on.

This patch adds test.GetCRDDirFromModule helper to get the location of
CRDs from the operator go module defining it.

It also adds GetOpenShiftCRDDir function to look up the OpenShift CRDs.
As those are not published we store a copy of them in lib-common.

Currently we have this code directly in placement-operator proposed right now https://github.com/openstack-k8s-operators/placement-operator/pull/9 but it can be useful for all the operators test envs.


stuggi (approved):
/lgtm

stuggi (commented):
with the new structure of lib-common, this should now be either a package in modules/common, or we create a separate modules/test module.

gibizer:
@stuggi: I pushed a rebased version now that adds a test module in lib-common to hold this.

gibizer:
I sort of tested this together with the placement change(https://github.com/openstack-k8s-operators/placement-operator/pull/21) that start using the helpers from lib-common so I think these two change will work together. The "sort of" is there as the helper does not work with a "replace" line in the go.mod file so I did some manual hacking to test the integration.

PR #28, opened 2022-07-12
Check length of the serviceList before indexing

gibizer:
Database.setDatabaseHostname() call assumed that
common.GetServicesListWithLabel returns an error if no service is found.
However that call returns an empty list instead. This could lead to
panic in setDatabaseHostname when tried to index an empty list.

This patch extend the error guard condition to cover the empty list
case.

Closes openstack-k8s-operators#27


stuggi (approved):
/lgtm

PR #31, opened 2022-07-26
Allow some customization of endpoint URL format

abays:
New approach to `ExposeEndpoints` func that allows for customization of the URL generated for Keystone


stuggi (approved):
/lgtm

PR #32, opened 2022-07-27
Don't return an error if resource is not found in keystone

stuggi:
GetUser, GetService and GetRole returned an error if the
specified resource with the name could not be found, which
is still ok when checked if it exists or need to be created.
In that case return nil and the caller need to check on that
and acts appropriate.


abays (approved):
/lgtm

PR #33, opened 2022-07-28
module path fixes. Remove duplicate pkg

dprince:
This updates the module path to match that of the directory structure.

Also removes a duplicate 'pkg' tree


abays (approved):
/lgtm

PR #35, opened 2022-07-29
Fix service delete when not found

abays:
If an OpenStack service has already been deleted (or just didn't exist in the first place), the `gophercloud` client returns an error.  This error is not a Kubernetes `NotFound` error, however, and does not match the usual check for such things.


stuggi (approved):
/lgtm

PR #36, opened 2022-07-29
Fix OS mod to use endpoint package from common mod

abays:
The `openstack` module had some old code that was still trying to reference the old layout of `lib-common` where the `EndpointX` constants were found elsewhere than they are now


olliewalsh (approved):
/lgtm

PR #37, opened 2022-07-29
Fix OS user create/delete when not found

abays:
If an openstack user does not already exist when attempting to create the user, do not treat it as an error


stuggi (approved):
/lgtm

PR #38, opened 2022-08-01
Loop through all module dirs for checks

stuggi:
- Instead of rely on go.work file, just loop through all module directories in /modules for all the check targets.
- Add make target for go.work and to .gitignore


abays (approved):
/lgtm

PR #39, opened 2022-08-05
Update condition package for the proposed way to use them

stuggi:
* This implements condition functions in the condition pkg according to [1]
* Adds unit tests to condition package
* Updates secret package for new condition usage
* Updates database package for new condition usage
* Also bumps controller tools to 0.9.2
    
[1] https://github.com/openstack-k8s-operators/docs/pull/2


modules/common/condition/funcs.go
```diff
@@ -0,0 +1,205 @@
...
+	// According to this the Ready condition always goes first, followed by all the other
+	// conditions sorted by Type. This makes it easy to identify the overall state of
+	// the service
+	sort.Slice(*conditions, func(i, j int) bool {
+		return less(&(*conditions)[i], &(*conditions)[j])
+	})
+}
+
+// less returns true if a condition is less than another with regards to the
+// to order of conditions designed for better consumption i.e. cli client.
```
gibizer, line 159: nit: regards to the order

modules/common/condition/funcs.go
```diff
@@ -0,0 +1,205 @@
...
+// conditions sorted by Type. This makes it easy to identify the overall state of
+// the service
+func less(i, j *Condition) bool {
+	return (i.Type == ReadyCondition || i.Type < j.Type) && j.Type != ReadyCondition
+}
+
+// hasSameState returns true if a condition has the same state of another
+func hasSameState(i, j *Condition) bool {
+	return i.Type == j.Type &&
+		i.Status == j.Status &&
```
gibizer, line 170: I'm not sure what are the rules of go indentation but this sames strange to have it over-indented with one column

modules/common/condition/types.go
```diff
@@ -19,111 +19,65 @@ limitations under the License.
...
+
+	// Status of the condition, one of True, False, Unknown.
+	Status corev1.ConditionStatus `json:"status"`
+
+	// Severity provides a classification of Reason code, so the current situation is immediately
+	// understandable and could act accordingly.
+	// It is meant for situations where Status=False and it should be indicated if it is just
+	// informational, warning (next reconciliation might fix it) or an error (e.g. DB create issue
+	// and no actions to automatically resolve the issue can/should be done).
+	// +optional
```
gibizer, line 62: Do we want to keep these fields optional? I can accept that Message being optional as the current Reason might not need further explanation, but I think Severity and Reason should always be provided. I also see that this patch provides a SeverityNone for the success case too.

modules/common/condition/types.go
```diff
@@ -19,111 +19,65 @@ limitations under the License.
...
-		conditions.set(currentCondition)
-	}
-
-	//
-	// set new condition
-	//
-	conditions.set(&c)
-}
+// List - A list of conditions
+type List []Condition
```
gibizer, line 83: do we need this? I see type Conditions used in the new code.

stuggi: no, this is a left over

modules/common/condition/funcs.go
```diff
@@ -0,0 +1,205 @@
...
+	// If the condition does not exist, add it, setting the transition time only if not already set
+	if !exists {
+		if c.LastTransitionTime.IsZero() {
+			c.LastTransitionTime = metav1.NewTime(time.Now().UTC().Truncate(time.Second))
+		}
+		*conditions = append(*conditions, *c)
+	}
+
+	// Sort conditions list
+	conditions.Sort()
```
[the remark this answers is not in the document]
stuggi: I have updated the comment on Set() to describe the condition sort

modules/common/condition/types.go
stuggi: done

modules/common/condition/types.go
stuggi: latest commit make sure TrueCondition and UnknownCondition set SeverityNone

stuggi: also removed optional tag

modules/common/condition/funcs.go
```diff
@@ -0,0 +1,205 @@
...
+func (conditions *Conditions) Set(c *Condition) {
+	// if it is an empty condition, just return and don't set it
+	if c == nil {
+		return
+	}
+
+	// Check for the existence of a particular condition type in a list of conditions
+	// and change it only if there is a status change
+	exists := false
+	for i, existingCondition := range *conditions {
```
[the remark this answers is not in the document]
gibizer: Ohh, you are right. I missed the need of the index. Ignore my comment.

modules/common/condition/funcs.go
```diff
@@ -0,0 +1,205 @@
...
+			exists = true
+			if !hasSameState(&existingCondition, c) {
+				if c.LastTransitionTime.IsZero() {
+					c.LastTransitionTime = metav1.NewTime(time.Now().UTC().Truncate(time.Second))
+				}
+				(*conditions)[i] = *c
+
+				break
+			}
+			c.LastTransitionTime = existingCondition.LastTransitionTime
```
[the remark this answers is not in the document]
gibizer: that part is clear. We only want to update the persisted LastTransitionTime if there was a state transition. At L76 we know that there was no state transition due to the break at 74 never triggered.

My question is more about why we are updating the LastTransitionTime of the parameter c? Does the caller expects that the passed in parameter gets updated based on the persisted data?

modules/common/condition/funcs.go
gibizer: I need to get used to these strangeness of golang

modules/common/condition/funcs.go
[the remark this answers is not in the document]
gibizer: thanks

modules/common/condition/types.go
gibizer: SeverityNone works for me for Status=True.

modules/common/condition/funcs.go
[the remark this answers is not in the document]
stuggi: now I got it, sorry for not getting it sooner :) . right this is a leftover from a different previous version. this can be removed. I'll add a commit to do this which also extends the test for validate the LastTransitionTime won't change.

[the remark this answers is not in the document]
gibizer: ack

abays (approved):
/lgtm

PR #41, opened 2022-08-10
Exit when go checks detect error

stuggi:
* ignore vendor dirs if created in modules folders
* Exit if there is an error found in check runs looping over folders
* Fix EndpointData lint error


abays (approved):
/lgtm

PR #42, opened 2022-08-10
Add condition messages for keystone service

abays:


stuggi:
Lets wait with this until tomorrow, I have implemented conditions in the keystoneservice controller of the keystone-operator and testing right now if we could just sync the Ready condition from there to those which own the object. that probably could be enough and if there is something to dig deeper because it is stuck you then check the keystoneservice.

abays:
Okay, switching to draft to wait and see what happens

abays:
Closing to opt for mirroring approach instead

PR #43, opened 2022-08-11
Add Mirror functionality to sync lower level object condition state into the upper level

stuggi:
*   Add Mirror(t Type) to reflect client object conditions state
    
    Mirror(t) mirrors Status, Message, Reason and Severity from the
    latest condition of a sorted conditionGroup list into a target
    condition of type t.
    The conditionGroup entries are split by Status with the order
    False, True, Unknown. If Status=False its again split into Severity
    with the order Error, Warning, Info.
    
    If the overall ReadyConditon is true, it is expected that this is
    the actual state and no other groups need to be checked.
    
    To e.g. mirror the keystoneservice object condition it only required
    to use:
    ~~~
    c := ksSvc.GetConditions().Mirror(condition.KeystoneServiceReadyCondition)
    if c != nil {
      instance.Status.Conditions.Set(c)
    }
    ~~~
    
    Example Depends-On: https://github.com/openstack-k8s-operators/keystone-operator/pull/84

### DIFF
--- a/modules/common/condition/funcs.go
+++ b/modules/common/condition/funcs.go
@@ -157,6 +157,13 @@ func (conditions *Conditions) Sort() {
 	})
 }
 
+// SortByLastTransitionTime - Sorts a list of conditions by the LastTransitionTime
+func (conditions *Conditions) SortByLastTransitionTime() {
+	sort.Slice(*conditions, func(i, j int) bool {
+		return lessLastTransitionTime(&(*conditions)[i], &(*conditions)[j])
+	})
+}
+
 // less returns true if a condition is less than another with regards to the
 // order of conditions designed for better consumption i.e. cli client.
 // According to this the Ready condition always goes first, followed by all the other
@@ -164,6 +171,12 @@ func (conditions *Conditions) Sort() {
 // the service
 func less(i, j *Condition) bool {
 	return (i.Type == ReadyCondition || i.Type < j.Type) && j.Type != ReadyCondition
+}
+
+// lessLastTransitionTime returns true if a conditions LastTransitionTime is not before
+// another ones
+func lessLastTransitionTime(i, j *Condition) bool {
+	return !i.LastTransitionTime.Before(&j.LastTransitionTime)
 }
 
 // hasSameState returns true if a condition has the same state of another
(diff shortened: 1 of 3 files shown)


stuggi:
moving to draft as it need a go.mod update

stuggi:
> need go.mod update when [openstack-k8s-operators/keystone-operator#84](https://github.com/openstack-k8s-operators/keystone-operator/pull/84) landed

really not sure what I thought here ... there is no dependency to this PR. only when using it in the above example ...

modules/common/condition/funcs.go
```diff
@@ -157,6 +157,13 @@ func (conditions *Conditions) Sort() {
 	})
 }
 
+// SortByLastTransitionTime - Sorts a list of conditions by the LastTransitionTime
```
abays, line 160: Is the sorting here descending or ascending?  `lessLastTransitionTime` function makes it seem like it is descending, but just want to be sure

abays: Read a bit further.  It is descending: `// get the first conditon of the group which is the one with the latest LastTransitionTime`

modules/common/condition/funcs_test.go
```diff
@@ -277,6 +278,88 @@ func TestMarkMethods(t *testing.T) {
...
+
+	g.Expect(lessLastTransitionTime(falseA, falseB)).To(BeFalse())
+	g.Expect(lessLastTransitionTime(falseB, falseA)).To(BeTrue())
+
+	conditions := Conditions{}
+	cl := CreateList(falseB, falseError, falseA)
+	conditions.Init(&cl)
+	conditions.SortByLastTransitionTime()
+
+	// unknownReady has the current time stamp, do is first
```
abays, line 300: `s/do/so`

modules/common/condition/funcs_test.go
```diff
@@ -277,6 +278,88 @@ func TestMarkMethods(t *testing.T) {
...
+func TestMirror(t *testing.T) {
+	g := NewWithT(t)
+
+	time1 := metav1.NewTime(time.Date(2020, time.August, 9, 10, 0, 0, 0, time.UTC))
+	time2 := metav1.NewTime(time.Date(2020, time.August, 10, 10, 0, 0, 0, time.UTC))
+	time3 := metav1.NewTime(time.Date(2020, time.August, 11, 10, 0, 0, 0, time.UTC))
+
+	trueA.LastTransitionTime = time1
+	falseB.LastTransitionTime = time2
+	falseError.LastTransitionTime = time3
```
abays, line 313: `falseError` is modified here and never used below.  Should this be `falseBError` instead?

stuggi: falseError and  time3 can be removed. I changed the tests to use falseBError and missed to remove the falseError here.

modules/common/condition/funcs.go
```diff
@@ -218,3 +231,111 @@ func CreateList(conditions ...*Condition) Conditions {
 	}
 	return cs
 }
+
+// Mirror - mirros the latest condition in a sorted conditionGroup into a target
+// condition of type t. The conditionGroup entries is split by Status with the
+// oder False, True, Unknown. If Status=False its again split into Severity by
+// Error, Warning, Info.
+func (conditions *Conditions) Mirror(t Type) *Condition {
```
abays, line 239: It looks like `Type` is never considered when finding the condition to mirror.  Just wondering if someone called this with `Type` `X` and if it matched against a condition with `Type` `Y`, if we might see unexpected behavior (i.e. you would return a `Type` `X` with `Reason`, `Severity` and/or `Message` belonging to a condition that was actually `Type` `Y`).  Or perhaps that is not even possible and I am misreading something.

stuggi: t Type is the target condition you want to mirror to. It is not expected that this type is in the conditions list you want to mirror from. mirror gets the latest condition from the list using the in the comment described logic on what gets prioritized, no matter if the type matches. The use case is you have a new overall condition type from a service and mirror the latest state on this from the actual service to the level up.

stuggi: maybe we should call it tt, or targetType?

abays: Ah, I think I get it now.  So the condition of the level up will be whatever `Type` it already is, but its `Message` (and maybe `Reason` and `Severity`) is to be determined from the condition that the `Mirror` function picks.

I guess what I found confusing was the statement `mirrors the latest condition...into a target condition of type t`, being that conditions contain a `Type`, so mirroring a condition made me think everything about the mirrored condition should be copied (including `Type`).  But yeah, if we're explicitly passing a "target" `Type`, it makes sense that it would take precedence.

stuggi: right that comment is confusing. I'll try to get a better comment there.

stuggi: updated the comment and rebased. I hope the new comment makes it more clear on what it does or does not

modules/common/condition/funcs.go
```diff
@@ -218,3 +231,111 @@ func CreateList(conditions ...*Condition) Conditions {
...
+		case SeverityInfo:
+			return 2
+		}
+	case corev1.ConditionTrue:
+		return 3
+	case corev1.ConditionUnknown:
+		return 4
+	}
+
+	// this hopefully never happens
```
abays, line 339: For this to happen, someone would have to set the condition `Status` as `corev1.ConditionFalse` and then somehow set a non-existent arbitrary string for the `Severity`.  Or is even `corev1.ConditionStatus` just ultimately a string underneath and thus can be "corrupted"?

stuggi: something like that. basically this is a catchall for everything we don't accept/expect.

gibizer:
/lgtm 
thanks for the test cases that helped a lot to understand the mirroring logic

abays (approved):
/lgtm

PR #46, opened 2022-08-16
Refresh statefulset object on create/update

abays:
Adds a missing refresh of the lib-common's `StatefulSet` object's underlying `appsv1.StatefulSet` object after it is created/updated


stuggi (approved):
/lgtm

PR #47, opened 2022-08-18
Add IsError(c *Condition) to check if a condition has ErrorReason

stuggi:
Helper func to check if a condition represents an error. IsError is true if the condition is
a) not nil,
b) `status=False` and
c) `condition.Reason == condition.ErrorReason`

Otherwise it returns False if the condition is not True or if the condition does not exist (is nil).


abays (approved):
/lgtm

PR #48, opened 2022-08-19
Add GetHigherPrioCondition to compare prio of two conditions

stuggi:
GetHigherPrioCondition validates the priority of two conditions based on groupOrder(c) and returns the one which has precedence of the other.
If one of them is nil, the non nil get returned.

### DIFF
--- a/modules/common/condition/funcs_test.go
+++ b/modules/common/condition/funcs_test.go
@@ -34,15 +34,16 @@ var (
 	unknownReady = UnknownCondition(ReadyCondition, RequestedReason, ReadyInitMessage)
 	trueReady    = TrueCondition(ReadyCondition, ReadyMessage)
 
-	unknownA    = UnknownCondition("a", "reason unknownA", "message unknownA")
-	falseA      = FalseCondition("a", "reason falseA", SeverityInfo, "message falseA")
-	trueA       = TrueCondition("a", "message trueA")
-	unknownB    = UnknownCondition("b", "reason unknownB", "message unknownB")
-	falseB      = FalseCondition("b", "reason falseB", SeverityInfo, "message falseB")
-	falseBError = FalseCondition("b", "reason falseBError", SeverityError, "message falseBError")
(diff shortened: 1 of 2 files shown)


modules/common/condition/funcs_test.go
```diff
@@ -369,6 +370,38 @@ func TestIsError(t *testing.T) {
...
+	c = GetHigherPrioCondition(falseInfo, falseError)
+	g.Expect(hasSameState(c, falseError)).To(BeTrue())
+
+	// When bith Status=False, Severity Error has higher prio then Warning
+	c = GetHigherPrioCondition(falseWarning, falseError)
+	g.Expect(hasSameState(c, falseError)).To(BeTrue())
+
+	// When bith Status=False, Severity Warning has higher prio then Warning
+	c = GetHigherPrioCondition(falseWarning, falseInfo)
+	g.Expect(hasSameState(c, falseWarning)).To(BeTrue())
```
abays, line 402: Small nit: `s/bith/both`

stuggi:
There is one situation which is not considered... both have the same prio. need to add this

abays:
> There is one situation which is not considered... both have the same prio. need to add this

It looks like it would just default to the second condition in the comparison.  Did you want to be more explicit than that?

```
	if cond1 != nil && cond2 != nil {
		if groupOrder(*cond1) < groupOrder(*cond2) {
			return cond1
		}

		return cond2
	}
```

stuggi:
> > There is one situation which is not considered... both have the same prio. need to add this
> 
> It looks like it would just default to the second condition in the comparison. Did you want to be more explicit than that?
> 
> ```
> 	if cond1 != nil && cond2 != nil {
> 		if groupOrder(*cond1) < groupOrder(*cond2) {
> 			return cond1
> 		}
> 
> 		return cond2
> 	}
> ```

I enhanced it to return the one with the later LastTransitionTime.

abays (approved):
/lgtm

PR #51, opened 2022-09-05
Initial Ceph module support

fmount:
This patch adds the initial shared struct and functions to enable Ceph support
in OSP operators. Given that most of the operators share the same Ceph backend
data structure and utilities, it's useful to start this module and avoid a lot of
code duplication. This will be useful to build and to push the Ceph information to
the underlying storage operators (and other entities) that are going to interact
with Ceph.

The structure of the module is supposed to have:

- types: where the data structure is defined (and can be eventually
  extended)

- funcs: where the functions are exported and can be used by the
  operators

- util: where utility functions are implemented

Signed-off-by: Francesco Pantano <fpantano@redhat.com>


fmount:
@abays @fultonj this is a follow up of the work we did in the operators: it's a good idea having a ceph module that can be consumed by any operator in the future, so we can keep them slim enough and focus on the deployment/configuration logic instead of dealing with ceph structure internal processing.

The patch [1], which is still a draft, is a code refactoring that simplifies the operator and just includes what it needs. In the future, we might want to change that part and provide a different logic in the underlying operator.

[1] https://github.com/openstack-k8s-operators/cinder-operator/pull/11

fmount:
@dprince do you think it's better having a totally separated ceph module outside of lib-common? The goal is to avoid copy/paste the same functions to the storage operators that are supposed to interact with ceph and get that backend.

fmount:
> @dprince do you think it's better having a totally separated ceph module outside of lib-common? The goal is to avoid copy/paste the same functions to the storage operators that are supposed to interact with ceph and get that backend.

As discussed, moving the ceph module under the `storage` group

dprince:
@fmount I guess it depends on how many dependencies you end up needed in each module. Are they similar dependencies? If so then having them in the combined same directory tree would probably be fine. If not perhaps some benefit in splitting them out.

dprince:
@fmount can you add the go.sum as well. We usually add it to the repo I think

fmount:
> @fmount can you add the go.sum as well. We usually add it to the repo I think

No go.sum atm since we have no dependencies on the module

dprince (approved):

modules/storage/ceph/types.go
```diff
@@ -0,0 +1,63 @@
...
+
+package ceph
+
+// Defaults is used as type to reference defaults
+type Defaults string
+
+// Spec is the Cinder Ceph Backend struct defining defaults
+type Spec struct {
+	Defaults *Defaults
+}
```
abays, line 25: I can't find where this is used.  Is it something that would be imported elsewhere externally?

fmount: oh thanks, good catch, since now the module is shared and I reworked it, I don't need that part anymore.

abays (approved):
/lgtm

fultonj:
Thanks @fmount I see Dan and Andrew approve so I'm going to merge.

PR #52, opened 2022-09-07
Support multiple DB service instance

gibizer:
To support multi cell deployment we need to support multiple DB service instances as in a realistic multi cell deployment each Nova cell needs its own DB service instance.

So this patch introduces CreateOrPatchDBWithLabel to the database module where the caller can specify a label selector map to filter for a specific MariaDB service instance to use for its DB schema. The existing and heavily used CreateOrPatchDB is defaulted to use the "default" DB service instance but deprecated for removal after each service controller starts using  CreateOrPatchDBWithLabel instead.

### DIFF
--- a/modules/database/funcs.go
+++ b/modules/database/funcs.go
@@ -54,23 +54,25 @@ func NewDatabase(
 func (d *Database) setDatabaseHostname(
 	ctx context.Context,
 	h *helper.Helper,
+	labels map[string]string,
 ) error {
 
-	selector := map[string]string{
-		"app": "mariadb",
-	}
+	labels["app"] = "mariadb"
+
 	serviceList, err := service.GetServicesListWithLabel(
 		ctx,
 		h,
 		h.GetBeforeObject().GetNamespace(),
-		selector,
+		labels,
 	)
 	if err != nil || len(serviceList.Items) == 0 {
 		return fmt.Errorf("Error getting the DB service using label %v: %w",
-			selector, err)
+			labels, err)
 	}
 
-	// can we expect there is only one DB instance per namespace?
+	// We can expect that only one DB service instance matching the labels in
+	// the current namespace. If not then the caller needs to more specific by
+	// by adding more labels
 	if len(serviceList.Items) > 1 {
 		return util.WrapErrorForObject(
 			fmt.Sprintf("more then one DB service found %d", len(serviceList.Items)),
@@ -99,10 +101,25 @@ func (d *Database) GetDatabase() *mariadbv1.MariaDBDatabase {
 
 //
 // CreateOrPatchDB - create or patch the service DB instance
-//
+// Deprecated. Use CreateOrPatchDBWithLabel instead. If you want to use the
+// default the DB service instance of deployment then pass label
+// "cr": "mariadb-openstack" to CreateOrPatchDBWithLabel
 func (d *Database) CreateOrPatchDB(
 	ctx context.Context,
 	h *helper.Helper,
+) (ctrl.Result, error) {
+	return d.CreateOrPatchDBWithLabel(
+		ctx, h, map[string]string{"cr": "mariadb-openstack"})
+}
+
+//
+// CreateOrPatchDBWithLabel - create or patch the service DB instance on
+// the DB service selected via labels. To get the default DB Service instance
+// pass "cr": "mariadb-openstack" via the labels parameter.
+func (d *Database) CreateOrPatchDBWithLabel(
+	ctx context.Context,
+	h *helper.Helper,
+	labels map[string]string,
 ) (ctrl.Result, error) {
 
 	db := &mariadbv1.MariaDBDatabase{
@@ -117,7 +134,7 @@ func (d *Database) CreateOrPatchDB(
 	}
 
 	// set the database hostname on the db instance
-	err := d.setDatabaseHostname(ctx, h)
+	err := d.setDatabaseHostname(ctx, h, labels)
 	if err != nil {
 		return ctrl.Result{}, err
 	}


modules/database/funcs.go
```diff
@@ -99,11 +101,30 @@ func (d *Database) GetDatabase() *mariadbv1.MariaDBDatabase {
...
+// default DB service (labelled with cr=mariadb-openstack) will be used.
+//
+func (d *Database) CreateOrPatchDBWithLabel(
+	ctx context.Context,
+	h *helper.Helper,
+	labels map[string]string,
+) (ctrl.Result, error) {
+
+	if len(labels) == 0 {
+		labels["cr"] = "mariadb-openstack"
```
SeanMooney, line 126: so the useage for nova_cell1 or nova_cell2 will be to set 

labels["app"] = "mariadb"
lables["cr"] = "mariadb-nova-cell-1" 

that will be used as a selector as if we had done 

selector := map[string]string{
		"app": "mariadb",
        "cr": "mariadb-nova-cell-1" 
	}

when invoking service.GetServicesListWithLabel on line 66.

although we actually wont need to manually set app since you have set it on line 60.

modules/database/funcs.go
```diff
@@ -54,23 +54,25 @@ func NewDatabase(
 func (d *Database) setDatabaseHostname(
 	ctx context.Context,
 	h *helper.Helper,
+	labels map[string]string,
 ) error {
 
-	selector := map[string]string{
-		"app": "mariadb",
-	}
+	labels["app"] = "mariadb"
```
SeanMooney, line 60: I'm wondering if line 60 is actually need it as i would have guess the cr would have been enough

i assume ewe have callers that wont just pass the cr name as is being default in line 126 below

SeanMooney (commented):
i think this looks good to me over all
just two questions inline but i think this will help us move forward.

modules/database/funcs.go
gibizer: I did not want to change the "app" part of the selector to keep the existing behavior, i.e. not to break the existing operators. But in general we might want to consider using a more generic lookup to support both the adoption and the external DB use cases. As the interface now takes a label selector map it is easy to change lib common to use different default lookups if needed.

Yes, the NovaCell CR will have a "DB label selector" field that can be set like "cr":"mariadb-nova-cell-1" during deployment. And then the controller would use that info to call CreateOrPatchDBWithLabel and select the specific DB Service instance to use for the given nova cell instance.

modules/database/funcs.go
gibizer: Probably L126 is enough now. But I would like to keep the "app" filtering now. We can change this later easily. 

Currently the MariaDB CR  creation will result in the following Service instance:

```
$ oc describe service/openstack
Name:              openstack
Namespace:         openstack
Labels:            app=mariadb
                   cr=mariadb-openstack
                   owner=mariadb-operator
Annotations:       <none>
Selector:          app=mariadb
...
```
When an external service is defined via https://www.sobyte.net/post/2022-04/kubernetes-ext-service/ we might not have "cr" label (we could add them manually but it would not be true as there would be no CR behind that external DB). So we might want to filter for "app" still in that case and maybe use the Service.name filtering to select the nova-cell1 DB Service instance.

SeanMooney: ack that makes sense.

i was not pushing to drop this by the way just wanted to make sure i understood why you had boot when the cr would be enough in this specific case

lets keep it as you have it to maintain the same behavior as previously

modules/database/funcs.go
jistr: I should have said, the app=mariadb in CreateOrPatchDB wouldn't be default, that would simply be hardcoded behavior as it is right now. But moving this hardcoding from setDatabaseHostname to CreateOrPatchDB would mean that CreateOrPatchDBWithLabel would be unaffected by that hardcoding.

jistr: And CreateOrPatchDBWithLabel when `len(labels) == 0` should perhaps just fail.

SeanMooney (approved):
From my persepctive this makes sense and will enable us to deploy multiple db instance either for nova cells or for different  instance per OpenStack service to allow nova to be scasled indepently of neutron.

on that basis +2

as an aside i think we can likely use the same mechanism to fine a ExternalMariaDBInstacne CR in the same way.

we ill need to define what sub field are the common  subset that will form the API contract that we can depend on 

likely that will be the database hostname,user and password.

i see supporting external DB and Message buts as an important long term capability to maintain and a nessiarly table stakes feature for a 1.0 release but to enable adoption but also to support large customers that have dedicated team the manage/backup and optimse databases on behalf of the organisation as a whole.

just as we will support external ceph, we should support external dbs too to allow the opentack admin to focus on OpenStack and the DBAs to do what they do best.

i don't think we can ever only support external databases and it probably should not be our default but it should be a supported option and this will help make that posible

modules/database/funcs.go
gibizer: > Re defaults, how about putting app=mariadb default to CreateOrPatchDB - that way we preserve current behavior for users of that method. And then not handling anything for len(labels) == 0 in CreateOrPatchDBWithLabel. That means anyone who is switching from CreateOrPatchDB to CreateOrPatchDBWithLabel will have to provide all their desired labels explicitly.

Make sense I will do it.

> Another topic: From the adoption perspective, i'm wondering how could we support that use case -- prioritizing external adoption DB connection, and only if that doesn't exist, using podified MariaDB connection. Could we be doing two lookups here, first by adding some magic label "adoption=yes" (or something like that slightly_smiling_face ) to the set of provided labels, and only if that doesn't exist, look up the service by the provided labels only?

Nova will need to expose the label selector for DB Service instances on the CR as Nova needs to support multiple DB Service instances. So for Nova using a DB would mean setting the specific labels in the CR that matches with the Service instance that points to DB service to be used. So at start the labels in the Nova CR could be configured to select the Service instance that points to the external DB. And when we switch to the internal one the label in the Nova CR need to be changed to match with the internal Service instance.

For those service operators where a single (default) DB Service instance is used and therefore now label selector is provided on the CR we can do what you suggest and implement a logic and look for two set of default labels in precedence order. But alternatively we can keep a single default labels in this query and instead change the label on the Service instance to match the default query. So at deployment with external DB the a Service instance mapped to the external DB is created with the label that will match our default query. So the service operators will use that DB by default. Then when we deploy the internal DB then a new Service instance is created for that DB first without matching labels. When we switch we from external to internal, we need to change the labels on the external Service instance not to match the default query any more and change the label on the internal Service instance to match the default query. Then we probably need to kick the service operators to detect the change and reconcile to the new DB url.

jistr: > But alternatively we can keep a single default labels in this query and instead change the label on the Service instance to match the default query. So at deployment with external DB the a Service instance mapped to the external DB is created with the label that will match our default query. So the service operators will use that DB by default. Then when we deploy the internal DB then a new Service instance is created for that DB first without matching labels. When we switch we from external to internal, we need to change the labels on the external Service instance not to match the default query any more and change the label on the internal Service instance to match the default query. Then we probably need to kick the service operators to detect the change and reconcile to the new DB url.

+1 that sounds good to me.

gibizer: Moved the defaulting.

stuggi (approved):
/lgtm

PR #53, opened 2022-09-13
Allow ExposeEndpoint protocol to be other than TCP

steveb:
This change keeps the current default behaviour of assuming the Service is exposing TCP services, but adds a Protocol field to endpoint.Data to allow for different protocols.

Specifically, this is to allow a Service to be exposed for an Ironic TFTP service (UDP, port 69)


steveb:
Actually, its not strictly an endpoint, so I should create the Service directly

PR #54, opened 2022-09-13
add database finalizers add and remove logic 

pinikomarov:


modules/database/funcs.go
```diff
@@ -102,6 +103,7 @@ func (d *Database) GetDatabase() *mariadbv1.MariaDBDatabase {
 //
 func (d *Database) CreateOrPatchDB(
 	ctx context.Context,
+	instance *keystonev1.KeystoneAPI,
```
stuggi, line 106: we can not do that. we can not pass in a crd specific instance. the database functions are used by all the operators. the helper has the object if it is needed.

modules/database/funcs.go
```diff
@@ -135,6 +137,9 @@ func (d *Database) CreateOrPatchDB(
 			return err
 		}
 
+		// If the service object doesn't have our finalizer, add it.
+		controllerutil.AddFinalizer(instance, h.GetFinalizer())
```
stuggi, line 141: we want to set the finalizer on the `db` CR, not on the owner object. so I think it should be `controllerutil.AddFinalizer(db, h.GetFinalizer())` instead

modules/database/funcs.go
pinikomarov: ok

modules/database/funcs.go
pinikomarov: ok

modules/database/funcs.go
```diff
@@ -48,6 +48,17 @@ func NewDatabase(
 	}
 }
 
+// recieve a db pointer and remove it's finalizer
+// so it's set for deletion
+func Delete(
```
stuggi, line 53: not sure if we need a func for it. check how we do the finalizer remove in [1] for keystoneservice and keystoneendpoint. I think we could do the same for the mariadbdatabase finalizer.

[1] https://github.com/openstack-k8s-operators/placement-operator/blob/master/controllers/placementapi_controller.go#L210-L222

modules/database/funcs.go
```diff
@@ -48,6 +48,17 @@ func NewDatabase(
...
+	db *mariadbv1.MariaDBDatabase,
+	ctx context.Context,
+	helper *helper.Helper,
+
+) {
+	controllerutil.RemoveFinalizer(db, helper.GetFinalizer())
```
stuggi, line 59: you'd have to update the object when you removed the finalizer.

modules/database/funcs.go
```diff
@@ -229,3 +231,30 @@ func (d *Database) getDBWithName(
 
 	return nil
 }
+
+func GetServiceDbByName(
+	ctx context.Context,
+	h *helper.Helper,
+	name string,
+	namespace string,
+) (*mariadbv1.MariaDBDatabase, error) {
```
gibizer, line 240: This makes our Database abstraction leaky. Can we move the finalizer handling to the Database object completely?

modules/database/funcs.go
```diff
@@ -135,6 +134,9 @@ func (d *Database) CreateOrPatchDB(
 			return err
 		}
 
+		// If the service object doesn't have our finalizer, add it.
+		controllerutil.AddFinalizer(db, h.GetFinalizer())
```
gibizer, line 138: +1 I agree to hide the finalizer handling behind the Database abstraction in lib-common.

modules/database/funcs.go
gibizer: As the client will use DeleteFinalizer, you don't need to export his function any more so please make it private by renaming it to getServiceDbByName

modules/database/funcs.go
```diff
@@ -229,3 +231,51 @@ func (d *Database) getDBWithName(
...
+		return nil, util.WrapErrorForObject(
+			fmt.Sprintf("DB error %s %s ", name, h.GetBeforeObject().GetNamespace()),
+			h.GetBeforeObject(),
+			err,
+		)
+	}
+	return db, nil
+}
+
+func (d *Database) DeleteFinalizer(
```
gibizer, line 262: Does it compile with https://github.com/openstack-k8s-operators/placement-operator/pull/50/files#diff-c0f1c404ea10bc348fba06aab6f455b95b7513a337e09b325ebea615fc0ee587R211 ? It seems the placement patch calls this without the Database receiver.

Also I think you don't use `d` below so technically this does not need to be defined with a Database receiver.

gibizer (changes requested):

modules/database/funcs.go
gibizer: Reading the placement patch made me realize that you might have the Database receiver here so that you can call
```
db := NewDatabase()
db.DeleteFinalizer()
```
during the reconcileDelete() code path.

If that is the plan, then I would encourage to store the name and namespace of the MariaDBDatabase in the Database struct too so that the caller does not need to pass it to DeleteFinalizer any more.

modules/database/funcs.go
stuggi: @pinikomarov I thought you wanted to use GetServiceDbByName to return an initialized Database, so that you could do this in the cleanup:
db := GetServiceDbByName()
db.DeleteFinalizer()

instead of calling 
db := NewDatabase()
db.DeleteFinalizer()

gibizer: Ahh OK that can also be an option. So GetServiceDbByName() would return a *Database instead of *MariaDBDatabase. That works for me. But please rename GetServiceDbByName() to  GetDatabaseByName(). i) a service might have multiple DBs so the ServiceDb sounds strange. ii) this would made the name symmetric with NewDatabase.

modules/database/funcs.go
pinikomarov: @stuggi @gibizer 
A little confused about my options here:
So I'm going for the logic of :

In lib common/funcs return an initialized  *database by using :
func GetServiceDbByName(
  ctx context.Context,
  h *helper.Helper,
  name string,
  namespace string,
) (*Database, error) {
  **db := NewDatabase()**
[..]
  return db, nil

Following that , I can in  DeleteFinalizer do this :
func (d *Database) DeleteFinalizer(
  ctx context.Context,
  h *helper.Helper,
  name string,
  namespace string,
) error {
  serviceDb, err := GetServiceDbByName(ctx, h, name, namespace)
  #not sure ^^ this belongs here or in the placementapi_controller
[..]
controllerutil.RemoveFinalizer(serviceDb, helper.GetFinalizer())

Then in  placementapi_controller.go I can instead of :
  err := database.DeleteFinalizer(ctx, helper, placement.ServiceName,
instance.Namespace)

do :

  db := GetServiceDbByName()
  db.DeleteFinalizer()

So that's one option .
Thing is I'm not sure how to pass the namesace and service name to the db.DeleteFinalizer() without passing them as parameters
like : GetServiceDbByName(ctx, h, name, namespace)
@gibizer  you said adding them to the *Database struct so something like:
..if GetServiceDbByName() returns the *Database..add those fields to the struct:
db := db{
    name: name,
    namespace: namespace,
}
then in the DeleteFinalizer() access the name and namespace like : d.name and d.namespace
or supply the d *Database object as:
controllerutil.RemoveFinalizer(d, helper.GetFinalizer())

Hope you get my code path confusion here..:)

gibizer: I think I like the following approach:
lib-common:
```golang
func GetDatabaseByName(
	ctx context.Context,
	h *helper.Helper,
	name string,
	namespace string,
) (*Database, error) {
    // you can create a Database here by only filling what we need (name and namespace)
    db := &Database{
		name:         name,
		namespace:    namespace,
    }
    // the querying the MariaDBDatabase and store it in db by calling
    db.getDBWithName(ctx, h)
    return db
```

```golang
func (d *Database) DeleteFinalizer(
    ctx context.Context,
    h *helper.Helper,
) error {
    controllerutil.RemoveFinalizer(d.database, helper.GetFinalizer())
```

placement:
```golang
db := GetDatabaseByName(ctx, helper, <dbname>, instance.Namespace)
db.DeleteFinalizer()
```

pinikomarov: thanks!
I'll run some checks , see how that compiles

pinikomarov: @gibizer  ,ok  with the latest changes you mentioned and a few patches I made seems to work at for the keystoneAPI : 
```sh
oc get MariaDBDatabase -o json|jq .items[].metadata.finalizers
[
  "KeystoneAPI"
]
[
  "PlacementAPI"
]
[r@r install_yamls]$ oc delete KeystoneAPI keystone
keystoneapi.keystone.openstack.org "keystone" deleted
[r@r install_yamls]$  oc get MariaDBDatabase -o json|jq .items[].metadata.finalizers
[
  "PlacementAPI"
]

```
thanks

modules/database/funcs.go
```diff
@@ -229,3 +231,34 @@ func (d *Database) getDBWithName(
...
+	ctx context.Context,
+	h *helper.Helper,
+	name string,
+) *Database {
+	// you can create a Database here by only filling what we need (name and namespace)
+	db := &Database{
+		databaseName: name,
+	}
+	// then querying the MariaDBDatabase and store it in db by calling
+	db.getDBWithName(ctx, h)
```
gibizer, line 248: you need to handle the possible error returned here

modules/database/funcs.go
```diff
@@ -229,3 +231,34 @@ func (d *Database) getDBWithName(
...
 	return nil
 }
+
+//
+// Returns a *Database object with specified name and namespace
+//
+func GetDatabaseByName(
+	ctx context.Context,
+	h *helper.Helper,
+	name string,
```
gibizer, line 241: Let's call this `databaseName` to make it symmetric with `NewDatabase`

modules/database/funcs.go
```diff
@@ -229,3 +231,34 @@ func (d *Database) getDBWithName(
...
+) *Database {
```
gibizer, line 242: as `getDBWithName` might fail `GetDatabaseByName` can fail too. So you need to return (*Database, error) to be able to signal failure to the caller too.

gibizer (changes requested):
I think we are getting closer. Thank you for refining this. 

Besides the inline comments. I have two requests
1. Could you please clarify the title of the PR "use keystone api obj ..." It it not clear to me how keystone api obj is in the scope of this PR
2. Now you have a pretty long commit list in this PR containing various fixes. Before we merge could you please consolidate the list of commits? For example those that are fixing things introduced by an earlier commit in this PR can be squashed into the original commit. This cleanup helps keeping our git log more readable.

modules/database/funcs.go
pinikomarov: ok

modules/database/funcs.go
pinikomarov: ok

modules/database/funcs.go
pinikomarov: ok

modules/database/funcs.go
pinikomarov: ok , since the type of error is handled in getDBWithName I think returning an err or nil logic is ok here

pinikomarov:
@gibizer @stuggi I've completely mucked this pr with some dark --force git magic, 
continuing anew here : 
https://github.com/openstack-k8s-operators/lib-common/pull/87

PR #57, opened 2022-09-14
Update common and database module to remove not valid version

stuggi:
updates the two modules to remove the not valid version v0.0.0-00010101000000-000000000000


abays (approved):
/lgtm

PR #58, opened 2022-09-15
Move Service API struct to lib-common/api module

gibizer:
Move OpenStackServiceBase API struct to lib-common/api module

Each OpenStack service CRD will has a set of common fields like
ContainerImage, Replicas, NodeSelector, etc. So factor these common
fields out to a struct and move it to a common place. It is suggested
that each operator import this struct and embed it to the Spec struct
via:

	api.OpenStackServiceBase `json:",inline"`

The embedding and the inline tags makes this change almost fully backward
compatible with the original direct fields. The CRD will not change as
it will have the fields from OpenStackServiceBase directly inlined.
The controller code that reads Spec fields does not have to change as
the fields are directly accessible in the Spec struct due to unambiguous
embedding. However in the code where the Spec is instantiated (e.g. in
the tests) the Spec instantiation needs to use the

    OpenStackServiceBase: api.OpenStackServiceBase{...}

syntax to populate the embedded struct fields.


gibizer:
I will go and push a placement-operator update to demonstrate how to use this.

modules/api/common_types.go
```diff
@@ -0,0 +1,56 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package api
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
```
stuggi, line 22: I think we could also add the Debug type and maybe a generic PasswordSelector struct?

gibizer:
> I will go and push a placement-operator update to demonstrate how to use this.

Placement patch https://github.com/openstack-k8s-operators/placement-operator/pull/29

modules/api/common_types.go
```diff
@@ -0,0 +1,56 @@
...
+// Service contains the fields that are needed for each OpenStack service CRD
+type Service struct {
```
SeanMooney, line 24: a service is a standard type in kubernetes so we should not use this name

also we should likely embed this struct in the other operator rather then having it as a subfield via composition.

i would call it CRDBase

modules/api/common_types.go
SeanMooney: yep i think both makes sesne to add.

gibizer: Debug: I think each Service CRD will have a Debug.Service like parameter to stop the container before starting the service binary. But not each service will have Debug.DBSync as not each service (at least in nova) will run dbsync on the DB it uses. The APIDB in nova is shared between the NovaAPI, NovaScheduler, NovaConductor. Btw, if we are talking about Debug, I see PreserveJobs also something that belongs to Debug. But again maybe not all service will run Jobs.  Bottom line I'm not sure that Debug is generic enough to pulled out. 

PasswordSelector: That is definitely service specific, the default values there are different for each service type.

modules/api/common_types.go
gibizer: Good point about the naming. Let's call it OpenStackServiceBase. It is not CRDBase as there are CRDs which will not use it directly like the Nova CRD.

Could you be more specific what do you mean by embed vs. composition? In the commit message I suggest composition in a way that inlines the struct directly as fields on the API.

dprince:
How would this get used though? I really sort of like things like replicas defined at the top level of my CRD specs. spec.Replicas is much better than spec.Service.Replicas I think.

Yes there is some duplication in having these parameters defined in multiple places but so what. There isn't really much duplication of logic here. If you can figure out a way to share the names, but leave the struct implementation in each respective operator I feel we'd be much better off in terms of the resulting CRDs.

gibizer:
> How would this get used though? I really sort of like things like replicas defined at the top level of my CRD specs. spec.Replicas is much better than spec.Service.Replicas I think.

From the API perspective the struct is inlined so in the API there won't be any spec.Service it will be still spec.Replicas directly. On the controller implementation side there will be spec.Service.Replicas.

> 
> Yes there is some duplication in having these parameters defined in multiple places but so what. 

Duplication could mean at some point we inadvertently start to diverge.

> There isn't really much duplication of logic here. If you can figure out a way to share the names, but leave the struct implementation in each respective operator I feel we'd be much better off in terms of the resulting CRDs.

If only the resulting CRD is your concern then I think that is already solved by the "inline" tag I suggested in the commit message. 

At this point I'm not familiar enough with the different implementations in the different operators using these common fields. But I have a feeling that these common fields are use very similarly in the different controller. So we might even be able to pull out common implementation as well.

modules/api/common_types.go
SeanMooney: the password selector should liekly just be a reference to a secreate containing all the password that service needs

modules/api/common_types.go
SeanMooney: ya ok i cna get behind OpenStackServiceBase as the name your right its not the base of all CRDs

SeanMooney:
@dprince  if we embed the struct in the other struct by not giving it a name it would be referred as spec.replics. not spec.Service.Replicas 

that is what i meant by embedding vs composition to remove the extra layer of indriection

https://gobyexample.com/struct-embedding

SeanMooney:
an exmaple of embeding is

type NovaApi struct {
    Service
    str string // so other data
}

API := NovaApi{
 service: service{
   replica: 3
 } ,
 str: "my data"
}

you can then print replcia or sting the same way

 fmt.Println("replica:", api.base.replica)
 fmt.Println("also replica:", api.replica)
 fmt.Println("str:", api.str)

embedded stucts are unnamed stucts in a type definition that are inlined into the containing stuct

its go's answer to data member inheritance and is also how that works in c more or less.

modules/api/common_types.go
```diff
@@ -0,0 +1,56 @@
...
+	// +kubebuilder:validation:Required
+	// The service specific Container Image URL
+	ContainerImage string `json:"containerImage"`
```
bogdando, line 27: Given external dataplane with mixed-rhel cases, shouldn't we make this a map, like today tripleo framework aims to handle OS-dependent containers images type? This could be pointless for native resources, but making sense for external Nova computes.

gibizer: This is part of the Spec of the Nova<ServiceType> CR to describe what container image should be running in the pod. What would it mean if there would be a map of images? Which container image should be running in the pod?

bogdando: as I said, that would have not much sens for native services, but only exnternal ones, managed with ansible operator.
Those will not be deployed as pods, but information in this CR for container image to put into ansible inventory could be important for openstack meta operator watching these CRs... if that makes sens

SeanMooney: well the top level NOVA CR need an optional map of container images for each nova service CR that is deployed beneath it 

at the controller side only one image is going to be supproted at a time since the mix host os support is only for external dataplne

for the compute nodes i expect thso to be modeled out side of the top level NovaCRD and i expect they will have an image specific per compute node service and that might also be a map

e.g. {nova-compute:rhel10-nova-compute, nova-libirvirt:rhel9-nova-libvirt}

gibizer: > well the top level NOVA CR need an optional map of container images for each nova service CR that is deployed beneath it

The Nova CR will have it via the service CR instances contained in them. There is no need for a separate map. 

> 
> at the controller side only one image is going to be supproted at a time since the mix host os support is only for external dataplne
> 
> for the compute nodes i expect thso to be modeled out side of the top level NovaCRD and i expect they will have an image specific per compute node service and that might also be a map
> 
> e.g. {nova-compute:rhel10-nova-compute, nova-libirvirt:rhel9-nova-libvirt}

modules/api/common_types.go
gibizer: Done. I also refreshed the placement pull request using this in https://github.com/openstack-k8s-operators/placement-operator/pull/29

modules/api/common_types.go
gibizer: @SeanMooney: The PasswordSelector does not select a Secret. The Secret is selected via the Secret string field via the name of the Secret.  The PasswordSelector defines the field names the operator should look at in the Secret to find certain things. For me this is a bit too much indirection. I would hardcode the field names. But maybe @dprince or @stuggi has some good examples when this extra indirection is useful. Anyhow the current PasswordSelector cannot be pulled out to the generic Service struct as the default values in it are service specific.

Debug is setting debug mode for the operator behavior not Debug on the sevice binary.

gibizer:
> an exmaple of embeding is
> 
> type NovaApi struct { service str string // so other data }
> 
> API := NovaApi{ service: service{ replica: 3 } , str: "my data" }
> 
> you can then print replcia or sting the same way
> 
> fmt.Println("replica:", api.base.replica) fmt.Println("also replica:", api.replica) fmt.Println("str:", api.str)
> 
> embedded stucts are unnamed stucts in a type definition that are inlined into the containing stuct
> 
> its go's answer to data member inheritance and is also how that works in c more or less.

Thanks for the pointers. I learned something new about golang. :) I updated the commit message to promote embedding.
Also I refreshed the example patch using this in placement https://github.com/openstack-k8s-operators/placement-operator/pull/29

dprince:
One more comment would be API structs with operators are typically versioned. So v1beta1, v1beta2, etc. If we put this in lib-common and we later change the API how are we going to allow operators to update lib-common but avoid (potentially breaking) changes here if the operator using these doesn't support it.

gibizer:
I'm closing this for now as it seems this might be too much generalization. We can revisit this later when we gained experience how divergent the set of service parameters across our operators.

PR #59, opened 2022-09-19
Allow the Service Type and ExternalIPs to be set

steveb:
This accomodates Services which need to be a type other than ClusterIP (such as NodePort, etc).


steveb:
On second thoughts, special cases can just build the corev1.Service struct directly

PR #60, opened 2022-09-21
Fix hashing of ConfigMaps

Akrog:
Whenever an operator creates a new `ConfigMap` using methods `createOrPatchConfigMap` or `createOrGetCustomConfigMap` the calculated has will not be correct.

Due to this incorrect hash we end up seeing 1 unnecessary destroy/create cycle of the pods.

In the Cinder where this was observed we would see that at least the cinder-volume, cinder-backup, and cinder-scheduler containers would successfully run their init containers and then the services would start running displaying their logs, but a couple of seconds later they would just be terminated just to do the cycle again.

After 1 terminate/create cycle they would stabilize.

We could see that their respective `StatefulSet` had a `generation` count of 2 (when it should be 1) and this was caused by a change of the `ConfigMap` hash.

The problem comes from a difference in the hash calculation of a newly created `ConfigMap` and the same one when retrieved from OpenShift.

This happens because when we create the `ConfigMap` it will locally set the `TypeMeta` fields to empty strings (even if we set it otherwise), but when we get the `ConfigMap` from OpenShift it will have `Kind` and `APIVersion` set.

This patch solves the issue by only hashing the actual data from the `ConfigMap`, fields `Data` and `BinaryData` instead of the whole `ConfigMap`.

### DIFF
--- a/modules/common/configmap/configmap.go
+++ b/modules/common/configmap/configmap.go
@@ -34,6 +34,27 @@ import (
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 )
 
+
+
+// Hash function creates a hash of a ConfigMap's Data and BinaryData fields and
+// returns it as a safe encoded string.
+func Hash(configMap *corev1.ConfigMap) (string, error) {
+	type ConfigMapData struct {
+		Data map[string]string `json:"data,omitempty" protobuf:"bytes,2,rep,name=data"`
+		BinaryData map[string][]byte `json:"binaryData,omitempty" protobuf:"bytes,3,rep,name=binaryData"`
+	}
+
+	if configMap == nil {
+		return "", fmt.Errorf("nil ConfigMap doesn't have data to hash")
+	}
+
+	data := ConfigMapData{
+		Data: configMap.Data,
+		BinaryData: configMap.BinaryData,
+	}
+	return util.ObjectHash(data)
+}
+
 // createOrPatchConfigMap -
 func createOrPatchConfigMap(
 	ctx context.Context,
@@ -83,7 +104,7 @@ func createOrPatchConfigMap(
 		return "", op, fmt.Errorf("error create/updating configmap: %v", err)
 	}
 
-	configMapHash, err := util.ObjectHash(configMap)
+	configMapHash, err := Hash(configMap)
 	if err != nil {
 		return "", op, fmt.Errorf("error calculating configuration hash: %v", err)
 	}
@@ -128,7 +149,7 @@ func createOrGetCustomConfigMap(
 		configMap.Data = foundConfigMap.Data
 	}
 
-	configMapHash, err := util.ObjectHash(configMap)
+	configMapHash, err := Hash(configMap)
 	if err != nil {
 		return "", fmt.Errorf("error calculating configuration hash: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


modules/common/configmap/configmap.go
```diff
@@ -83,7 +96,7 @@ func createOrPatchConfigMap(
 		return "", op, fmt.Errorf("error create/updating configmap: %v", err)
 	}
 
-	configMapHash, err := util.ObjectHash(configMap)
```
stuggi, line 86: nice finding! I think it might be ok to just calc the hash on the configmap Data `configMapHash, err := util.ObjectHash(configMap.Data)` as this is usually the source we need to track for changes and e.g. restart resources if this information changes? Same applies for secrets I guess [1].

[1] https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/secret/secret.go#L53

Akrog: We would need to hash at least `configMap.Data` and `configMap.BinaryData`.  I could create a specific structure to hold those 2 fields and hash them, but I don't know if that's actually clearer or better, because may at some point a consumer of a `ConfigMap` could be interested in detecting changes to the `ObjectMeta` field (I really don't know, so I wanted to err on the safe side).

Another solution would be to `Get` the `ConfigMap` from the OpenShift cluster to have the `TypeMeta` fields filled in by the cluster itself.

Akrog:
Updated the new function name from `ConfigMapHash` to `Hash` as suggested by the golint job.

modules/common/configmap/configmap.go
stuggi: right `configMap.Data` and `configMap.BinaryData` missed the second. I am just thinking if it makes sense for our case to hash the metadata. Lets say a new release of an operator adds another label to the CM template. If that change is not intended that the consumer of the CM to pick a different CM we don't want the pod to be recreated. So just hashing the actual data which is used by the consumer is probably what we want. If we decide/agree on that, we can just create a hash of the two hashes and return this, which is probably what you said (create structure to hold those two and hash them).

If we want to keep the has on the full CM, I'd vote for your other solution to just `Get` the CM after it got created. It won't required any future change if there needs to be any additional filed be set.

Akrog: Updated the patch using a struct to only hash `Data` and `BinaryData`.

modules/common/configmap/configmap.go
```diff
@@ -34,6 +34,23 @@ import (
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 )
 
+
+
+// Hash function creates a hash of a ConfigMap's Data and BinaryData fields and
+// returns it as a safe encoded string.
+func Hash(configMap *corev1.ConfigMap) (string, error) {
```
stuggi, line 41: `Hash` does not return an error. It should probably return one if `configMap` is nil

modules/common/configmap/configmap.go
```diff
@@ -34,6 +34,23 @@ import (
...
+	type ConfigMapData struct {
+		Data map[string]string `json:"data,omitempty" protobuf:"bytes,2,rep,name=data"`
+		BinaryData map[string][]byte `json:"binaryData,omitempty" protobuf:"bytes,3,rep,name=binaryData"`
+	}
+
+	data := ConfigMapData{
+		Data: configMap.Data,
+		BinaryData: configMap.BinaryData,
+	}
+	return util.ObjectHash(data)
```
stuggi, line 51: `return util.ObjectHash(data), nil`

modules/common/configmap/configmap.go
stuggi: ahh `util.OpjectHash` does. maybe we check for configMap is nil

stuggi: other then that +2 from me

Akrog: Done

stuggi (approved):
/lgtm we have todo the same for secrets. I'll do that as a follow up

SeanMooney:
it would have been nice to also see some unit test for this.
this type of bug is one of the things that has made use nervouse about how hashs are currently used to detect change in the operators today

PR #62, opened 2022-09-22
Improvements to openstack module

stuggi:
* Introduce <Type>NotFound error const
    Adds ServiceNotFound, RoleNotFound and UserNotFound const, which allows to check for the specific error message and prevents hard coding specific error messages in different places.

* Validate in AssignUserRole() if the user is already part of role
    Before assigning the user to a role, check if it is already part of it. In this case no update to the role assignment is required.

* Check service is already present in CreateService()
    When the service with Name and Type is already present in keystone there is no need to create it. In this case just use it.

Also adds a new condition `KeystoneEndpointReadyCondition` and rewords the `ReadyInitMessage` and `ReadyMessage` to be more generic as it is not always a service which gets configured.

### DIFF
--- a/modules/openstack/role.go
+++ b/modules/openstack/role.go
@@ -100,13 +100,31 @@ func (o *OpenStack) AssignUserRole(
 		return err
 	}
 
-	log.Info(fmt.Sprintf("Assigning userID %s to role %s - %s", userID, role.Name, role.ID))
-
-	err = roles.Assign(o.osclient, role.ID, roles.AssignOpts{
-		UserID:    userID,
-		ProjectID: projectID}).ExtractErr()
+	// validate if user is already assigned to role
+	listAssignmentsOpts := roles.ListAssignmentsOpts{
+		ScopeProjectID: projectID,
+		UserID:         userID,
+		RoleID:         role.ID,
+	}
+	allPages, err := roles.ListAssignments(o.osclient, listAssignmentsOpts).AllPages()
 	if err != nil {
 		return err
+	}
+
+	assignUser, err := allPages.IsEmpty()
+	if err != nil {
+		return err
+	}
+
+	if assignUser {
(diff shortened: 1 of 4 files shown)


modules/openstack/service.go
```diff
@@ -39,24 +39,44 @@ func (o *OpenStack) CreateService(
...
+		serviceID = service.ID
 	}
 
-	return service.ID, nil
+	return serviceID, nil
 }
 
+// ServiceNotFound - service not found error message"
+const ServiceNotFound = "service not found in keystone"
+
```
abays, line 79: Should we maybe put this at the top somewhere for organizational/readability purposes?

modules/openstack/user.go
```diff
@@ -71,6 +71,9 @@ func (o *OpenStack) CreateUser(
 	return userID, nil
 }
 
+// UserNotFound - user not found error message"
+const UserNotFound = "user not found in keystone"
+
```
abays, line 76: Same placement question here as for the new service const

modules/openstack/role.go
```diff
@@ -63,6 +64,9 @@ func (o *OpenStack) CreateRole(
 	return roleID, nil
 }
 
+// RoleNotFound - role not found error message"
+const RoleNotFound = "role not found in keystone"
+
```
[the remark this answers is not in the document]
abays: Oh boy, it's already at the top.  Nevermind. :)

modules/openstack/service.go
stuggi: done

modules/openstack/user.go
stuggi: done

modules/openstack/role.go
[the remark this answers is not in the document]
stuggi: done

abays (approved):
/lgtm

PR #64, opened 2022-09-22
Rework endpoint URL building

steveb:
* Include the TCP port number in the endpoint
* Use the service name as the hostname for internal endpoints, since that will work for openstack pod -> pod requests
* Get some parse validation but punt on ever parsing the path component because of %(project_id)s placeholder
* Special-case port 80 with http

Example endpoints created with this:
- interface: public url: http://ironic-public-openstack.apps.dell-opti.stevebaker.org:6385/v1/82b04483205f4c499826f51ec532c89f
- interface: internal url: http://ironic-internal:6385/v1/82b04483205f4c499826f51ec532c89f

Handling https remains a TODO

### DIFF
--- a/modules/common/endpoint/endpoint.go
+++ b/modules/common/endpoint/endpoint.go
@@ -19,6 +19,7 @@ package endpoint
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"strings"
 
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
@@ -120,22 +121,29 @@ func ExposeEndpoints(
 		//
 		// Update instance status with service endpoint url from route host information
 		//
-		// TODO: need to support https default here
-		var apiEndpoint string
-		if !strings.HasPrefix(route.GetHostname(), "http") {
-			apiEndpoint = fmt.Sprintf("http://%s%s", route.GetHostname(), data.Path)
-		} else {
-			apiEndpoint = fmt.Sprintf("%s%s", route.GetHostname(), data.Path)
-		}
-		// TODO: This check does not allow for certain "special" OpenStack endpoint
-		//       formats.  For instance, "http://<ip>:<port>/v2/%(project_id)s" fails
-		//       to parse because "%(" is an invalid escape character
-		// u, err := url.Parse(apiEndpoint)
-		// if err != nil {
-		// 	return endpointMap, ctrlResult, err
-		// }
+		var protocol string
+		var port string
+		hostname := route.GetHostname()
 
-		endpointMap[string(endpointType)] = apiEndpoint
+		// TODO: need to support https default here
+		if !strings.HasPrefix(hostname, "http") {
+			protocol = "http://"
+		} else {
+			protocol = ""
+		}
+		if data.Port == 80 && protocol == "http://" {
+			port = ""
+		} else {
+			port = fmt.Sprintf(":%d", data.Port)
+		}
+
+		// Do not include data.Path in parsing check because %(project_id)s
+		// is invalid without being encoded, but they should not be encoded in the actual endpoint
+		apiEndpoint, err := url.Parse(protocol + hostname + port)
+		if err != nil {
+			return endpointMap, ctrlResult, err
+		}
+		endpointMap[string(endpointType)] = apiEndpoint.String() + data.Path
 	}
 
 	return endpointMap, ctrl.Result{}, nil


steveb:
If this change is going in the right direction I think a cross-operator issue needs to be raised to ensure *.conf [keystone_authtoken]auth_url etc use the internal endpoint, I'm seeing public endpoints being used here, for example  https://github.com/openstack-k8s-operators/cinder-operator/blob/master/templates/cinder/config/cinder.conf#L46

stuggi:
> * Include the TCP port number in the endpoint
> * Use the service name as the hostname for internal endpoints, since that will work for openstack pod -> pod requests

I think this would only work if the full env is running on OCP. In our current approach we'll still have components, like computes, running external to the OCP cluster. They still need to be able to connect to the internal endpoint.

> * Get some parse validation but punt on ever parsing the path component because of %(project_id)s placeholder
> * Special-case port 80 with http
> 
> Example endpoints created with this:
> 
> * interface: public url: http://ironic-public-openstack.apps.dell-opti.stevebaker.org:6385/v1/82b04483205f4c499826f51ec532c89f
> * interface: internal url: http://ironic-internal:6385/v1/82b04483205f4c499826f51ec532c89f
> 
> Handling https remains a TODO

stuggi:
> > * Include the TCP port number in the endpoint
> > * Use the service name as the hostname for internal endpoints, since that will work for openstack pod -> pod requests
> 
> I think this would only work if the full env is running on OCP. In our current approach we'll still have components, like computes, running external to the OCP cluster. They still need to be able to connect to the internal endpoint.

We could still configure the service name on the services hosted on OCP, but in keystone we'd have to register a route for the endpoints

> 
> > * Get some parse validation but punt on ever parsing the path component because of %(project_id)s placeholder
> > * Special-case port 80 with http
> > 
> > Example endpoints created with this:
> > 
> > * interface: public url: http://ironic-public-openstack.apps.dell-opti.stevebaker.org:6385/v1/82b04483205f4c499826f51ec532c89f
> > * interface: internal url: http://ironic-internal:6385/v1/82b04483205f4c499826f51ec532c89f
> > 
> > Handling https remains a TODO

steveb:
> I think this would only work if the full env is running on OCP. In our current approach we'll still have components, like computes, running external to the OCP cluster. They still need to be able to connect to the internal endpoint.

Oh good point, this also touches on unanswered questions of the overall network architecture. I'll rework this change to go back to public names and always creating routes.

stuggi (approved):
/lgtm

abays (approved):
/lgtm

fmount:
@abays @steveb @stuggi  @fultonj  this patch has broken at least glance:

Services are properly exposed on 9292:

```
glance-admin                                           ClusterIP   10.217.5.165   <none>        9292/TCP    3s
glance-internal                                        ClusterIP   10.217.4.179   <none>        9292/TCP    3s
glance-public                                          ClusterIP   10.217.4.26    <none>        9292/TCP    2s
```

Routes have been created as expected, and they point to the right service:

```
glance-admin        glance-admin-openstack.apps-crc.testing               glance-admin        glance-admin                      None
glance-internal     glance-internal-openstack.apps-crc.testing            glance-internal     glance-internal                   None
glance-public       glance-public-openstack.apps-crc.testing              glance-public       glance-public                     None
```


However, the glance endpoints have been created like the following:

```
| 93171168406d46a5a71098ad1e1a09ea | regionOne | glance       | image        | True    | admin     | http://glance-admin-openstack.apps-crc.testing:9292    |
| a29a61b86ae14a3ba2feabacb4b46fd5 | regionOne | glance       | image        | True    | public    | http://glance-public-openstack.apps-crc.testing:9292   |
| e899f276fce74901b09419b4b1621872 | regionOne | glance       | image        | True    | internal  | http://glance-internal-openstack.apps-crc.testing:9292 |
```

and a simple `openstack image list` will fail because it tries to connect to the endpoint on 9292, which is wrong.

```
Starting new HTTP connection (2): glance-public-openstack.apps-crc.testing:9292
Unable to establish connection to http://glance-public-openstack.apps-crc.testing:9292/v2/images: HTTPConnectionPool(host='glance-public-openstack.apps-crc.testing', port=9292): Max retries exceeded with url: /v2/images (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fee170b40d0>: Failed to establish a new connection: [Errno 111] Connection refused'))
clean_up ListImage: Unable to establish connection to http://glance-public-openstack.apps-crc.testing:9292/v2/images: HTTPConnectionPool(host='glance-public-openstack.apps-crc.testing', port=9292): Max retries
exceeded with url: /v2/images (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fee170b40d0>: Failed to establish a new connection: [Errno 111] Connection refused'))
END return value: 1
```

A `route` simply exposes a service on a  hostname, so that external clients can reach it by name, and it points to a service which exposes the right port (where the service is bound), while `NodePort` or other `ingress` related strategies can be probably used to expose the k8s service on a given port on all the hosts.
Having the port defined in the endpoint url let the client think that the service is explicitly exposed on that `host:port`, and it will fail connecting to it.

I find as an improvement the fact you added `net/url` to make sure we have a sanity check over the created url, but I suspect the `data.Port` and  `data.Path` related changes can be removed until we figure out the right way to rework this part.

fultonj:
I hit the same problem in my environment.

Issue #65, opened 2022-10-02
CreateOrPatchDB fails if the CR Name and the requested Database schema name are different


gibizer:
Here the code uses h.GetBeforeObject().GetName() as the name of the new MariaDBDatabase CR
https://github.com/openstack-k8s-operators/lib-common/blob/66e396fb480ec8e0222e9a31df0ae6580af99542/modules/database/funcs.go#L119-L128

then later it uses the Database.databaseName as the name to look up the MariaDBDatabase CR
https://github.com/openstack-k8s-operators/lib-common/blob/66e396fb480ec8e0222e9a31df0ae6580af99542/modules/database/funcs.go#L217-L225

So if the two name are not equal then the DB creation fails. I suggest to always used the database.databaseName.

PR #66, opened 2022-10-02
Use Database.databaseName consistently to fix issue #65 

gibizer:
[database]Use databaseName for MariaDBDatabase Name

The CreateOrPatchDB() call used the caller's Name as the Name of the
MariaDBDatabase CR it created. But then later it tried to look it up by
using the Database.databaseName. When the two differs the CreateOrPatchDB
fails to create the DB.

Closes https://github.com/openstack-k8s-operators/lib-common/issues/65
Depends-on https://github.com/openstack-k8s-operators/lib-common/issues/67


abays (commented):
+1 from me, but should probably let @dprince or @stuggi review too

PR #67, opened 2022-10-02
Add NewDatabaseWithNamespace

steveb:
Deriving the database namespace from the before object doesn't work when that object has no namespace (such as for a Cluster scoped operator).

This change adds a variant of NewDatabase which allows a custom name and namespace.

### DIFF
--- a/modules/database/funcs.go
+++ b/modules/database/funcs.go
@@ -45,6 +45,27 @@ func NewDatabase(
 		databaseUser: databaseUser,
 		secret:       secret,
 		labels:       labels,
+		name:         "",
+		namespace:    "",
+	}
+}
+
+// NewDatabaseWithNamespace returns an initialized DB.
+func NewDatabaseWithNamespace(
+	databaseName string,
+	databaseUser string,
+	secret string,
+	labels map[string]string,
+	name string,
+	namespace string,
+) *Database {
+	return &Database{
+		databaseName: databaseName,
+		databaseUser: databaseUser,
+		secret:       secret,
+		labels:       labels,
+		name:         name,
+		namespace:    namespace,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


modules/database/funcs.go
```diff
@@ -45,6 +45,27 @@ func NewDatabase(
 		databaseUser: databaseUser,
 		secret:       secret,
 		labels:       labels,
+		name:         "",
```
gibizer, line 48: I don't think we want to have a different name and databaseName for a MariaDBDatabase.

modules/database/funcs.go
```diff
@@ -122,15 +143,25 @@ func (d *Database) CreateOrPatchDBWithLabel(
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      h.GetBeforeObject().GetName(),
-			Namespace: h.GetBeforeObject().GetNamespace(),
-		},
-		Spec: mariadbv1.MariaDBDatabaseSpec{
-			// the DB name must not change, therefore specify it outside the mutuate function
-			Name: d.databaseName,
-		},
+	if d.name == "" {
+		d.name = h.GetBeforeObject().GetName()
```
gibizer, line 147: Just a note that this is actually a bug to user the before object name here. https://github.com/openstack-k8s-operators/lib-common/pull/66

gibizer (commented):
I would not add a name field to the Database struct.

modules/database/funcs.go
gibizer: Hm. I might need to reconsider this.  The k8s Name field needs to be in a format of a subdomain so the charachter set is restricted. I.e. you cannot use underscore in the Name. But DB schema names can have underscores and nova heavily uses it by default e.g. nova_cell0. So we might really need both name and databaseName in the Database structure.

gibizer:
> I would not add a name field to the Database struct.

I changed my mind. We are better of separating the name of the MariaDBDatabase CR and the name of the DB schema it creates.

/lgtm

modules/database/funcs.go
```diff
@@ -219,24 +250,28 @@ func (d *Database) getDBWithName(
 	h *helper.Helper,
 ) error {
 	db := &mariadbv1.MariaDBDatabase{}
+	namespace := d.namespace
+	if namespace == "" {
+		namespace = h.GetBeforeObject().GetNamespace()
```
abays, line 255: If the object `h` is wrapping is cluster-scoped, couldn't this set `namespace` to the empty string?  If so, would that cause problems for the `Get` on line 257 (where the namespace is needed)?

gibizer: I think the expected usage is:
* if h wraps cluster scoped object then provide a namespace explicitly. This way L257 will have a proper namespace.
* if h wraps a namespaced object then you don't need to provide a namespace explicitly and the call will use the same namespace as h.

abays (commented):
+1 from me, but should probably let @dprince or @stuggi review too

PR #68, opened 2022-10-04
Make requeue timeouts more configurable

gibizer:
The overall goal is to make timeouts precisely configurable so that in envtest where DB, Job and Deployment creation success is simulated the test execution can be speeded up by setting a sub-second RequeueAfter value.

To achieve this:
* WaitForDBCreatedWithTimeout is introduced to take a timeout value in time.Duration instead of using a hard coded 5secs.

* Deployment and Job is changed to store time.Duration instead of seconds internally and a SetTimeout funcs are introduced to make that configurable

### DIFF
--- a/modules/common/deployment/deployment.go
+++ b/modules/common/deployment/deployment.go
@@ -38,7 +38,7 @@ func NewDeployment(
 ) *Deployment {
 	return &Deployment{
 		deployment: deployment,
-		timeout:    timeout,
+		timeout:    time.Duration(timeout) * time.Second,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


modules/common/deployment/deployment.go
```diff
@@ -38,7 +38,7 @@ func NewDeployment(
 ) *Deployment {
 	return &Deployment{
 		deployment: deployment,
-		timeout:    timeout,
+		timeout:    time.Duration(timeout) * time.Second,
```
abays, line 41: Is there perhaps a desire to make `timeout` a `Duration` type itself, like you've done for `WaitForDBCreatedWithTimeout`'s `requeueAfter` parameter?  I think it's fine the way you have it, but I was just curious.  Obviously changing it from `int` to `Duration` would require all the operator callers to update their use of `NewDeployment` when they update to the new `lib-common` version, so perhaps that's sub-optimal.

abays (approved):
/lgtm

gibizer:
Just rebased.

PR #69, opened 2022-10-09
Do not log error for DB creation

gibizer:
Creating or updating a MariaDBDatabase should not be reported as error as it is not really an error condition. The controller just need to requeue the reconcilation to wait for the MariaDBDatabase reconciliation but should not log an error with a stack trace.


stuggi (approved):
/lgtm

PR #70, opened 2022-10-10
Add more unit tests to util pkg in common mod

stuggi:


modules/common/util/map_test.go
```diff
@@ -7,28 +7,72 @@ import (
...
-	t.Run("Nils empty maps", func(t *testing.T) {
-		g := NewWithT(t)
+			want: map[string]string{
+				"a": "a",
+				"b": "b",
+				"c": "c",
+			},
+		},
+		{
+			name: "Merge maps with existing key, the value in the first map is preserved",
```
abays, line 49: Maybe this name should instead indicate something about merging empty maps?

Makefile
```diff
@@ -27,13 +27,28 @@ ENVTEST_K8S_VERSION = 1.24
 all: build
 
 .PHONY: fmt
-fmt: ## Run go fmt against code.
-	go fmt ./...
+fmt: gowork ## Run go fmt against code.
```
olliewalsh, line 30: maybe should make it clear in the comment how the fmt and gofmt targets differ

modules/common/util/funcs_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+	tests := []struct {
+		name string
+		data string
+		want string
+	}{
+		{
+			name: "Valid json string",
+			data: `{"some":"json"}`,
+			want: "",
```
olliewalsh, line 107: Would be better to include the err nil-ness (word?) too, e.g this would pass if an empty error string is returned

modules/common/util/funcs_test.go
```diff
@@ -0,0 +1,165 @@
...
+		},
+		{
+			name: "Empty string",
+			data: "",
+			want: "unexpected end of JSON input",
```
olliewalsh, line 112: Or maybe would be better to just validate that err is not nil? These failures are coming from the json parsing I assume and could change slightly between versions

abays (approved):
/lgtm

modules/common/util/hash_test.go
```diff
@@ -22,55 +22,91 @@ import (
...
+			hash, err := ObjectHash(tt.data)
+			g.Expect(err).To(BeNil())
+
+			g.Expect(hash).To(BeIdenticalTo(tt.want))
+		})
+	}
+}
+
 //
 // TestSetHash - create or patch the service DB instance
```
olliewalsh, line 52: comment incorrect?

stuggi: removed

modules/common/util/funcs_test.go
stuggi: done

modules/common/util/funcs_test.go
stuggi: done, see bellow

olliewalsh (approved):
/lgtm

PR #71, opened 2022-10-11
Fix subdir removal from GetAllTemplates(), add unit tests and comments

stuggi:
When there are multiple sub dirs to be removed from the templatedFiles list, the index no longer matches when the first one got removed. The removal loop should start from beginning when an entry got removed.

Also adds unit tests for the template util and comments to make it easier to understand the templating.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/70

### DIFF
--- a/modules/common/util/template_util.go
+++ b/modules/common/util/template_util.go
@@ -34,11 +34,12 @@ import (
 type TType string
 
 const (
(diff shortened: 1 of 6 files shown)


abays (approved):
/lgtm

PR #75, opened 2022-10-12
[test]Fix CRD utils after CRDs moved to API package

gibizer:
We moved the CRD definitions of keystone and mariadb to the "bases" directory  of the api package. Adapt the CRD test utils to that change.


stuggi (approved):
/lgtm

PR #76, opened 2022-10-13
Add helpers pkg

stuggi:
Adds the common helpers from current nova-operator as base for common testing helpers.


modules/test/helpers/configmap.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetConfigMap -
+func GetConfigMap(name types.NamespacedName) corev1.ConfigMap {
+	cm := &corev1.ConfigMap{}
+	gomega.Eventually(func(g gomega.Gomega) {
+		g.Expect(k8sClient.Get(ctx, name, cm)).Should(gomega.Succeed())
+	}, timeout, interval).Should(gomega.Succeed())
```
stuggi, line 30: timeout should probably be a parameter that each caller can have its own timetout/interval

stuggi: https://github.com/openstack-k8s-operators/lib-common/pull/76/commits/1a42a1fc9cf2a9b49b969dea1c6c1660647abe5c

gibizer:
Based on my nova experience a single timeout and interval value per test suite is enough. So I'm wondering how could we avoid repeating timeout and interval at each invocation as that makes the test cases noisy. What do you think about creating a struct that wraps the client and the timeouts and make our helper functions having that struct as a receiver? This way the test suite can instantiate the struct once with a timeout and interval value. But also a test case can decide that it needs different timeout value to create a separate struct instance.

stuggi:
> Based on my nova experience a single timeout and interval value per test suite is enough. So I'm wondering how could we avoid repeating timeout and interval at each invocation as that makes the test cases noisy. What do you think about creating a struct that wraps the client and the timeouts and make our helper functions having that struct as a receiver? This way the test suite can instantiate the struct once with a timeout and interval value. But also a test case can decide that it needs different timeout value to create a separate struct instance.

yeah that sounds good. let me do that.

modules/test/helpers/common.go
```diff
@@ -0,0 +1,60 @@
...
+	ginkgo "github.com/onsi/ginkgo/v2"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// TestHelper -
+type TestHelper struct {
+	k8sClient client.Client
+	ctx       context.Context
+	cancel    context.CancelFunc
```
gibizer, line 31: I'm not sure we need to hold the cancel here. It is not used by the helpers. It is only used by the test suite.

modules/test/helpers/common.go
```diff
@@ -0,0 +1,60 @@
...
+	timeout   time.Duration
+	interval  time.Duration
+}
+
+// NewDefaultTestHelper returns a TestHelper with some defaults
+func NewDefaultTestHelper(
+	k8sClient client.Client,
+) *TestHelper {
+	ctx, cancel := context.WithCancel(context.TODO())
```
gibizer, line 40: I would rather pass in the ctx than create it here.

modules/test/helpers/common.go
```diff
@@ -0,0 +1,60 @@
...
+
+	return &TestHelper{
+		k8sClient: k8sClient,
+		ctx:       ctx,
+		cancel:    cancel,
+		timeout:   time.Second * 10,
+		interval:  time.Millisecond * 200,
```
gibizer, line 47: I would make this configurable by passing in the values.

gibizer (changes requested):

modules/test/helpers/common.go
stuggi: yes, I just also added it as it is connected to the ctx as tuple. but yes when we pass the ctx as you stated bellow as a param we can remove it.

modules/test/helpers/common.go
stuggi: this was meant to get a helper with defaults for timeout/interval. I'll rename it to NewTestHelper and make them as params

modules/test/helpers/common.go
stuggi: done

modules/test/helpers/common.go
stuggi: done

modules/test/helpers/common.go
stuggi: done

gibizer: /me is sad about not having default parameters in golang.

Right now I'm not sure about what would be the right defaults. I'm actively fine tuning the nova defaults. So it is better to keep it configurable. When we have the default settled then we can move that default over to lib-common.

stuggi: sure, we can add then a func which initialize them with some good defaults.

gibizer (approved):
Thanks. Looks good to me. 
I see the following next steps: 
1. use this from nova
2. bump gingko to v2 in placement
3. use this from placement

modules/test/helpers/deployment.go
```diff
@@ -0,0 +1,53 @@
...
+
+	return deployment
+}
+
+// ListDeployments -
+func (tc *TestHelper) ListDeployments(namespace string) *appsv1.DeploymentList {
+	deployments := &appsv1.DeploymentList{}
+	gomega.Expect(tc.k8sClient.List(tc.ctx, deployments, client.InNamespace(namespace))).Should(gomega.Succeed())
+
+	return deployments
```
bogdando, line 39: same

modules/test/helpers/deployment.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+// GetDeployment -
+func (tc *TestHelper) GetDeployment(name types.NamespacedName) *appsv1.Deployment {
+	deployment := &appsv1.Deployment{}
+	gomega.Eventually(func(g gomega.Gomega) {
+		g.Expect(tc.k8sClient.Get(tc.ctx, name, deployment)).Should(gomega.Succeed())
+	}, tc.timeout, tc.interval).Should(gomega.Succeed())
+
+	return deployment
```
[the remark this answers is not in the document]
gibizer: the `deployment` variable is already inferred to have the type `*appsv1.Deployment` as the initialization of it `&appsv1.Deployment{}` has that type.

PR #77, opened 2022-10-14
Do not delete the Job until it's hash is persisted

gibizer:
We have a logic error around Job deletion.

1. controller calls DoJob() on lib-common's Job
2. lib-common waits for the Job to finish and if succeeds, it deletes the Job.
3. the controller calculates the hash of the Job and puts it to Status.Hash
4. the controller calls Status.Update() to persist the hash to the DB

If step 4. fails due to a conflict then the reconcile is requeued and the
controller will see that no Job exists (as it was deleted at 2.) and
recreates it. This means we can potentially run the same Job multiple
times if we are unlucky.

So this changes removed the automatic job deletion logic from DoJob and
instead provides a new DeleteAllSucceededJobs() call to the controllers
to use to clean up the Jobs.

**This is a breaking change**

The callers should do the following to adapt:
1) After DoJob() returned ctrl.Result{}, nil, (i.e. the Job succeeded)
   the callers should store the hash of the Job in its Status
2) The caller should persist the Status.
3) If persisting the Status succeeded then call DeleteAllSucceededJobs
   providing the hash of the Jobs to be deleted.

----
We took the opportunity and did other cleanups in the Job interface that are also breaking changes:

1) The Job.preserved field is removed from the NewJob() signature
   the Job cleanup needs to be done by the client anyhow.
2) The type of the timeout parameter of NewJob changed from int (seconds) to
   time.Duration to expose the fine grain configurability directly
   instead of calling the SetTimeout() separately.

Callers should do the following:
1) Adapt to the new signature of NewJob()

----

Adaptations:
- Nova: https://github.com/openstack-k8s-operators/nova-operator/pull/108
- Placement: https://github.com/openstack-k8s-operators/placement-operator/pull/51
- Keystone: https://github.com/openstack-k8s-operators/keystone-operator/pull/112
- Glance: https://github.com/openstack-k8s-operators/glance-operator/pull/66
- Cinder: https://github.com/openstack-k8s-operators/cinder-operator/pull/61
- Neutron: https://github.com/openstack-k8s-operators/neutron-operator/pull/44
- Octavia: https://github.com/openstack-k8s-operators/octavia-operator/pull/17
- Ironic: https://github.com/openstack-k8s-operators/ironic-operator/pull/25
- MariaDB: https://github.com/openstack-k8s-operators/mariadb-operator/pull/40

### DIFF
--- a/modules/common/job/job.go
+++ b/modules/common/job/job.go
@@ -41,16 +41,14 @@ const hashAnnotation = "hash"
 func NewJob(
 	job *batchv1.Job,
 	jobType string,
-	preserve bool,
-	timeout int,
+	timeout time.Duration,
 	beforeHash string,
 ) *Job {
 
 	return &Job{
 		job:        job,
 		jobType:    jobType,
-		preserve:   preserve,
-		timeout:    time.Duration(timeout) * time.Second, // timeout to set in s to reconcile
+		timeout:    timeout,
 		beforeHash: beforeHash,
 		changed:    false,
 	}
@@ -150,12 +148,6 @@ func (j *Job) HasChanged() bool {
 // GetHash func
 func (j *Job) GetHash() string {
 	return j.hash
-}
-
-// SetTimeout defines the duration used for requeueing while waiting for the job
-// to finish.
-func (j *Job) SetTimeout(timeout time.Duration) {
-	j.timeout = timeout
 }
 
 // DeleteJob func
--- a/modules/common/job/job.go
+++ b/modules/common/job/job.go
@@ -29,10 +29,13 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"errors"
 )
+
+const hashAnnotation = "hash"
 
 // NewJob returns an initialized Job.
 func NewJob(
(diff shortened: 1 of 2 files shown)


gibizer:
I proposed PRs for all the affected operators.

modules/common/job/job.go
```diff
@@ -63,6 +64,9 @@ func (j *Job) createJob(
 		if err != nil {
 			return err
 		}
+		// Add the job hash as an annotation, this is used by the DeleteAllSucceededJobs
+		// to filter on jobs by hash
+		j.job.Annotations = util.MergeStringMaps(j.job.Labels, map[string]string{hashAnnotation: j.hash})
```
stuggi, line 69: shouldn't this be `util.MergeStringMaps(j.job.Annotations, ...` ?

modules/common/job/job.go
```diff
@@ -172,18 +160,75 @@ func DeleteJob(
...
+	h *helper.Helper,
+	jobHashes []string,
+) error {
+	jobs := &batchv1.JobList{}
+
+	err := h.GetClient().List(ctx, jobs, client.InNamespace(h.GetBefore().GetNamespace()))
+	if err != nil {
+		return err
+	}
+	h.GetBeforeObject()
```
stuggi, line 198: not sure I understand this call

gibizer (commented):
So after this fix was still able to reproduce occasional double Job execution. It turned out that we can read stale data from a cache in operator-sdk. So simply moving the deletion logic after the Job hash is persisted is not enough. @abays suggested to simply delay the Job deletion by setting `ttlSecondsAfterFinished` on the Job. So with a big enough ttl value we can circumvent this issue.

gibizer:
We have a better solution now that we see the bigger picture (i.e. stale reads). So I will propose a new PR.

PR #78, opened 2022-10-18
Default TTLSecondsAfterFinished on Jobs

gibizer:
Currently lib-common automatically and immediately deletes the Job after it finished if preserve is set to false. However
* if the caller of DoJob fails to persist the fact that the Job is done (e.g. due to conflict) OR
* the caller's next reconcile reads stale data from cache seeing that the Job is not finished but also does not exists as it was already deleted
then the client has no way to decide that the Job does not need to be run again. This could lead to duplicate Job execution.

So this patch delays the automatic Job deletion with a default TTL of 10 minutes counted from the finish of the Job. We believe that this is plenty of time to allow the caller to successfully persist the Job state even if it has to retry it and to let the operator-sdk update the cache.

See the previous [fix attempt](https://github.com/openstack-k8s-operators/lib-common/pull/77) (a partial fix) and the [nova issue](https://github.com/openstack-k8s-operators/nova-operator/issues/110) about the stale read for more details.

Closes: https://github.com/openstack-k8s-operators/nova-operator/issues/110

### DIFF
--- a/modules/common/job/job.go
+++ b/modules/common/job/job.go
@@ -81,10 +81,34 @@ func (j *Job) createJob(
 	return ctrl.Result{}, nil
 }
 
+func (j *Job) defaultTTL() {
+	// preserve has higher priority than having any kind of TTL
+	if j.preserve {
+		// so we reset TTL to avoid automatic deletion
+		j.job.Spec.TTLSecondsAfterFinished = nil
+		return
+	}
+	// if the client set a specific TTL then we honore it.
+	if j.job.Spec.TTLSecondsAfterFinished != nil {
+		return
+	}
+	// we are here as preserve is false and no TTL is set. We apply a default
+	// TTL:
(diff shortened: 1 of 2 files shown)


modules/common/job/job.go
```diff
@@ -103,10 +127,20 @@ func (j *Job) DoJob(
 		j.changed = true
 	}
 
+	// NOTE(gibi): This should be in NewJob but then the defaulting would affect
```
stuggi, line 130: could reset the TTL for the object we pass to calc the hash and restore afterwards to not have an affect on the hash.

stuggi: right its complexity at a different place. I should have done that hash calc in NewJob right from the beginning. I can follow up doing that later with providing patches to the other operators.

stuggi (approved):
/lgtm

abays (approved):
/lgtm

PR #79, opened 2022-10-18
Do not include data.Port in the keystone endpoint

fmount:
Keystone endpoint points to the `Route` created on top of the specific `Service`, which is exposed on `data.Port`.
However, having the port in the `endpoint url` let the openstack cli fail because it tries to reach the `Route` on the port specified in the url.
This change just removes the port from the generated keystone endpoint.

Signed-off-by: Francesco Pantano <fpantano@redhat.com>

### DIFF
--- a/modules/common/endpoint/endpoint.go
+++ b/modules/common/endpoint/endpoint.go
@@ -18,7 +18,6 @@ package endpoint
 
 import (
 	"context"
-	"fmt"
 	"net/url"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


fmount:
This is a follow up of [1] where the issue I saw in the glance-operator is described.
The main purpose of this change is to avoid having the `data.Port` included in the keystone endpoint that points to the `Route` which exposes the `GlanceAPI` service.

[1] https://github.com/openstack-k8s-operators/lib-common/pull/64#issuecomment-1282965000

steveb (commented):
/lgtm

stuggi (approved):
/lgtm

PR #81, opened 2022-10-24
Only try to update TTL on an existing Job

gibizer:
With #78 lib-common uses TTL on Jobs to clean them up and to support PreserveJob the TTL can be change on a Job even after it is finished. However #78 has a bug as it tries to CreateOrPatch the Job _after_ it is deleted but with an empty Job definition. This fails with

>   "error": "an empty namespace may not be set during creation"

This patch makes sure that lib-common only tries to update the TTL if the Job exists.

### DIFF
--- a/modules/common/job/job.go
+++ b/modules/common/job/job.go
@@ -170,12 +170,20 @@ func (j *Job) DoJob(
 	}
 
 	// allow updating TTLSecondsAfterFinished even after the job is finished
-	_, err = controllerutil.CreateOrPatch(ctx, h.GetClient(), job, func() error {
-		job.Spec.TTLSecondsAfterFinished = j.job.Spec.TTLSecondsAfterFinished
-		return nil
-	})
-	if err != nil {
+	job, err = GetJobWithName(ctx, h, j.job.Name, j.job.Namespace)
+	if err != nil && !k8s_errors.IsNotFound(err) {
 		return ctrl.Result{}, err
+	}
+
+	if err == nil {
+		_, err = controllerutil.CreateOrPatch(ctx, h.GetClient(), job, func() error {
+			job.Spec.TTLSecondsAfterFinished = j.job.Spec.TTLSecondsAfterFinished
+			return nil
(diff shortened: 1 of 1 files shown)


modules/common/job/job.go
```diff
@@ -170,14 +170,22 @@ func (j *Job) DoJob(
...
-		job.Spec.TTLSecondsAfterFinished = j.job.Spec.TTLSecondsAfterFinished
-		return nil
-	})
-	if err != nil {
+	job, err = GetJobWithName(ctx, h, j.job.Name, j.job.Namespace)
+	if err != nil && !k8s_errors.IsNotFound(err) {
 		return ctrl.Result{}, err
 	}
 
+	if err == nil {
```
bogdando, line 178: this condition looks redundant, err will alsways be nil at this point

modules/common/job/job.go
```diff
@@ -170,14 +170,22 @@ func (j *Job) DoJob(
...
+			job.Spec.TTLSecondsAfterFinished = j.job.Spec.TTLSecondsAfterFinished
+			return nil
+		})
+
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
 	return ctrl.Result{}, nil
```
bogdando, line 189: may be this return path can be merged with the above condition, so that at the line 174, it will just return on any err != nill?

bogdando (changes requested):

modules/common/job/job.go
gibizer: Yes, this is the meat of the patch. We should not try to update the Job with a new TTL if since the Job is deleted.

modules/common/job/job.go
gibizer: Yes, we can exit early a L174 if there is any error. But we still need L189 for the case where the Job exists, and the TTL update happened successfully at L182. So we cannot reduce the number of returns.

abays (approved):
/lgtm

PR #82, opened 2022-10-24
[StatefulSet]Allow fine grained RequeueAfter value

gibizer:
Make the RequeueAfter timeout value configurable so that in an envtest environment it can be set to a lower than 1 second value to speed up the testing.


abays (approved):
/lgtm

PR #83, opened 2022-10-25
Return by pointer for consistency

bogdando:
Signed-off-by: Bogdan Dobrelya <bogdando@yahoo.com>

### DIFF
--- a/modules/test/helpers/configmap.go
+++ b/modules/test/helpers/configmap.go
@@ -23,21 +23,21 @@ import (
 )
 
 // GetConfigMap -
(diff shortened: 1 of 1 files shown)


bogdando:
@stuggi @gibizer PTAL

bogdando:
Actually, we must reverse, see the testing
by pointer https://go.godbolt.org/z/jaccozvT6
smart https://go.godbolt.org/z/99Wjf8nqf <<< use THIS

bogdando:
this one is good to go as is. I moved the return by pointer vs value discussion elsewhere :)

gibizer (approved):
I agree to change this to pointer to have consistency across the helpers.

PR #84, opened 2022-10-26
Add missing ServiceName to StatefulSet

cschwede:
ServiceName is required if a Service should govern a StatefulSet.


abays (approved):
/lgtm

PR #85, opened 2022-10-26
Add clusterIP to Service

cschwede:
StatefulSets require a headless service to manage the network identity of the pods[1]. A headless service itself is defined by specifying "None" as clusterIP[2].

[1] k8s.io/docs/concepts/workloads/controllers/statefulset/#limitations [2] k8s.io/docs/concepts/services-networking/service/#headless-services


abays (approved):
`ClusterIP` defaults to empty string, so existing behavior is preserved while allowing for the `None` specification if so desired

/lgtm

PR #88, opened 2022-11-02
Extra volumes support to lib-common/modules/storage

fmount:
This purpose of this patch is to introduce the required data structures and functions needed by the `OpenStack` operators to mount `ExtraVolumes` other than the default ones required by each service. 
Extra volumes can be `Secrets`, `ConfigMaps` and any other k8s entity that can be mounted to pods.
The common storage module provides the basic building block for this feature, and the operator related data structures are extended to include additional parameters that can be used as conditions in the operator logic. 
The `GetVolumes` and `GetVolumeMounts` functions are able to process both `internal` (the ability to propagate an `extraVolume`
within the single operator pods) and `external propagation` (the ability to allocate and mount a volume across different storage operators (`Glance`, `Cinder`, `Manila`, `Ansible`, etc), which implements a way to filter volumes a pod is supposed to mount according to the propagation policy defined in the main `CRD`.
ExtraMounts can be defined in the global `CRD` structure: they can be k8s `Secrets`, `ConfigMaps` or other entities that should be propagated to the operators pods (e.g. `Cinder`, `Glance`, `AnsibleEE`, `OpenStack` operator and so forth).


fmount:
@abays @fultonj wondering if the best approach is to squash the commits of this change or leaving them to keep the history

fmount:
> @abays @fultonj wondering if the best approach is to squash the commits of this change or leaving them to keep the history
So given the amount of commits, I decided to squash them in a single commit and give a better description of the work.

@abays this is the first PR of a series, the operators will be able to include this definition in their CRDs and mount extra volumes.
A demo that describes how this change works is here: https://asciinema.org/a/533951

modules/storage/cinder.go
```diff
@@ -0,0 +1,76 @@
...
+// CinderExtraVolMounts exposes additional parameters processed by the cinder-operator
+// and defines the common VolMounts structure provided by the main storage module
+type CinderExtraVolMounts struct {
+	// +kubebuilder:validation:Optional
+	Name string `json:"name,omitempty"`
+	// +kubebuilder:validation:Optional
+	Region string `json:"region,omitempty"`
+	// +kubebuilder:validation:Required
+	VolMounts []VolMounts `json:"extraVol,omitempty"`
+}
```
abays, line 54: It looks like this struct is the same across `Cinder`, `Glance` and `Openstack`.  Is there a reason to not just combine them as a single struct type?  Perhaps you're envisioning more fields in the future that might be needed for one service but not another?

fultonj: It's the same today but we can imagine them differing in the future per type so that's why each service has its own version.

fultonj (approved):
I've tested this library and works as planned. It's an important next step for storage backend configuration. Thanks for this work Francesco!

abays (approved):
/lgtm

modules/storage/cinder.go
fmount: Yeah, it's true that for now these data structures are pretty much the same, and
the same thing applies to the functions related to cinder and glance.
However, in my vision I see two main points that led me to keep them separated:

1. the data structures can diverge at some point (same thing for the associated
functions); glance might add parameters to handle more granular policies while cinder
can improve the `ExtraVolType` to mount only **"trusted"** volumes; this is just an example
of two implementations that can go in a different direction;

2. my idea is to minimize the changes required to the operators in case something
   change in lib-common: if I collapse everything in a single data structure, a
   PR that improves `Glance` can break `Cinder`, or just have side effects to that
   (or any other) component that is happy with the current status of the module.
   In general `GetVolumes` and `GetVolumeMounts` are common functions for all the
   operators (so that can stay in storage.go), and all the services can "extend"
   this model and build their own logic on top of it.
   The `ansible-operator` doesn't need the extra logic provided for cinder and glance,
   and it can just rely on the basic storage feature [0], hence no need to build
   anything on top of it.

With this change I'm also trying to establish a pattern that should be
followed in future developments:

- `storage.go, utils.go` -> common functions implemented for the whole storage module
- `<operator>.go` -> it's available in the storage module namespace, and it inherits the
                              basic features specializing the behavior for a given service

By doing this I can improve my service within the storage module w/o affecting any other
component that relies to this module, but, at the same time, I can build on top of the basic
building block provided by storage.go.

[0]  https://github.com/fmount/ansibleee-operator/blob/extra_volumes/api/v1alpha1/ansibleee_types.go#L54

modules/storage/storage.go
```diff
@@ -0,0 +1,70 @@
...
+	// external-data-plane related propagation policy
+	Compute               ServiceType            = "Compute"
+)
+
+// VolMounts is the data structure used to expose Volumes and Mounts that can
+// be added to a pod according to the defined Propagation policy
+type VolMounts struct {
+	// +kubebuilder:validation:Required
+	// Propagation defines which pod should mount the volume
+	Propagation  []ExtraVolumePropagation `json:"propagation,omitempty"`
```
stuggi, line 49: should not have omitempty if it is required

modules/storage/openstack.go
```diff
@@ -0,0 +1,34 @@
...
+
+// OpenStackExtraVolMounts exposes the main parameters processed by the underlying
+// operators and defines the common VolMounts structure provided by the storage module
+type OpenStackExtraVolMounts struct {
+	// +kubebuilder:validation:Optional
+	Name string `json:"name,omitempty"`
+	// +kubebuilder:validation:Optional
+	Region string `json:"region,omitempty"`
+	// +kubebuilder:validation:Required
+	VolMounts []VolMounts `json:"extraVol,omitempty"`
```
stuggi, line 33: should not have omitempty if it is required

modules/storage/glance.go
```diff
@@ -0,0 +1,70 @@
...
+
+// GlanceExtraVolMounts exposes additional parameters processed by the glance-operator
+// and defines the common VolMounts structure provided by the main storage module
+type GlanceExtraVolMounts struct {
+	// +kubebuilder:validation:Optional
+	Name string `json:"name,omitempty"`
+	// +kubebuilder:validation:Optional
+	Region string `json:"region,omitempty"`
+	// +kubebuilder:validation:Required
+	VolMounts []VolMounts `json:"extraVol,omitempty"`
```
stuggi, line 47: should not have omit empty if it is required

modules/storage/cinder.go
```diff
@@ -0,0 +1,76 @@
...
+
+// CinderExtraVolMounts exposes additional parameters processed by the cinder-operator
+// and defines the common VolMounts structure provided by the main storage module
+type CinderExtraVolMounts struct {
+	// +kubebuilder:validation:Optional
+	Name string `json:"name,omitempty"`
+	// +kubebuilder:validation:Optional
+	Region string `json:"region,omitempty"`
+	// +kubebuilder:validation:Required
+	VolMounts []VolMounts `json:"extraVol,omitempty"`
```
stuggi, line 53: should not have omitempty if it is required

modules/storage/cinder.go
stuggi: Two questions
* Do I get it right that these operator type specific types are only used in those operators? like Cinder in cinder-operator, Glance in glance-operator, ... . If so, shouldn't those be part of the operator and the struct be part of their versioned api? * From what you said, you expect those might change in future, so I think we might want to version them like an operator api, as a module version is probably not enough? Having those in the operator itself might not required that?

modules/storage/storage.go
```diff
@@ -0,0 +1,70 @@
...
+const (
+	// PropagationEverywhere is used to define a propagation policy that allows
+	// to get the volumes mounted to all the OpenStack services
+	PropagationEverywhere ExtraVolumePropagation = "All"
+	// DBSync represents a common ServiceType defined by the OpenStack operators
+	// that keeps track of the DBSync pod
+	DBSync                ServiceType            = "DBSync"
+	// Compute represents a common ServiceType that can be translated into an
+	// external-data-plane related propagation policy
+	Compute               ServiceType            = "Compute"
```
stuggi, line 41: Is this specific for Computes, since there can be other external data plane components. would they their own  ServiceType or is an "External" ServiceType generic enough?

modules/storage/storage.go
fmount: It's "required" in the sense that from meta-operator  we need to propagate volumes to the right services (External propagation) and we might want to select, within the single operator, which pod can mount what (so we need this parameter in the CR definition), but given this structure can be used in different contexts (e.g. ansible-operator [1]) where
the assumption is to get all the defined extra volumes and mount them, we might want to avoid the parameter, hence I'm keeping it optional.

[1] https://github.com/openstack-k8s-operators/ansibleee-operator/pull/6/files#diff-b7931efcf77c2d3bcfe0479b278a92deaec40ae5a56ccd6067f52985c01d8ab7R23

modules/storage/openstack.go
fmount: Same story of  Propagation field can be applied for `Region` and `Name` (just because atm we don't have logic to
filter volumes based on Name or Region), but I agree that `VolMounts` should be required if the `ExtraMounts` are defined.
Fixing this in the next update.

modules/storage/glance.go
fmount: Yeah, fixing this in my next update.

modules/storage/cinder.go
fmount: same here

modules/storage/storage.go
fmount: For `Compute` I mean the `external-data-plane` which is supposed to see files propagated via `ansible-operator`, which is the only external component I'm aware so far.
We can change it to `External` if you think that `Compute` can result confusing and it doesn't fit well for the use case, but in theory I think the only node type outside of OCP will be Compute (managed by the jobs triggered by the `ansible-operator`).
Let me know your thoughts, this can be an easy change while we're in the middle of reviewing the lib-common/storage content.

modules/storage/cinder.go
fmount: Fair point, and you're right about thinking that this should be versioned as long as the structure evolves; we can probably reduce the maintenance over lib-common instead of growing it: I'm just not sure this struct will change too much over the time, and considering that the operators define the same structure on each service that can use it, the idea is to reduce redundancy and avoid messing up w/ the operator flow, but maybe this "versioning" work can be part of a follow up change once the feature is introduced and stable enough?
For now I know that any code change can go in a single place in lib-common, which is easier to maintain, but I can definitely experiment with it and propose a follow up as long as we have the big picture in mind.

stuggi: I agree that it the struct likely does not change too much and I like the approach. I just want to make us think about how we handle changes to the struct for the future. we need to be able to do auto api version transitions of the operator using this module and expose the ExtraMounts. since all methods/func depend on the structs, I think we could just move all to a v1beta1 package to introduce the version and keep it forever if there is no next version required.

modules/storage/glance.go
```diff
@@ -0,0 +1,70 @@
...
+
+// GlanceExtraVolMounts exposes additional parameters processed by the glance-operator
+// and defines the common VolMounts structure provided by the main storage module
+type GlanceExtraVolMounts struct {
+	// +kubebuilder:validation:Optional
+	Name string `json:"name,omitempty"`
+	// +kubebuilder:validation:Optional
+	Region string `json:"region,omitempty"`
+	// +kubebuilder:validation:Required
+	VolMounts []VolMounts `json:"extraVol"`
```
gibizer, line 47: So just to see if I understood the intention properly. A single VolMounts is a list of Volumes&Mounts with a specific propagation rule set. So with two VolMounts here we can have two sets of Volumes&Mounts that are propagated by different rules. Is it so?

modules/storage/cinder.go
```diff
@@ -0,0 +1,76 @@
...
+	// +kubebuilder:validation:Optional
+	Region string `json:"region,omitempty"`
+	// +kubebuilder:validation:Required
+	VolMounts []VolMounts `json:"extraVol"`
+}
+
+// AppendVolumeMount is a function that appends a valid Volume to the existing corev1.VolumeMount array
+func (c *CinderExtraVolMounts) AppendVolumeMount(vm []corev1.VolumeMount, extraVol []CinderExtraVolMounts, svc []ServiceType) []corev1.VolumeMount {
+
+	if len(extraVol) > 0 {
```
gibizer, line 59: This feels unnecessary. If `extraVol` is an empty list then the below for loop will be noop anyway.

modules/storage/cinder.go
```diff
@@ -0,0 +1,76 @@
...
+			vm = GetVolumeMounts(vm, extraVolumeMounts.VolMounts, svc)
+		}
+	}
+	return vm
+}
+
+// AppendVolume is a function that appends a valid Volume to the existing corev1.Volume array
+func (c *CinderExtraVolMounts) AppendVolume(vm []corev1.Volume, extraVol []CinderExtraVolMounts, svc []ServiceType) []corev1.Volume {
+
+	if len(extraVol) > 0 {
```
gibizer, line 70: ditto

modules/storage/cinder.go
fmount: Right, that's a good point we were discussing w/ @gibizer in other comments.
I think we have a couple of choices here:

1. versioning the lib-common module (as you mentioned in your previous comment with the v1beta package part)
2. Moving the operators specific implementation in the operators itself

The advantage of doing .2 is that we can version a `CinderExtraVolMounts` structure within `cinder/api/v1beta/cinder_types.go`,  but we're splitting the complexity between lib-common and cinder-operator, which logically makes sense, just not sure if it's be best path we can take.
I think in general the versioning is a problem in all  lib-common modules: if you break keystone you end up breaking all the operators that use it, unless you explicitly  downgrade lib-common in go.mod: that's why I would just think more about the problem and propose a pattern that solves this potential problem globally.

modules/storage/storage.go
fmount: Yes, that works, I'll do in the next iteration.

modules/storage/storage.go
fmount: At this stage Compute works, but I agree we should align and follow up here according to the evolution of this work: `External` might not be a good fit if we still need to propagate volumes in a more granular way, but of course, given that `Nova` can adopt this method, we should address the requirements of that use case.

modules/storage/storage.go
```diff
@@ -0,0 +1,73 @@
...
+	// +kubebuilder:validation:Optional
+	ExtraVolType ExtraVolType             `json:"extraVolType,omitempty"`
+	// +kubebuilder:validation:Required
+	Volumes      []corev1.Volume          `json:"volumes"`
+	// +kubebuilder:validation:Required
+	Mounts       []corev1.VolumeMount     `json:"mounts"`
+}
+
+// GetVolumes is a function that takes extraVolumes and try to append to the corev1.Volume list
+func GetVolumes(vm []corev1.Volume, exvol []VolMounts, propagationList []ServiceType) []corev1.Volume {
```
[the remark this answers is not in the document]
fmount: ack, I'm fine about using whatever is best and make it more clear.

modules/storage/openstack.go
```diff
@@ -0,0 +1,34 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package storage
+
+const (
+	// OpenStack is the global ServiceType that refers to all the components deployed
+	// by the meta-operator
+	OpenStack ServiceType = "Openstack"
```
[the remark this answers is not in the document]
fmount: Yes, I've just imagined a world where `meta-operator` can propagate volumes to the services it orchestrates: this is supposed to be used at that level in case we don't want to specify each service:

```
Propagation:
- Glance
- Cinder
- Compute
```
vs: 

```
Propagation:
- OpenStack
```

but in the long term vision it can be different by `Everywhere` to cover multiple use cases.
If it helps, we can keep it simple and remove it for now, as long as the whole `openstack.go` that can be moved to the meta-operator and get versioned along with the defined API.

modules/storage/utils.go
```diff
@@ -0,0 +1,72 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+// canPropagate is a utility function that checks if a Volume meet the propagation policy
+func canPropagate(prop ExtraVolumePropagation, svcs []ServiceType) bool {
+	if prop == PropagationEverywhere {
+		return true
+	}
+	for _, service := range svcs {
+		if ServiceType(prop) == service {
```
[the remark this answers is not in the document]
fmount: I think so, I'm just thinking about having a unified approach for propagation.
We currently have different types (const mostly), pointing to strings, at least:
- ServiceType
- ExtraVolumePropagation

We can simplify that code (which can result cleaner), just having a single, `PropagationType` which defines the propagation behavior.

modules/storage/storage.go
```diff
@@ -0,0 +1,73 @@
...
+	Propagation  []ExtraVolumePropagation `json:"propagation,omitempty"`
+	// Label associated to a given extraMount
+	// +kubebuilder:validation:Optional
+	ExtraVolType ExtraVolType             `json:"extraVolType,omitempty"`
+	// +kubebuilder:validation:Required
+	Volumes      []corev1.Volume          `json:"volumes"`
+	// +kubebuilder:validation:Required
+	Mounts       []corev1.VolumeMount     `json:"mounts"`
+}
+
```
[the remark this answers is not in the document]
fmount: Different k8s types with different calls made by operators (you can observe a pattern in all the operators, e.g.,  [1]), so ... I'd say we `could` have a unified function for both, but I'd like to coordinate this work (a follow up?) from the operators to lib-common, and not the reverse.
Even though I think you're point is a good one, I'd like to avoid a massive change in the operators, while having CI, API versioning, and things like that would help having smooth changes w/o introducing bugs.

[1] https://github.com/fmount/cinder-operator/blob/master/pkg/cinder/volumes.go

modules/storage/glance.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// AppendVolumeMount is a function that appends a valid Volume to the existing corev1.VolumeMount array
+func (g *GlanceExtraVolMounts) AppendVolumeMount(vm []corev1.VolumeMount, extraVol []GlanceExtraVolMounts, svc []ServiceType) []corev1.VolumeMount {
+
+	if len(extraVol) > 0 {
```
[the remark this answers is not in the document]
fmount: Not sure I recall why I added this there, but yes, if length is zero there are no iterations

modules/storage/glance.go
fmount: Exactly, your understanding is correct:
```
...
...
  extraMounts:
      extraVol:
        - propagation:
          - Cinder
          - Compute
          volumes:
          - name: ceph
            projected:
              sources:
              - secret:
                  name: ceph-client-conf
              - secret:
                  name: ceph-client-conf2
          mounts:
          - name: ceph
            mountPath: "/etc/ceph"
            readOnly: true
        - propagation:
          - Glance
          - Compute
          extraVolType: Ceph
          volumes:
          - name: ceph2
            projected:
              sources:
              - secret:
                  name: ceph-client-conf
          mounts:
          - name: ceph2
            mountPath: "/etc/ceph2"
            readOnly: true
```

Two set of volumes with two different propagation rules.

modules/storage/cinder.go
fmount: +1

modules/storage/utils.go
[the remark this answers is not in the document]
gibizer: It really depends on what type of rules we want to support.

If we only need exact string match as a rule, like in the current PRs, then having a single type make sense to me.

gibizer (commented):
I have two general questions:
1. do we need a way to limit the mount path? I mean do we really want to allow mounting things to `/etc/passwd` path or to `/sys`? 

2. I'm not sure I found the place where we implement the propagation by instance name like `volume1`. Also what happens if the name of the instance matches with some pre-defined constants. E.g.: Somebody creates a `NovaCell` CR named `All`? Do we need distinguish between the constants like `All`, `DbSync` and between a case when the propagation rule wants to target an instance by its name? I.e. can we the later changed to use `<Kind>:<Name>` format, like `CinderVolume:volume1` and `NovaCell:All`

modules/storage/openstack.go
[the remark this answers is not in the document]
fmount: +1 yes, not used atm, I'm fine removing it as `All` is basically the same thing atm.
When we have a clear idea of the orchestration logic, we can improve the propagation related filters accordingly.

modules/storage/utils.go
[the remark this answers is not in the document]
fmount: +1 agreed here, I'll simplify this part in my next iteration and update the PR accordingly.

modules/storage/storage.go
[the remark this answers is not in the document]
fmount: Ok, I think I see your point, which is actually interesting: let me think a bit about it and I'll try to propose such a
valid update that cover this part.
Keep in mind that, in the current status (for the purpose of introducing the feature), I'm not looking into revisiting
the operators structure (e.g. have a single `GetVolume` to rule them all), so the code you see here is the result
of extending the operator existing functions and keep this logic into `lib-common`.
We can choose what's best to keep in `lib-common` and what is supposed to be operator related logic: the cleaner
we design this behavior, the better.
As per your comment here, we can have only `1 propagation function`, which actually returns a `VolMounts` array
that represents the list of `filtered VolMounts` (which contains `Volumes` and `VolumeMounts`): the resulting structure
can be merged w/ the current struct explicitly provided at the operator level.

Let me try an example to better clarify.
At `Glance` level (in `lib-common/modules/storage/glance.go`) I can have something like:

```
func (g *GlanceExtraVolMounts) Propagate(svc []PropagationType) []VolMounts {
...
...
}
```

which calls, behind the scenes, `storage.go` (that still represents the basic building block for Propagation), and,
according to the propagation list I can return a list of `VolMounts` that can be applied to the pods.

`storage.go` can be changed and  should have something like:

```
func (v *VolMounts) Propagate(svc []PropagationType) []VolMounts {

	var vl []VolMounts

	for _, p := range v.Propagation {
		if canPropagate(p, svc) {
			vl = append(vl, VolMounts {
				Volumes: v.Volumes,
				Mounts: v.Mounts,
			})
		}
	}

	return vl
}
```

and Glance can actually build/filter on top of the basic propagation function (e.g., processing additional parameters
in the future).
Similar thing should happen with Cinder, I used Glance in this example because it's easier.

The operator side of things (e.g., `pkg/glance/volumes.go`), however, should be something like:

```
        // for any GlanceExtraVolumes struct, propagate the VolMounts
	for _, exv := range extraVol {
		for _, vol := range exv.Propagate(svc) {
			vm = append(vm, vol.Mounts...)
		}
	}
```

`VolMounts` is an array, and the propagation functions are applied/called by the defined struct, hence an
iteration over there is necessary.
This approach simplify **a lot** the amount of code in the `lib-common` module, and still gives the opportunity
to build (in the future) on top of this feature.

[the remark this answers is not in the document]
gibizer: I feel like we are converging to the same idea.

modules/storage/go.mod
```diff
@@ -2,11 +2,27 @@ module github.com/openstack-k8s-operators/lib-common/modules/storage
 
 go 1.18
 
-require github.com/onsi/gomega v1.20.2
+require (
+	github.com/onsi/gomega v1.20.2
```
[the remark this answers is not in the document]
fmount: removing it, it's actually not used in the ceph/ submodule

modules/storage/openstack.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+// OpenStackExtraVolMounts exposes the main parameters processed by the underlying
+// operators and defines the common VolMounts structure provided by the storage module
+type OpenStackExtraVolMounts struct {
+	// +kubebuilder:validation:Optional
+	Name string `json:"name,omitempty"`
+	// +kubebuilder:validation:Optional
+	Region string `json:"region,omitempty"`
```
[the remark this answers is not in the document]
fmount: Yes, they're eventually needed to perform additional filtering on top of the basic propagation we provide. I'm really not sure these parameters are needed at this stage (because we're not using them), but I see:

1. potential work that can be done when at some point openstackcontrolplane will support multiple deployments
2. it's an example that shows how an operator related data structure can be extended

modules/storage/glance.go
```diff
@@ -0,0 +1,70 @@
...
+
+const (
+	// GlanceExtraVolTypeUndefined can be used to label an extraMount which
+	// is not associated with a specific backend
+	GlanceExtraVolTypeUndefined ExtraVolType = "Undefined"
+	// GlanceExtraVolTypeCeph can be used to label an extraMount which
+	// is associated to a Ceph backend
+	GlanceExtraVolTypeCeph      ExtraVolType = "Ceph"
+
+	// GlanceAPI defines the glance-api group
```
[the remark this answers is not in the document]
fmount: done, we can converge to `PropagationType` because we're doing string comparison, no need to have multiple types that are strings and refer to the same pattern

modules/storage/storage.go
```diff
@@ -0,0 +1,73 @@
...
+// ServiceType identifies the Service, Group or instance (e.g. the backend) that
+// receives an Extra Volume that can potentially be mounted
+type ServiceType string
+
+const (
+	// PropagationEverywhere is used to define a propagation policy that allows
+	// to get the volumes mounted to all the OpenStack services
+	PropagationEverywhere ExtraVolumePropagation = "All"
+	// DBSync represents a common ServiceType defined by the OpenStack operators
+	// that keeps track of the DBSync pod
```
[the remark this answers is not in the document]
fmount: No, it's the opposite actually: we can keep track of the `caller` in the operator, like the following example (taken w/o any context from Glance):
```
VolumeMounts: GetVolumeMounts([]storage.GlanceExtraVolMounts{}, []storage.PropagationType{storage.DBSync}),
```
and we can make sure we don't propagate volumes to pods that don't need them.
`DBSync` is a Job that creates a pod that doesn't need any of these extra volumes, that's why we also defined this group that can skip any `extraVolumes` propagation.

fmount:
> I have two general questions:
> 
>     1. do we need a way to limit the mount path? I mean do we really want to allow mounting things to `/etc/passwd` path or to `/sys`?

No, there isn't any logic to prevent people to "blacklist" a set of forbidden mountpoints: I understand the concern of having the ability to mount a volume at any point, and we can probably solve this in a follow up change where we can define a blacklist array.
In k8s, if you define a pod adding a volume to a `/sys` or `/etc/passwd mountpoint`, no logic prevents to do that (or it's just my lack on knowledge on this area): exposing this set of structures on a different `CRD` is basically not adding any overlay on top of what k8s provides, but I agree we can introduce such a feature that prevents people to mount volumes everywhere.

> 
>     2. I'm not sure I found the place where we implement the propagation by instance name like `volume1`. Also what happens if the name of the instance matches with some pre-defined constants. E.g.: Somebody creates a `NovaCell` CR named `All`? Do we need distinguish between the constants like `All`, `DbSync` and between a case when the propagation rule wants to target an instance by its name? I.e. can we the later changed to use `<Kind>:<Name>` format, like `CinderVolume:volume1` and `NovaCell:All`

This is definitely a good point, we can introduce that kind of format if it helps covering a corner case where people are creating Cells with arbitrary names that can conflicts with the pre-defined constants. However, I'd like to do such a change in a
second iteration, where we can focus on consolidating this convention instead of a mixed number of changes where the purpose is to introduce the basics of this feature.
In the current status, creating something like this in the operator:

```
instance :=  storage.GlanceAPI + ":" + <instance_name_from_the_defined_crd>
... []storage.ServiceType{storage.Glance, storage.GlanceAPI, instance})
```

allows to work with that pattern, but we lack of additional checks in lib-common that can be part of an additional PR
In summary, I'd like to move forward with this change, and later improve the pattern used to process Global/Groups/instance service types.

modules/storage/cinder.go
Akrog: I think it would be better not having the `<operator>.go` modules here, as it seems conceptually wrong to have a cinder specific structure residing outside of the cinder-operator.

fmount: ack, and actually, as we were discussing w/ @stuggi we should have them in the operator api/ section, so that this structure can be properly versioned: the advantage is to avoid a change in lib-common that can potentially break the operators code.
I'll try this in my next iteration in a bit.

fmount: @Akrog @stuggi  in the last update here I removed the `<operator>.go` data structures which are now versioned as part of the operator(s) API (see [1][2][3]).
I guess we can now test all this part and move forward with the lib-common change.

[1] https://github.com/openstack-k8s-operators/cinder-operator/pull/65/files#diff-4c0f5ae542ebffc716c3eb8a28926941a358007b4bac2a3efd262c173578685aR198
[2] https://github.com/openstack-k8s-operators/glance-operator/pull/75/files#diff-f462211fddba68b6fd25298dffed9956b8b2613a6cf7d3e55f95e59a596fc0c7R191
[3] https://github.com/openstack-k8s-operators/openstack-operator/pull/38/files#diff-e003a80b7e277d733db592a080dfad3eaa21c77b28dc5df577166743b54c898bR178

fmount:
@gibizer  @Akrog @stuggi @fultonj this change leaves space for improvements, and we can follow up
 on how things can be optimized.

I'll summarize here the current status of this PR and the potential interesting things we can do next:

1. The operators related data structures are moved and versioned in the API
   (within each operator);

2. the main `storage.go` module represents the basic building block, providing the `optional` Propagation
   feature and the related basic function. Propagation is nothing more than the ability to mount volumes to
   specific `Services` or `instances` (according to the `CR` definition) and some operators (e.g., the
   `ansibleee-operator`) can just ignore the propagation policy and mount whatever is defined at the top
   level CR, otherwise, it will just mount what is supposed to be propagated to the external nodes (see [1])

4. The storage module exposes both Volumes and VolumeMounts, and they're processed together within
     the same function to make sure a pair of (Volume, VolumeMount) is returned

What we can do next:

1. Define a blacklist (or something that gives that idea) for not allowed paths: as per my conversation with
   @gibizer  it would be great preventing people to mount volumes on specific mountpoints (e.g., sys/ or
    /etc/passwd), but this is something that should be discussed and properly designed (for instance we
   need to make sure if it's something that can be filtered in advance at the operator level or a common
   pattern we'd like to tie into lib-common)

2. We can improve the way we filter volumes and define a more complex and robust syntax to filter
     instances: e.g., we can have "CinderVolume:<instance>" instead of just "<instance>" when we do
     propagation, so that an instance is explicitly tied to a Service group. Since we're talking about string
     matching, this syntax already works, but it would be great introducing a validation over the propagation
     allowed syntax


Being that said, unless you have reasons to don't merge, I'd like to move forward at least with this change so
we can merge the ansible-operator PR before the POC (which is based on this work for the ceph integration
part.


FYI @jlarriba @odyssey4me

[1] https://asciinema.org/a/533951

stuggi:
I am good with this to land if @gibizer @Akrog @fultonj are and follow up on the ideas which came up during the discussion here.

modules/storage/storage.go
```diff
@@ -0,0 +1,84 @@
...
+const (
+	// PropagationEverywhere is used to define a propagation policy that allows
+	// to get the volumes mounted to all the OpenStack services
+	PropagationEverywhere PropagationType = "All"
+	// DBSync represents a common ServiceType defined by the OpenStack operators
+	// that keeps track of the DBSync pod
+	DBSync PropagationType = "DBSync"
+	// Compute represents a common ServiceType that can be translated into an
+	// external-data-plane related propagation policy
+	Compute PropagationType = "Compute"
```
Akrog, line 40: nit: I think a "CinderConsumer" would be nice to have in the future.

Akrog (approved):
I tested the lib-common, glance-operator- cinder-operator, and openstack-operator PRs and they worked as expected with the Ceph use case.

I'm ok with merging the code and improve it afterwards as we see fit.

gibizer (approved):
@fmount thanks it looks good to me.

jlarriba (approved):
I think the best approach is to have the basic functionality now so we can showcase it in the EoY demo and do more polishing/adjusting to specific needs later on.

stuggi (approved):
/lgtm

PR #91, opened 2022-11-22
Fix logging if Secret not found

cschwede:
GetSecret returns nil as secret if no Secret is found, and this raises a panic in util.logObjectParams with a runtime error.

Also changed the log message to be more inline with other similar messages.


stuggi (approved):
/lgtm

PR #107, opened 2022-12-07
Introduce operator-lint

gibizer:
As lib-common now has test helpers using gomega we enable operator-lint that has one important check for gomega code T001.

[1] https://github.com/gibizer/operator-lint/


stuggi (approved):
/lgtm

PR #117, opened 2022-12-15
Generic instance update func

abays:
Intended to be used with our deferred-instance-patch-at-end-of-reconcile pattern

### DIFF
--- a/modules/common/helper/helper.go
+++ b/modules/common/helper/helper.go
@@ -17,6 +17,7 @@ limitations under the License.
 package helper
 
 import (
+	"context"
 	"encoding/json"
 
 	"github.com/go-logr/logr"
@@ -28,6 +29,9 @@ import (
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 // Helper is a utility for ensuring the proper patching of objects.
@@ -162,6 +166,67 @@ func (h *Helper) calculateChanges(after client.Object) (map[string]bool, error) 
 		res[key] = true
 	}
 	return res, nil
+}
+
+// PatchInstance - Patch an instance's metadata and/or status if they have changed (based
+// on comparison with its old state as represented in the helper.Helper)
+//
+// NOTE: This function is mainly intended for use in Podified Operators with their
+// deferred-instance-persistence pattern within the reconcile loop.  If used in this manner,
+// any error returned by this function should be set as the error returned by the encompassing
+// Reconcile(...) function so that the error from PatchInstance is properly propagated.
+//
+// Example:
+//
+// func (r *SomeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, _err error) {
+//     ...
+//     defer func() {
+//         ...
+//         err := instance.PatchInstance(ctx, h, instance)
+//         if err != nil {
+//             _err = err
+//             return
+//         }
+//     }
+//     ...
+// }
+//
+func (h *Helper) PatchInstance(ctx context.Context, instance client.Object) error {
+	var err error
+
+	l := log.FromContext(ctx)
+
+	if err = h.SetAfter(instance); err != nil {
+		l.Error(err, "Set after and calc patch/diff")
+		return err
+	}
+
+	changes := h.GetChanges()
+	patch := client.MergeFrom(h.GetBeforeObject())
+
+	if changes["metadata"] {
+		err = h.GetClient().Patch(ctx, instance, patch)
+		if k8s_errors.IsConflict(err) {
+			l.Info("Metadata update conflict")
+			return err
+		} else if err != nil && !k8s_errors.IsNotFound(err) {
+			l.Error(err, "Metadate update failed")
+			return err
+		}
+	}
+
+	if changes["status"] {
+		err = h.GetClient().Status().Patch(ctx, instance, patch)
+		if k8s_errors.IsConflict(err) {
+			l.Info("Status update conflict")
+			return err
+
+		} else if err != nil && !k8s_errors.IsNotFound(err) {
+			l.Error(err, "Status update failed")
+			return err
+		}
+	}
+	return nil
 }
 
 // ToUnstructured - convert to unstructured


modules/common/instance/instance.go
```diff
@@ -0,0 +1,85 @@
...
+func PatchInstance(ctx context.Context, h *helper.Helper, instance client.Object) error {
+	var err error
+
+	if err = h.SetAfter(instance); err != nil {
+		util.LogErrorForObject(h, err, "Set after and calc patch/diff", instance)
+		return err
+	}
+
+	changes := h.GetChanges()
+	patch := client.MergeFrom(h.GetBeforeObject())
```
bogdando, line 60: I understand this is just a refactoring (moving the code around), but wouldn't it make sense to adjust it a bit by avoiding that patch calculated twice - here, and when we do calculate changes inside of SetAfter https://github.com/abays/lib-common/blob/feb92dd06374e5e2591ef53780b8fe788c60ca46/modules/common/helper/helper.go#L147 ?

we can store patch in the helper and just pick it here. That would also help to avoid conflicts when forming patches at different time moments, so that the most recent one could be bease on the no longer actual GetChanges result?  And the problems with applying it, would shift to the end of the patching, which looks more precise...

bogdando (dismissed):

modules/common/instance/instance.go
```diff
@@ -0,0 +1,85 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// PatchInstance - Patch an instance's metadata and/or status if they have changed (based
+// on comparison with its old state as represented in the helper.Helper)
+//
+// NOTE: This function is mainly intended for use in Podified Operators with their
+// deferred-instance-persistence pattern within the reconcile loop.  If used in this manner,
+// any error returned by this function should be set as the error returned by the encompassing
+// Reconcile(...) function.
```
gibizer, line 34: Thanks. I have a small addition:
```
any error returned by this function should be set as the error returned by the encompassing
// Reconcile(...) function so that the error from PatchInstance is properly propagated.
```

modules/common/instance/instance.go
gibizer: `client.MergeFrom()`  only defines the _base_ of the patch, but does not do the actual diffing. The second diff calculation actually happens inside the `h.GetClient().Patch()` call. As that is out of our control I'm not sure how can we reduce the number of patch calculations.

gibizer (changes requested):
The `instance` is a pretty generic name for the package. I would call this `reconciler` instead. That also helps showing the intention that PatchInstance should be use in a Reconciler.

stuggi:
@abays @gibizer wondering if it would make sense to have that as part of the helper, so that we call helper.Patch() to patch the object the helper is for ?

stuggi:
> @abays @gibizer wondering if it would make sense to have that as part of the helper, so that we call helper.Patch() to patch the object the helper is for ?

ok, @abays just told me that there is some import cycle then because of the util.log... we could update util/log.go to pass in the logger from the caller ... but that requires changes all over the place where we use the logger funcs. use basic logging in patching could be an alternative. 

seems we need to put some more thinking in our logging. I wanted to get back to that and make that better to have e.g. some log level ... but other prios ...

if we agree that this patching func is better located in the helper as it is tightly coupled, we might want to go with basic logging and update that when we got back to logging in general, instead of relocate the patchInstance later ...

gibizer:
> > @abays @gibizer wondering if it would make sense to have that as part of the helper, so that we call helper.Patch() to patch the object the helper is for ?
> 
> ok, @abays just told me that there is some import cycle then because of the util.log... we could update util/log.go to pass in the logger from the caller ... but that requires changes all over the place where we use the logger funcs. use basic logging in patching could be an alternative.
> 
> seems we need to put some more thinking in our logging. I wanted to get back to that and make that better to have e.g. some log level ... but other prios ...
> 
> if we agree that this patching func is better located in the helper as it is tightly coupled, we might want to go with basic logging and update that when we got back to logging in general, instead of relocate the patchInstance later ...

What if we introduce a new `LogForObject` variant that takes a logger instead of the Helper obj. And first use it in `Patch` when it moved to `Helper`.  But also deprecate the current `LogForObject` variant and gradually change the operators to use the new variant.

I also see a generic pattern here. `Helper` is a collection of loosely couple stuff: logger, clients, finalizers, an old copy of the instance, etc. We are passing a `Helper` instance to everywhere. So we basically couple everything to the Helper even if only part of the Helper is needed in a certain place.  This creates unnecessary coupling. If I could I would split up the Helper to smaller pieces:
1. The logic that stores the old instance and calculates the diff for patching. This would only be needed for the actual PatchInstance() call as we are moving all our instance update to that single place. So this part of the helper does not need to be passed to function called Reconcile.
2. The rest of the stored fields and Getters can be split out to independent functions or even removed and let a Reconcile classes hold the fields as it already hold some of them like clients. E.g. `GetFinalizer()` can be rewritten to  `GetFinalizer(obj client.Object, client client.Client)` as it does not depend on any other part of the `Helper`.

stuggi:
> > > @abays @gibizer wondering if it would make sense to have that as part of the helper, so that we call helper.Patch() to patch the object the helper is for ?
> > 
> > 
> > ok, @abays just told me that there is some import cycle then because of the util.log... we could update util/log.go to pass in the logger from the caller ... but that requires changes all over the place where we use the logger funcs. use basic logging in patching could be an alternative.
> > seems we need to put some more thinking in our logging. I wanted to get back to that and make that better to have e.g. some log level ... but other prios ...
> > if we agree that this patching func is better located in the helper as it is tightly coupled, we might want to go with basic logging and update that when we got back to logging in general, instead of relocate the patchInstance later ...
> 
> What if we introduce a new `LogForObject` variant that takes a logger instead of the Helper obj. And first use it in `Patch` when it moved to `Helper`. But also deprecate the current `LogForObject` variant and gradually change the operators to use the new variant.
> 
> I also see a generic pattern here. `Helper` is a collection of loosely couple stuff: logger, clients, finalizers, an old copy of the instance, etc. We are passing a `Helper` instance to everywhere. So we basically couple everything to the Helper even if only part of the Helper is needed in a certain place. This creates unnecessary coupling. If I could I would split up the Helper to smaller pieces:
> 
> 1. The logic that stores the old instance and calculates the diff for patching. This would only be needed for the actual PatchInstance() call as we are moving all our instance update to that single place. So this part of the helper does not need to be passed to functions called from Reconcile.
> 2. The rest of the stored fields and Getters can be split out to independent functions or even removed and let a Reconcile classes hold the fields as it already hold some of them like clients. E.g. `GetFinalizer()` can be rewritten to  `GetFinalizer(obj client.Object, client client.Client)` as it does not depend on any other part of the `Helper`.

Its a good idea to add a new variant for LogForObject by doing a log package and have all the future logging stuff there.
Thats what we should do as part or as a prio step for this PR.

Re the other split, probably best to discuss that again after the break. GetFinalizer() is not meant to get the finalizers of
the object. It is to have the finalizer string for the object created during initialization and when add/remove/check if it
is on object can be used. But lets have that conversation later.

abays:
> Its a good idea to add a new variant for LogForObject by doing a log package and have all the future logging stuff there.
> Thats what we should do as part or as a prio step for this PR

Let's do that as a separate PR first.  Who wants to do it? :)

modules/common/instance/instance.go
abays: Ack, done

gibizer:
> > Its a good idea to add a new variant for LogForObject by doing a log package and have all the future logging stuff there.
> > Thats what we should do as part or as a prio step for this PR
> 
> Let's do that as a separate PR first. Who wants to do it? :)

I will make a quick stab of it.

gibizer:
> > > Its a good idea to add a new variant for LogForObject by doing a log package and have all the future logging stuff there.
> > > Thats what we should do as part or as a prio step for this PR
> > 
> > 
> > Let's do that as a separate PR first. Who wants to do it? :)
> 
> I will make a quick stab of it.

https://github.com/openstack-k8s-operators/lib-common/pull/125 I try to take an approach which is more aligned with the existing logr and context interfaces. But I provided some compatibility methods so we can discuss our options. I will push a nova change to show the real usage of it.

gibizer:
And here is the nova example https://github.com/openstack-k8s-operators/nova-operator/pull/206

gibizer:
Testing of #125 (https://github.com/openstack-k8s-operators/lib-common/pull/125#issuecomment-1357923783) shows that we don't need any fancy logging interface massaging as the logger from the ctx already has all the information we want in the log.

modules/common/instance/instance.go
```diff
@@ -0,0 +1,85 @@
...
+//         }
+//     }
+//     ...
+// }
+//
+func PatchInstance(ctx context.Context, h *helper.Helper, instance client.Object) error {
+	var err error
+
+	if err = h.SetAfter(instance); err != nil {
+		util.LogErrorForObject(h, err, "Set after and calc patch/diff", instance)
```
gibizer, line 55: all the logging can be done via `log.FromContext(ctx).Error` or `log.FromContext(ctx).Info` without loosing any instance information from the log as the logger in the ctx already has the instance information attached. See https://github.com/openstack-k8s-operators/lib-common/pull/125#issuecomment-1357923783

abays: Awesome.  Then we can use that to avoid the import cycle while moving this into the `helper` package.

modules/common/instance/instance.go
bogdando: Thank you for explaining. Perhaps we still could avoid doing a 2nd call for ``client.MergeFrom()``

modules/common/instance/instance.go
bogdando: https://github.com/openstack-k8s-operators/lib-common/pull/125

modules/common/instance/instance.go
gibizer: I dont see a second call to `h.GetChanges` i only see one here.

modules/common/instance/instance.go
```diff
@@ -0,0 +1,85 @@
...
+		return err
+	}
+
+	changes := h.GetChanges()
+	patch := client.MergeFrom(h.GetBeforeObject())
+
+	if changes["metadata"] {
+		err = h.GetClient().Patch(ctx, instance, patch)
+		if k8s_errors.IsConflict(err) {
+			util.LogForObject(h, "Metadata update conflict", instance)
```
gibizer, line 65: Lets change this to
```golang
l := log.FromContext(ctx)
l.Info(...) 
```
The instance related information is already stored in and emitted by that logger

modules/common/instance/instance.go
```diff
@@ -0,0 +1,85 @@
...
+			return err
+		} else if err != nil && !k8s_errors.IsNotFound(err) {
+			util.LogErrorForObject(h, err, "Metadate update failed", instance)
```
gibizer, line 68:
```
l.Error()
```

modules/common/instance/instance.go
```diff
@@ -0,0 +1,85 @@
...
+			return err
+		}
+	}
+
+	if changes["status"] {
+		err = h.GetClient().Status().Patch(ctx, instance, patch)
+		if k8s_errors.IsConflict(err) {
+			util.LogForObject(h, "Status update conflict", instance)
```
gibizer, line 76:
```
l.Info()
```

modules/common/instance/instance.go
```diff
@@ -0,0 +1,85 @@
...
+			return err
+
+		} else if err != nil && !k8s_errors.IsNotFound(err) {
+			util.LogErrorForObject(h, err, "Status update failed", instance)
```
gibizer, line 80:
```
l.Error()
```

gibizer (changes requested):

abays:
> The `instance` is a pretty generic name for the package. I would call this `reconciler` instead. That also helps showing the intention that PatchInstance should be use in a Reconciler.

@gibizer With the new logging approach, which would remove the import cycle loop and thus allow this code to exist in the `helper` package, would you still recommend moving this code to a separate `reconciler` package?

gibizer:
> > The `instance` is a pretty generic name for the package. I would call this `reconciler` instead. That also helps showing the intention that PatchInstance should be use in a Reconciler.
> 
> @gibizer With the new logging approach, which would remove the import cycle loop and thus allow this code to exist in the `helper` package, would you still recommend moving this code to a separate `reconciler` package?

Lets move it to the `helper` package with a `Helper` receiver. Even if we blow up that `Helper` struct later the PatchInstance will be close to the struct that holds the `before` copy of the instance (currently stored in the Helper).

gibizer (approved):
Looks good to me. Thanks @abays!

modules/common/instance/instance.go
abays: L199 should result in `h.changes` being set and then we acquire those cached changes at L204.  So it seems like we already are using the stored result.  Perhaps I am misunderstanding what you mean.

stuggi (approved):
/lgtm

modules/common/instance/instance.go
bogdando: I was confused by client.MergeFrom(), sorry

PR #125, opened 2022-12-19
Introduce the log package

gibizer:
This package is intended to server as the common logging interface for the openstack operators.

### DIFF
--- a/modules/common/log/log.go
+++ b/modules/common/log/log.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2022 Red Hat
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package log
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+func defaultParams(object metav1.Object) []interface{} {
+	return []interface{}{
+		"ObjectType", fmt.Sprintf("%T", object),
+		"ObjectNamespace", object.GetNamespace(),
+		"ObjectName", object.GetName()}
+}
+
+// Info logs a non-error message to the logger with the default object
+// parameters ObjectType, ObjectNamespace, ObjectName
+func Info(
+	logger logr.Logger,
+	msg string,
+	object metav1.Object,
+	params ...interface{},
+) {
+	params = append(params, defaultParams(object)...)
+	logger.Info(msg, params...)
+}
+
+// Error logs an error message to the logger with the default object
+// parameters ObjectType, ObjectNamespace, ObjectName
+func Error(
+	logger logr.Logger,
+	err error,
+	msg string,
+	object metav1.Object,
+	params ...interface{},
+) {
+	params = append(params, defaultParams(object)...)
+	logger.Error(err, msg, params...)
+}
+
+// IntoContext returns a new context with a logger that always logs the default
+// object parameters ObjectType, ObjectNamespace, ObjectName
+//
+// The intended use is to upgrade the context at the start of the
+// Reconcile call and use the new ctx to pass arond and use the embeded
+// logger:
+//
+//    func (r *FooReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+//        ctx = log.IntoContext(ctx, req, "Foo")
+//        // ...
+//        log.FromContext(ctx).Info("Succesfully reconciled")
+//        return ctrl.Result{}, nil
+//    }
+//
+func IntoContext(ctx context.Context, req ctrl.Request, kind string) context.Context {
+	return log.IntoContext(
+		ctx,
+		log.FromContext(
+			ctx,
+			"ObjectType", kind,
+			"ObjectNamespace", req.Namespace,
+			"ObjectName", req.Name,
+		),
+	)
+}
+
+// FromContext returns the logger stored in the context
+func FromContext(ctx context.Context) logr.Logger {
+	return log.FromContext(ctx)
+}


gibizer:
There is an example PR that partially switches over nova to the new interface https://github.com/openstack-k8s-operators/nova-operator/pull/206

gibizer:
It turns out that the this whole thing is not necessary if we use the logger from the ctx directly as it already has the extra fields added to it about the instance that is being reconciled. Here is an example log from this patch applied to nova.
```
2022-12-19T17:21:17.306+0100	INFO	Reconciling	{"controller": "nova", "controllerGroup": "nova.openstack.org", "controllerKind": "Nova", "Nova": {"name":"nova","namespace":"openstack"}, "namespace": "openstack", "name": "nova", "reconcileID": "154c6142-0dfe-4587-a655-b91f566bd131", "ObjectType": "Nova", "ObjectNamespace": "openstack", "ObjectName": "nova"}
```
The fields that is added automatically:
- "controller": "nova"
- "controllerGroup": "nova.openstack.org"
- "controllerKind": "Nova"
- "Nova": {"name":"nova","namespace":"openstack"}
- "namespace": "openstack"
- "name": "nova"
- "reconcileID": "154c6142-0dfe-4587-a655-b91f566bd131"
The fields we add via lib-common:
- "ObjectType": "Nova"
- "ObjectNamespace": "openstack"
- "ObjectName": "nova"

So clearly what we try to do here is redundant and already there if we use the logger from the context we get as the parameter fo the Reconcile call. So a simple `log.FromContext(ctx)` is enough in each operators to have structural logging with the information we need. Also ctx can be passed to lib-common so lib-common can do the same if logging is needed.

stuggi:
> It turns out that the this whole thing is not necessary if we use the logger from the ctx directly as it already has the extra fields added to it about the instance that is being reconciled. Here is an example log from this patch applied to nova.
> 
> ```
> 2022-12-19T17:21:17.306+0100	INFO	Reconciling	{"controller": "nova", "controllerGroup": "nova.openstack.org", "controllerKind": "Nova", "Nova": {"name":"nova","namespace":"openstack"}, "namespace": "openstack", "name": "nova", "reconcileID": "154c6142-0dfe-4587-a655-b91f566bd131", "ObjectType": "Nova", "ObjectNamespace": "openstack", "ObjectName": "nova"}
> ```
> 
> The fields that is added automatically:
> 
> * "controller": "nova"
> * "controllerGroup": "nova.openstack.org"
> * "controllerKind": "Nova"
> * "Nova": {"name":"nova","namespace":"openstack"}
> * "namespace": "openstack"
> * "name": "nova"
> * "reconcileID": "154c6142-0dfe-4587-a655-b91f566bd131"
> 
> The fields we add via lib-common:
> 
> * "ObjectType": "Nova"
> * "ObjectNamespace": "openstack"
> * "ObjectName": "nova"
> 
> So clearly what we try to do here is redundant and already there if we use the logger from the context we get as the parameter fo the Reconcile call. So a simple `log.FromContext(ctx)` is enough in each operators to have structural logging with the information we need. Also ctx can be passed to lib-common so lib-common can do the same if logging is needed.

With this, best to switch to that, update operators and lib-common to not use the current funcs, deprecate and remove the logging from lib-common. As there is only Info/Error, we might want to think about give a definition for the verbosity levels for Info [1] ?
~~~
Verbosity-levels on info logs. This gives developers a chance to indicate arbitrary grades of importance for info logs, 
without assigning names with semantic meaning such as "warning", "trace", and "debug."
Superficially this may feel very similar, but the primary difference is the lack of semantics. Because verbosity is a 
numerical value, it's safe to assume that an app running with higher verbosity means more (and less important) logs 
will be generated.
~~~

If people feel better to have names instead of numerical value, we could also define const in lib-common for those, like 0 == Warn, 1 == Trace, 2 == Debug. Or at least set a level where Debug starts.

We also need to expose the Level to be able to be set by the user.

[1] https://github.com/go-logr/logr#background

bogdando (approved):
> There is an example PR that partially switches over nova to the new interface https://github.com/openstack-k8s-operators/nova-operator/pull/206 

works like a charm, well done

stuggi:
> > It turns out that the this whole thing is not necessary if we use the logger from the ctx directly as it already has the extra fields added to it about the instance that is being reconciled. Here is an example log from this patch applied to nova.
> > ```
> > 2022-12-19T17:21:17.306+0100	INFO	Reconciling	{"controller": "nova", "controllerGroup": "nova.openstack.org", "controllerKind": "Nova", "Nova": {"name":"nova","namespace":"openstack"}, "namespace": "openstack", "name": "nova", "reconcileID": "154c6142-0dfe-4587-a655-b91f566bd131", "ObjectType": "Nova", "ObjectNamespace": "openstack", "ObjectName": "nova"}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The fields that is added automatically:
> > 
> > * "controller": "nova"
> > * "controllerGroup": "nova.openstack.org"
> > * "controllerKind": "Nova"
> > * "Nova": {"name":"nova","namespace":"openstack"}
> > * "namespace": "openstack"
> > * "name": "nova"
> > * "reconcileID": "154c6142-0dfe-4587-a655-b91f566bd131"
> > 
> > The fields we add via lib-common:
> > 
> > * "ObjectType": "Nova"
> > * "ObjectNamespace": "openstack"
> > * "ObjectName": "nova"
> > 
> > So clearly what we try to do here is redundant and already there if we use the logger from the context we get as the parameter fo the Reconcile call. So a simple `log.FromContext(ctx)` is enough in each operators to have structural logging with the information we need. Also ctx can be passed to lib-common so lib-common can do the same if logging is needed.
> 
> With this, best to switch to that, update operators and lib-common to not use the current funcs, deprecate and remove the logging from lib-common. As there is only Info/Error, we might want to think about give a definition for the verbosity levels for Info [1] ?
> 
> ```
> Verbosity-levels on info logs. This gives developers a chance to indicate arbitrary grades of importance for info logs, 
> without assigning names with semantic meaning such as "warning", "trace", and "debug."
> Superficially this may feel very similar, but the primary difference is the lack of semantics. Because verbosity is a 
> numerical value, it's safe to assume that an app running with higher verbosity means more (and less important) logs 
> will be generated.
> ```
> 
> If people feel better to have names instead of numerical value, we could also define const in lib-common for those, like 0 == Warn, 1 == Trace, 2 == Debug. Or at least set a level where Debug starts.
> 
> We also need to expose the Level to be able to be set by the user.
> 
> [1] https://github.com/go-logr/logr#background

also if we not use a typed wrapper, we should wrap errors outside of the reconcile loop/lib-common where it makes sense using fmt.Errorf() with %w . Wrapping an error with %w makes it available to errors.Is and errors.As:

~~~
err := fmt.Errorf("access denied: %w", ErrPermission)
...
if errors.Is(err, ErrPermission) ...
~~~

gibizer:
- On the verbosity levels. I don't feel the need right now to have more than Info and Error which we already have. 

- On the error wrapping I'm not sure I follow. I suspect this is related to `util.WrapErrorForObject` If so then I guess we have two independent goals based on that function: 
   - i) make sure that when we create new errors from existing ones we wrap them with %w to allow the error type information to be propagated. This feels like something that should checked from a linter. 
   - ii) make sure that the new error has a message that contains the type, and namespaced name of an object. I think this is solved in most cases at the logging level where the object is included. If other than the object under reconciliation is needed in the error then that would remain to be handled by each caller.

gibizer:
>     * On the error wrapping I'm not sure I follow. I suspect this is related to `util.WrapErrorForObject` If so then I guess we have two independent goals based on that function:
>       
>       * i) make sure that when we create new errors from existing ones we wrap them with %w to allow the error type information to be propagated. This feels like something that should checked from a linter.

OK there is a linter for it already in golangci-lint: https://github.com/polyfloyd/go-errorlint

It is disabled by default in golangci-lint: https://golangci-lint.run/usage/linters/ . I enabled it locally for nova and tested it. It catches the non wrapped errors case

```
$ cat .golangci.yaml 
linters:
  # Enable specific linter
  # https://golangci-lint.run/usage/linters/#enabled-by-default
  enable:
    errorlint

gibi@bedrock:~/upstream/git/openstack-k8s-operators/nova-operator D[master]                                                                                                                         [001] [11:05:44]
$ git diff
diff --git a/controllers/nova_controller.go b/controllers/nova_controller.go
index 934f1e2..20281a0 100644
--- a/controllers/nova_controller.go
+++ b/controllers/nova_controller.go
@@ -98,6 +98,9 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
        }
        util.LogForObject(h, "Reconciling", instance)
 
+       err2 := fmt.Errorf("level1")
+       _ = fmt.Errorf("level2 %v", err2)
+
        // initialize status fields
        if err = r.initStatus(ctx, h, instance); err != nil {
                return ctrl.Result{}, err

$ pre-commit run -a
make-manifests...........................................................Passed
make-generate............................................................Passed
make-operator-lint.......................................................Passed
go fmt...................................................................Passed
go vet...................................................................Passed
go-mod-tidy..............................................................Passed
go lint..................................................................Passed
golangci-lint............................................................Failed
- hook id: golangci-lint
- exit code: 1

controllers/nova_controller.go:102:30: non-wrapping format verb for fmt.Errorf. Use `%w` to format errors (errorlint)
	_ = fmt.Errorf("level2 %v", err2)
	                            ^

check for added large files..............................................Passed
fix utf-8 byte order marker..............................................Passed
check for case conflicts.................................................Passed
check that executables have shebangs.....................................Passed
check that scripts with shebangs are executable..........................Passed
check for merge conflicts................................................Passed
check for broken symlinks............................(no files to check)Skipped
detect destroyed symlinks................................................Passed
check yaml...............................................................Passed
check json...............................................................Passed
detect private key.......................................................Passed
fix end of files.........................................................Passed
don't commit to branch...................................................Failed
- hook id: no-commit-to-branch
- exit code: 1
trim trailing whitespace.................................................Passed
```

gibizer:
I suggest to close this PR and start changing the exiting operators to use the logger from the context.

abays (approved):
/lgtm

abays:
> I suggest to close this PR and start changing the exiting operators to use the logger from the context.

Oh, just saw this.  Sure, we can do that.

stuggi:
> * On the verbosity levels. I don't feel the need right now to have more than Info and Error which we already have.

correct, I agree that right now its not required. We just need to monitor and if people start to use it that we have a common definition of the levels.

> * On the error wrapping I'm not sure I follow. I suspect this is related to `util.WrapErrorForObject` If so then I guess we have two independent goals based on that function:
>   
>   * i) make sure that when we create new errors from existing ones we wrap them with %w to allow the error type information to be propagated. This feels like something that should checked from a linter.

right

>   * ii) make sure that the new error has a message that contains the type, and namespaced name of an object. I think this is solved in most cases at the logging level where the object is included. If other than the object under reconciliation is needed in the error then that would remain to be handled by each caller.

With wrapping I mean that we don't just return the error and it is hard to track where exact the error happened, so its best to wrap it and add a message to be able to better identify it. And use %w so that you can still have the possibility to use errors.Is or so.

> > ```
> > * On the error wrapping I'm not sure I follow. I suspect this is related to `util.WrapErrorForObject` If so then I guess we have two independent goals based on that function:
> >   
> >   * i) make sure that when we create new errors from existing ones we wrap them with %w to allow the error type information to be propagated. This feels like something that should checked from a linter.
> > ```
> 
> OK there is a linter for it already in golangci-lint: https://github.com/polyfloyd/go-errorlint
> 
> It is disabled by default in golangci-lint: https://golangci-lint.run/usage/linters/ . I enabled it locally for nova and tested it. It catches the non wrapped errors case
> 
> ```
> $ cat .golangci.yaml 
> linters:
>   # Enable specific linter
>   # https://golangci-lint.run/usage/linters/#enabled-by-default
>   enable:
>     errorlint
> 
> gibi@bedrock:~/upstream/git/openstack-k8s-operators/nova-operator D[master]                                                                                                                         [001] [11:05:44]
> $ git diff
> diff --git a/controllers/nova_controller.go b/controllers/nova_controller.go
> index 934f1e2..20281a0 100644
> --- a/controllers/nova_controller.go
> +++ b/controllers/nova_controller.go
> @@ -98,6 +98,9 @@ func (r *NovaReconciler) Reconcile(ctx context.Context, req ctrl.Request) (resul
>         }
>         util.LogForObject(h, "Reconciling", instance)
>  
> +       err2 := fmt.Errorf("level1")
> +       _ = fmt.Errorf("level2 %v", err2)
> +
>         // initialize status fields
>         if err = r.initStatus(ctx, h, instance); err != nil {
>                 return ctrl.Result{}, err
> 
> $ pre-commit run -a
> make-manifests...........................................................Passed
> make-generate............................................................Passed
> make-operator-lint.......................................................Passed
> go fmt...................................................................Passed
> go vet...................................................................Passed
> go-mod-tidy..............................................................Passed
> go lint..................................................................Passed
> golangci-lint............................................................Failed
> - hook id: golangci-lint
> - exit code: 1
> 
> controllers/nova_controller.go:102:30: non-wrapping format verb for fmt.Errorf. Use `%w` to format errors (errorlint)
> 	_ = fmt.Errorf("level2 %v", err2)
> 	                            ^
> 
> check for added large files..............................................Passed
> fix utf-8 byte order marker..............................................Passed
> check for case conflicts.................................................Passed
> check that executables have shebangs.....................................Passed
> check that scripts with shebangs are executable..........................Passed
> check for merge conflicts................................................Passed
> check for broken symlinks............................(no files to check)Skipped
> detect destroyed symlinks................................................Passed
> check yaml...............................................................Passed
> check json...............................................................Passed
> detect private key.......................................................Passed
> fix end of files.........................................................Passed
> don't commit to branch...................................................Failed
> - hook id: no-commit-to-branch
> - exit code: 1
> trim trailing whitespace.................................................Passed
> ```

great, if this linter works well in nova, we should enable it in all of the other operators.

PR #130, opened 2022-12-22
Add basic framework for CronJob resource type

kajinamit:
This introduces the basic implementation to manage CronJob resource type, so that individual operators can use this framework to set up house-keeping cron jobs such as db purge.

### DIFF
--- a/modules/common/cronjob/cronjob.go
+++ b/modules/common/cronjob/cronjob.go
@@ -1,0 +1,129 @@
+/*
+Copyright 2022 Red Hat
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cronjob
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	batchv1 "k8s.io/api/batch/v1"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// NewCronJob returns an initialized CronJob.
+func NewCronJob(
+	cronjob *batchv1.CronJob,
+	timeout time.Duration,
+) *CronJob {
+	return &CronJob{
+		cronjob: cronjob,
+		timeout: timeout,
+	}
+}
+
+// CreateOrPatch - creates or patches a cronjob, reconciles after Xs if object won't exist.
+func (cj *CronJob) CreateOrPatch(
+	ctx context.Context,
+	h *helper.Helper,
+) (ctrl.Result, error) {
+	op, err := controllerutil.CreateOrPatch(ctx, h.GetClient(), cj.cronjob, func() error {
+		err := controllerutil.SetControllerReference(h.GetBeforeObject(), cj.cronjob, h.GetScheme())
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
+	if err != nil {
+		if k8s_errors.IsNotFound(err) {
+			h.GetLogger().Info(fmt.Sprintf("CronJob %s not found, reconcile in %s", cj.cronjob.Name, cj.timeout))
+			return ctrl.Result{RequeueAfter: cj.timeout}, nil
+		}
+		return ctrl.Result{}, err
+	}
+	if op != controllerutil.OperationResultNone {
+		h.GetLogger().Info(fmt.Sprintf("CronJob %s - %s", cj.cronjob.Name, op))
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// Delete - delete a cronjob.
+func (cj *CronJob) Delete(
+	ctx context.Context,
+	h *helper.Helper,
+) error {
+	err := h.GetClient().Delete(ctx, cj.cronjob)
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		err = fmt.Errorf("Error deleting cronjob %s: %v", cj.cronjob.Name, err)
+		return err
+	}
+
+	return nil
+}
+
+// GetCronJob - get the cronjob object.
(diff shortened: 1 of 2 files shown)


modules/common/cronjob/cronjob.go
```diff
@@ -0,0 +1,114 @@
...
+	cj.timeout = timeout
+}
+
+// GetCronJobWithName func
+func GetCronJobWithName(
+	ctx context.Context,
+	h *helper.Helper,
+	name string,
+	namespace string,
+) (*batchv1.CronJob, error) {
```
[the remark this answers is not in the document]
kajinamit: I'm fine with updating the function accordingly. My main aim is to provide the consistent interface with the other modules.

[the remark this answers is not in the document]
kajinamit: To make sure my understanding is correct, I assume the expectation is to add GetCronJobByName function, which internally use GetCronJobWithName, right ? If that is correct then I'd prefer leaving that now, since we are not really using Get*WithName functions in other operators atm and adding a new interface does not affect subsequent implementation in operators side.

modules/common/cronjob/cronjob.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		if k8s_errors.IsNotFound(err) {
+			h.GetLogger().Info(fmt.Sprintf("CronJob %s not found, reconcile in %s", cj.cronjob.Name, cj.timeout))
+			return ctrl.Result{RequeueAfter: cj.timeout}, nil
+		}
+		return ctrl.Result{}, err
+	}
+	if op != controllerutil.OperationResultNone {
+		h.GetLogger().Info(fmt.Sprintf("CronJob %s - %s", cj.cronjob.Name, op))
+		return ctrl.Result{RequeueAfter: cj.timeout}, nil
```
gibizer, line 66: If the caller 'Owns' CronJob then this requeue is not necessary. The state change of the  CronJob object will trigger a reconciliation automatically.

modules/common/cronjob/cronjob.go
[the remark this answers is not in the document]
gibizer: I agree with @stuggi here.

gibizer (changes requested):

modules/common/cronjob/cronjob.go
kajinamit: Fixed thanks for catching this.

modules/common/cronjob/cronjob.go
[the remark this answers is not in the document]
stuggi: @kajinamit your understanding is correct. Have GetCronJobByName() like https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/database/funcs.go#L312 and use a getCronJobWithName like https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/database/funcs.go#L268.  I think we should already do this for CronJob and then enhance the others first by adding GetCronJobByName and when all consumers got updated then to the adjustment to lib-common.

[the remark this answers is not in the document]
kajinamit: I updated the PR to include GetCronJobByName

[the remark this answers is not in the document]
gibizer: One note about https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/database/funcs.go#L312 that the name there is the dbName instead of the name of the k8s resource, i.e. metadata.Name. It think that is a sub-optimal simplification there. So when we follow this pattern please try to idenify CronJob by the resource name.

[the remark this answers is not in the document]
kajinamit: IIUC the current proposed implementation should look up the resource by resource name and namespace name but please correct me if I'm wrong .

modules/common/cronjob/cronjob.go
```diff
@@ -0,0 +1,129 @@
...
+
+	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// NewCronJob returns an initialized CronJob.
+func NewCronJob(
+	cronjob *batchv1.CronJob,
+	timeout int,
```
gibizer, line 36: As a generic rule I suggest to always use time.Duration for timeout parameters. This allows the caller to specify a lot more fine grained timeout value. 
One reason is that in envtest, running locally, I can use sub second timeout values to speed up the test execution.
Another good thing about time.Duration that it encodes the unit of the timeout value in the type itself.

modules/common/cronjob/cronjob.go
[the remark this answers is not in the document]
gibizer: yes, this code does that. thanks.

gibizer (changes requested):

modules/common/cronjob/cronjob.go
kajinamit: Sounds reasonable. I pushed one more commit to convert usage of int in all modules.

gibizer: @kajinamit I appreciate your eagerness to fix it all over the place but please note that this is a breaking change in other operators that already using the interface base on the int type. So I suggest to fix it for CronJob here as it is not used yet. And do the overall cleanup separately by not just fixing the lib-common interface but also adjusting the operator codes that are using it.

kajinamit: I agree with the concern but I'm more concerned with creating inconsistent interfaces and ended up leaving it eventually. That would cause confusion for long term maintenance.

I can update the other modules to adapt to this interface change, but if now is not a good timing to make this change in all operators then I prefer using it consistently across all and then replace all by time.Duration at the same time.

modules/common/cronjob/cronjob.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+// NewCronJob returns an initialized CronJob.
+func NewCronJob(
+	cronjob *batchv1.CronJob,
+	timeout time.Duration,
+) *CronJob {
+	return &CronJob{
+		cronjob: cronjob,
+		timeout: timeout * time.Second,
```
gibizer, line 40: if timeout is a Duration, then we should not multiply it with time.Second here.

modules/common/cronjob/cronjob.go
kajinamit: I agree with the plan. Updated the change to touch only cronjob for now. Will create a separate PR(s) to update the other modules.

gibizer (changes requested):

modules/common/cronjob/cronjob.go
gibizer: Cool, thanks. Add me to the new PR and will try to help push it through

gibizer (commented):
Looks good to me.

modules/common/cronjob/cronjob.go
kajinamit: Yes we have to use the raw Duration.

stuggi (approved):
/lgtm

PR #131, opened 2023-01-04
Add statefulset/conditions ops for test helper

bogdando:
With that, other operators can refer those for providing
envtest coverage for statefulsets and conditions involving
test cases.


modules/test/helpers/conditions.go
```diff
@@ -0,0 +1,87 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+type conditionsGetter interface {
+	GetConditions(name types.NamespacedName) condition.Conditions
+}
+
+type conditionGetterFunc func(name types.NamespacedName) condition.Conditions
+
+func (f conditionGetterFunc) GetConditions(name types.NamespacedName) condition.Conditions {
```
bogdando, line 30: There are still *unused* problems with those. Quite a few operators copy and paste these. Let's figure out how to properly plug that into the test helper

bogdando:
this is currently needed in order to align nova and placement tests, but other operators can benefit too

modules/test/helpers/conditions.go
bogdando: I'm a bit lost with adapting those as proper interfaces after moved from nova tests

modules/test/helpers/common.go
```diff
@@ -19,17 +19,32 @@ import (
...
 )
 
+type conditionsGetter interface {
+	GetConditions(name types.NamespacedName) condition.Conditions
+}
+
+type ConditionGetterFunc func(name types.NamespacedName) condition.Conditions
+
+// ConditionGetterFunc -
+func (f ConditionGetterFunc) GetConditions(name types.NamespacedName) condition.Conditions {
```
bogdando, line 37: with this, nova tests refactored can pass.
I cen see other operators use GetConditions() quite extensively, I believe this should help with testing expected conditions there as well

modules/test/helpers/conditions.go
```diff
@@ -0,0 +1,77 @@
...
+		g.Expect(actualCondition.Reason).To(
+			t.Equal(expectedReason),
+			"%s condition has a different reason. Actual condition: %v", conditionType, actualCondition)
+		g.Expect(actualCondition.Message).To(
+			t.Equal(expecteMessage),
+			"%s condition has a different message. Actual condition: %v", conditionType, actualCondition)
+	}, tc.timeout, tc.interval).Should(t.Succeed())
+
+	tc.logger.Info("ExpectConditionWithDetails succeeded", "type", conditionType, "expected status", expectedStatus, "on", name)
+}
```
gibizer, line 77: No newline at the end of the file. Interesting that it is not caught by go fmt or other lint

gibizer (changes requested):
just two small nit but otherwise looks good

gibizer (approved):
/lgtm

PR #133, opened 2023-01-16
Add annotations package with GetNADAnnotation

stuggi:
To attach additional interfaces via multus to resources the `k8s.v1.cni.cncf.io/networks` annotation needs to be added to the pod spec. e.g.

k8s.v1.cni.cncf.io/networks: '[{"name": "internalapi", "namespace": "openstack"},{"name": "storage", "namespace": "openstack"}]'

GetNADAnnotation() returns this annotation based on the namespace and the list of network-attachment-definitions passed in.

### DIFF
--- a/modules/common/annotations/network.go
+++ b/modules/common/annotations/network.go
@@ -1,0 +1,55 @@
+/*
+Copyright 2023 Red Hat
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


stuggi:
https://github.com/openstack-k8s-operators/keystone-operator/pull/149 shows on how that can be used by an operator

abays (approved):
/lgtm

PR #134, opened 2023-01-16
Add db/trans/keystone objects ops for test helper

bogdando:
With that, other operators can refer those for providing envtest coverage for mariadb database schemas, transportURL objeccts, and keystone services, endpoints, APIs in functional test cases.

Signed-off-by: Bohdan Dobrelia <bdobreli@redhat.com>


bogdando:
PTAL

modules/test/helpers/database.go
```diff
@@ -0,0 +1,97 @@
...
+func (tc *TestHelper) CreateDBService(namespace string, mariadbCRName string, spec corev1.ServiceSpec) types.NamespacedName {
+	// The Name is used as the hostname to access the service. So
+	// we generate something unique for the MariaDB CR it represents
+	// so we can assert that the correct Service is selected.
+	serviceName := "hostname-for-" + mariadbCRName
+	service := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      serviceName,
+			Namespace: namespace,
+			// NOTE(gibi): The lib-common databvase module looks up the
```
gibizer, line 38: nit: database

modules/test/helpers/keystone.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// CreateKeystoneAPI -
+func (tc *TestHelper) CreateKeystoneAPI(namespace string) types.NamespacedName {
```
gibizer, line 29: @stuggi I vaguely remember that we talked about moving this to Keystone itself. But now that I played with the idea I think that is a complex change. That would need a new module in keystone-operator like `keystone-operator/test` so that nova-operator can import that without importing the whole keystone implementation. We could put it to keystone-operator/api but that feels like mixing things into the api module that is not really an api. So I think it is good that we move this here to lib-common. What do you think?

gibizer:
I'm OK with this but waiting for @stuggi's response

modules/test/helpers/keystone.go
bogdando: @stuggi PTAL

stuggi: sorry, I totally missed this in the mass of emails. I try to do better :) 
Right, we talked about this, but agree that its better to have this in the the helpers here in lib-common.

stuggi (approved):
/lgtm

PR #144, opened 2023-01-19
Add CodeQL workflow to find security vulnerabilities

fao89:
https://codeql.github.com/

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,39 @@
+name: "CodeQL"
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '37 1 * * 6'
+
+jobs:
+  analyze:
+    name: Analyze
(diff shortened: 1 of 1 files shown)


stuggi:
Is this something you want to propose we should enable on all operators?

stuggi:
> Is this something you want to propose we should enable on all operators?

I mean I am fine with it, I just think we should do it everywhere not only on individual repos

fao89:
> Is this something you want to propose we should enable on all operators?

I think we can use this repo as a test,
yesterday I saw the insights about this module
https://deps.dev/go/github.com%2Fopenstack-k8s-operators%2Flib-common%2Fmodules%2Fopenstack/v0.0.0-20230119120622-e0f68a14523a
and I became curious if CodeQL could catch these securities suggestions earlier for us

stuggi (approved):

PR #155, opened 2023-01-31
Use time.Duration for timeout

kajinamit:
instead of int. This allows us to use more fine grained timeout value such as 0.1, and encodes the unit in the type itself.

### DIFF
--- a/modules/common/deployment/deployment.go
+++ b/modules/common/deployment/deployment.go
@@ -34,11 +34,11 @@ import (
 // NewDeployment returns an initialized Deployment.
 func NewDeployment(
 	deployment *appsv1.Deployment,
-	timeout int,
+	timeout time.Duration,
 ) *Deployment {
 	return &Deployment{
 		deployment: deployment,
-		timeout:    time.Duration(timeout) * time.Second,
+		timeout:    timeout,
 	}
 }
 
@@ -109,12 +109,6 @@ func (d *Deployment) Delete(
 // GetDeployment - get the deployment object.
 func (d *Deployment) GetDeployment() appsv1.Deployment {
 	return *d.deployment
-}
-
-// SetTimeout defines the duration used for requeueing while waiting for the deployment
-// to finish.
-func (d *Deployment) SetTimeout(timeout time.Duration) {
-	d.timeout = timeout
 }
 
 // GetDeploymentWithName func
--- a/modules/common/job/job.go
+++ b/modules/common/job/job.go
@@ -39,7 +39,7 @@ func NewJob(
 	job *batchv1.Job,
 	jobType string,
 	preserve bool,
-	timeout int,
+	timeout time.Duration,
 	beforeHash string,
 ) *Job {
 
@@ -47,7 +47,7 @@ func NewJob(
 		job:        job,
 		jobType:    jobType,
 		preserve:   preserve,
-		timeout:    time.Duration(timeout) * time.Second, // timeout to set in s to reconcile
+		timeout:    timeout,
 		beforeHash: beforeHash,
 		changed:    false,
 	}
@@ -197,12 +197,6 @@ func (j *Job) HasChanged() bool {
 // GetHash func
 func (j *Job) GetHash() string {
 	return j.hash
-}
-
-// SetTimeout defines the duration used for requeueing while waiting for the job
-// to finish.
-func (j *Job) SetTimeout(timeout time.Duration) {
-	j.timeout = timeout
 }
 
 // DeleteJob func
--- a/modules/common/pvc/pvc.go
+++ b/modules/common/pvc/pvc.go
@@ -34,7 +34,7 @@ import (
 // NewPvc returns an initialized Pvc.
 func NewPvc(
 	pvc *corev1.PersistentVolumeClaim,
-	timeout int,
+	timeout time.Duration,
 ) *Pvc {
 	return &Pvc{
 		pvc:     pvc,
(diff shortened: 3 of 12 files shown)


kajinamit:
I noticed that a few requeuTimeout parameters take int. Should we update these as well ?

https://github.com/openstack-k8s-operators/lib-common/blob/675e99666dbc83fb48d366549d11fd13d849289e/modules/common/configmap/configmap.go#L247
https://github.com/openstack-k8s-operators/lib-common/blob/675e99666dbc83fb48d366549d11fd13d849289e/modules/common/secret/secret.go#L357

modules/common/deployment/deployment.go
```diff
@@ -34,11 +34,11 @@ import (
 // NewDeployment returns an initialized Deployment.
 func NewDeployment(
 	deployment *appsv1.Deployment,
-	timeout int,
+	timeout time.Duration,
 ) *Deployment {
 	return &Deployment{
 		deployment: deployment,
-		timeout:    time.Duration(timeout) * time.Second,
+		timeout:    timeout,
```
gibizer, line 41: you can drop `SetTimeout` below. It was a stop gap solution to re-define smaller than 1 second timeout during test without breaking the `NewDeployment` interface. But now that we are changing that interface the timeout can be defined directly in `NewDeployment()` with less than 1 sec granularity.  I think it is only used by nova-operator.

modules/common/job/job.go
```diff
@@ -39,15 +39,15 @@ func NewJob(
...
+	timeout time.Duration,
 	beforeHash string,
 ) *Job {
 
 	return &Job{
 		job:        job,
 		jobType:    jobType,
 		preserve:   preserve,
-		timeout:    time.Duration(timeout) * time.Second, // timeout to set in s to reconcile
+		timeout:    timeout,
```
gibizer, line 50: you can drop `SetTimeout`

modules/common/pvc/pvc.go
```diff
@@ -80,7 +80,7 @@ func (p *Pvc) CreateOrPatch(
 	if err != nil {
 		if k8s_errors.IsNotFound(err) {
 			h.GetLogger().Info(fmt.Sprintf("Pvc %s not found, reconcile in %ds", pvc.Name, p.timeout))
-			return ctrl.Result{RequeueAfter: time.Duration(p.timeout) * time.Second}, nil
+			return ctrl.Result{RequeueAfter: p.timeout}, nil
```
gibizer, line 83: The above log line has `%ds` that will be wrong now. You can simply say `%s` and Duration will produce a human readable string.

modules/common/route/route.go
```diff
@@ -99,7 +99,7 @@ func (r *Route) CreateOrPatch(
 	if err != nil {
 		if k8s_errors.IsNotFound(err) {
 			h.GetLogger().Info(fmt.Sprintf("Route %s not found, reconcile in %ds", route.Name, r.timeout))
-			return ctrl.Result{RequeueAfter: time.Duration(r.timeout) * time.Second}, nil
+			return ctrl.Result{RequeueAfter: r.timeout}, nil
```
gibizer, line 102: use `%s` tor print the timeout value above

modules/common/service/service.go
```diff
@@ -96,7 +96,7 @@ func (s *Service) CreateOrPatch(
 	if err != nil {
 		if k8s_errors.IsNotFound(err) {
 			h.GetLogger().Info(fmt.Sprintf("Service %s not found, reconcile in %ds", service.Name, s.timeout))
-			return ctrl.Result{RequeueAfter: time.Duration(s.timeout) * time.Second}, nil
+			return ctrl.Result{RequeueAfter: s.timeout}, nil
```
gibizer, line 99: use `%s` in the log line above

modules/common/statefulset/statefulset.go
```diff
@@ -34,11 +34,11 @@ import (
 // NewStatefulSet returns an initialized NewStatefulset.
 func NewStatefulSet(
 	statefulset *appsv1.StatefulSet,
-	timeout int,
+	timeout time.Duration,
 ) *StatefulSet {
 	return &StatefulSet{
 		statefulset: statefulset,
-		timeout:     time.Duration(timeout) * time.Second,
+		timeout:     timeout,
```
gibizer, line 41: you can drop `SetTimeout`

gibizer:
> I noticed that a few requeueTimeout parameters take int. Should we update these as well ?
> 
> https://github.com/openstack-k8s-operators/lib-common/blob/675e99666dbc83fb48d366549d11fd13d849289e/modules/common/configmap/configmap.go#L247
> 
> https://github.com/openstack-k8s-operators/lib-common/blob/675e99666dbc83fb48d366549d11fd13d849289e/modules/common/secret/secret.go#L357

Yes please. :)

gibizer (changes requested):
Thanks for starting this cleanup!
I have couple of requests inline.
Also right before we merge this we should notify the devs of the operators that this is a breaking change in lib-common so they should be careful bumping to this version blindly.

modules/common/deployment/deployment.go
kajinamit: Thanks for sharing that backgroupd. Now this replacement would make more sense :-)

modules/common/job/job.go
kajinamit: done

modules/common/pvc/pvc.go
kajinamit: done

modules/common/service/service.go
kajinamit: done

modules/common/route/route.go
kajinamit: done

modules/common/statefulset/statefulset.go
kajinamit: done

kajinamit:
@gibi thanks for your comments. Updated the PR to address these. The latest version covers requeueTimeout I mentioned. I'd probably bring this to next sync to make people aware of this upcoming change.

gibizer:
> @gibi thanks for your comments. Updated the PR to address these. The latest version covers requeueTimeout I mentioned. I'd probably bring this to next sync to make people aware of this upcoming change.

Looks good to me. Let's bring this up on the next sync and then merge it.

abays (approved):
/lgtm

PR #156, opened 2023-01-31
Add remove function to conditions

rabi:
We would need to remove a condition at times from the conditions slice. This adds the new helper function for it.

### DIFF
--- a/modules/common/condition/funcs.go
+++ b/modules/common/condition/funcs.go
@@ -86,6 +86,21 @@ func (conditions *Conditions) Set(c *Condition) {
 	conditions.Sort()
 }
 
+// Remove a condition from the slice of conditions
+func (conditions *Conditions) Remove(t Type) {
+	if conditions == nil || len(*conditions) == 0 {
+		return
+	}
+	newConditions := make([]Condition, 0, len(*conditions)-1)
+	for _, condition := range *conditions {
+		if condition.Type != t {
+			newConditions = append(newConditions, condition)
+		}
+	}
+
+	*conditions = newConditions
+}
+
 // Get returns the condition with the given type, if the condition does not exists,
 // it returns nil.
 func (conditions *Conditions) Get(t Type) *Condition {
--- a/modules/common/condition/funcs_test.go
+++ b/modules/common/condition/funcs_test.go
@@ -165,6 +165,69 @@ func TestSet(t *testing.T) {
 	g.Expect(c2.LastTransitionTime).To(BeIdenticalTo(time1))
 }
 
+func TestRemove(t *testing.T) {
+	tests := []struct {
+		name       string
+		conditions Conditions
+		cType      Type
+		expected   Conditions
+	}{
+		{
+			name: "present",
+			conditions: CreateList(
+				unknownReady,
+				unknownA,
+				unknownB,
+			),
+			cType: unknownA.Type,
+			expected: CreateList(
+				unknownReady,
+				unknownB,
+			),
+		},
+		{
+			name: "unknownReady-remove",
+			conditions: CreateList(
+				unknownReady,
+				unknownA,
+				unknownB,
+			),
+			cType: unknownReady.Type,
+			expected: CreateList(
+				unknownA,
+				unknownB,
+			),
+		},
+		{
+			name: "absent",
+			conditions: CreateList(
+				unknownReady,
+				unknownA,
+			),
+			cType: unknownB.Type,
+			expected: CreateList(
+				unknownReady,
+				unknownA,
+			),
+		},
+		{
+			name:       "empty",
+			conditions: Conditions{},
+			cType:      unknownA.Type,
+			expected:   Conditions{},
+		},
+	}
+
+	g := NewWithT(t)
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tt.conditions.Remove(tt.cType)
+			g.Expect(tt.expected).To(haveSameConditionsOf(tt.conditions))
+		})
+	}
+}
+
 func TestHasSameState(t *testing.T) {
 	g := NewWithT(t)
 


modules/common/condition/funcs.go
```diff
@@ -86,6 +86,21 @@ func (conditions *Conditions) Set(c *Condition) {
 	conditions.Sort()
 }
 
+// Remove a condition from the slice of conditions
+func (conditions *Conditions) Remove(conditionType Type) {
+	if conditions == nil || len(*conditions) == 0 {
```
stuggi, line 91: might also want to use c.Has() to check if the type is in the conditions list and return if not, but then we'd run through them twice, once for checking and bellow to remove it.

stuggi: an option would be to change c.Has() to also return the index of the found condition and then delete the index in Remove(), but this would be a breaking change and would require updates to operators.

modules/common/condition/funcs.go
```diff
@@ -86,6 +86,21 @@ func (conditions *Conditions) Set(c *Condition) {
...
+		return
+	}
+	newConditions := make([]Condition, 0, len(*conditions)-1)
```
stuggi, line 94: if the condition type is not part of the conditions list, the `newConditions` will not have the correct size, its not really an issue as `append()` will handle that.

modules/common/condition/funcs.go
```diff
@@ -86,6 +86,21 @@ func (conditions *Conditions) Set(c *Condition) {
 	conditions.Sort()
 }
 
+// Remove a condition from the slice of conditions
+func (conditions *Conditions) Remove(conditionType Type) {
```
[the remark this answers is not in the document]
rabi: OK, will change. I don't prefer single character arguments. I thought single character names for method receivers and loop variables is ok in go, but not arguments or variables.

modules/common/condition/funcs.go
rabi: Yeah, it's just initial size for the slice and in normal scenario would be fine.

modules/common/condition/funcs_test.go
```diff
@@ -165,6 +165,55 @@ func TestSet(t *testing.T) {
...
+			condition: nil,
+			want:      CreateList(unknownReady),
+		},
+		{
+			name:      "Add unknownB condition",
+			condition: unknownB,
+			want:      CreateList(unknownReady, unknownB),
+		},
+		{
+			name:      "Add another condition unknownA, gets sorted",
```
gibizer, line 187: these names feels like copy paste from above. I suggest to align them to the current test case.

modules/common/condition/funcs_test.go
```diff
@@ -165,6 +165,55 @@ func TestSet(t *testing.T) {
...
+			name:      "Add same condition unknownA, won't duplicate",
+			condition: unknownA,
+			want:      CreateList(unknownReady, unknownA, unknownB),
+		},
+	}
+
+	g := NewWithT(t)
+
+	conditions.Init(nil)
+	g.Expect(conditions).To(haveSameConditionsOf(CreateList(unknownReady)))
```
gibizer, line 201: it is already tested above that an empty conditions init will result in a single Ready condition

modules/common/condition/funcs_test.go
```diff
@@ -165,6 +165,55 @@ func TestSet(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			conditions.Set(tt.condition)
+
+			g.Expect(conditions).To(haveSameConditionsOf(tt.want))
+		})
+	}
```
gibizer, line 210: ditto, this is already tested above that Set() works as expected

modules/common/condition/funcs_test.go
```diff
@@ -165,6 +165,55 @@ func TestSet(t *testing.T) {
...
+
+	conditions.Remove(unknownA.Type)
+	g.Expect(conditions.Has(unknownA.Type)).To(BeFalse())
+	g.Expect(conditions.Has(unknownB.Type)).To(BeTrue())
```
gibizer, line 214: OK, this is the meat of the test. But I suggest to have another case where you try to remove a type that is not in the lest.

Also try to remove the unknowReady as that is somewhat special.

gibizer (changes requested):
I have couple of test suggestions

modules/common/condition/funcs_test.go
rabi: Align? You want the initial conditions slice to be created differently?

modules/common/condition/funcs_test.go
rabi: This is just creating an initial set to remove conditions... so we can't use other functions in this test :)

modules/common/condition/funcs_test.go
gibizer: The L171 - L196 code is testing what happens when you add different conditions to the list one by one. Asserting things like sorting behavior of the list. (Hence the name of here says "gets sorted").  But this logic is already tested by other test cases above. So please don't repeat it.

What you need is a condition list to remove conditions from. That can be done as simply as
```golang
	conditions := CreateList(unknownReady, unknownA, unknownB)
```
Then you can do L198-... to test the removals as you proposed.

gibizer (changes requested):

modules/common/condition/funcs_test.go
rabi: I'm sure this can be optimized. Not sure what's the harm in having a few lines of extra test code to create the list like in other tests. I can change if you insist though.

gibizer: This is not optimization this is removal of duplicated test code. Less code means less logic to understand and maintain.

rabi: By optimization I did not mean only code optimization for performance, but also 'test' code readability. Anyway, now that I understand the ginkgo/gomega frameworks better, updated the tests. Can we now move on with it to unblock the openstack-operator patch?.

modules/common/condition/funcs.go
gibizer: I think c.Has() can be kept. and a new c.IndexOf() can be added that returns -1 or the index of the item. But it can be done independently.

gibizer (approved):
Looks good.

PR #159, opened 2023-01-31
Add GetServiceWithName and GetServicesPortDetails

stuggi:
Adds functions to set a Service with name in namespace and one to get a ServicePort with its name from the service.

### DIFF
--- a/modules/common/service/service.go
+++ b/modules/common/service/service.go
@@ -25,6 +25,7 @@ import (
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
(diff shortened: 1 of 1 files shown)


modules/common/service/service.go
```diff
@@ -176,3 +177,40 @@ func GetServicesListWithLabel(
...
+		return nil, err
+	}
+
+	return service, nil
+}
+
+// GetServicesPortDetails - Return ServicePort with name from a Service
+func GetServicesPortDetails(
+	ctx context.Context,
+	h *helper.Helper,
```
abays, line 204: Doesn't look like we need the context or the helper here

stuggi: right, will remove it

modules/common/service/service.go
```diff
@@ -176,3 +177,38 @@ func GetServicesListWithLabel(
...
+func GetServiceWithName(
+	ctx context.Context,
+	h *helper.Helper,
+	name string,
+	namespace string,
+) (*corev1.Service, error) {
+
+	service := &corev1.Service{}
+
+	err := h.GetClient().Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, service)
```
abays, line 191: Just one question.  You mentioned in the other PR (https://github.com/openstack-k8s-operators/lib-common/pull/158) that you wanted to use KClient to not be restricted to cache-only pods in watched namespaces.  Do we have any interest in doing that here too?

abays (approved):
/lgtm

PR #160, opened 2023-01-31
Add MetalLB service type

stuggi:
Adds a MetalLBService type to create its LoadBalancer service. Also extends the endpoint Data with MetalLBData with the required input to create the MetalLB service.

### DIFF
--- a/modules/common/endpoint/endpoint.go
+++ b/modules/common/endpoint/endpoint.go
@@ -48,6 +48,22 @@ type Data struct {
 	Port int32
 	// An optional path suffix to append to route hostname when forming Keystone endpoint URLs
 	Path string
+	// details for metallb service generation
+	MetalLB *MetalLBData
+}
+
+// MetalLBData - information specific to creating the MetalLB service
+type MetalLBData struct {
+	// Name of the metallb IpAddressPool
+	IPAddressPool string
+	// use shared IP for the service
+	SharedIP bool
+	// sharing key which gets set as the annotation on the LoadBalancer service.
+	// Services which share the same VIP must have the same SharedIPKey. Gets default to the IPAddressPool if
+	// SharedIP is true, but no SharedIPKey set.
+	SharedIPKey string
+	// if set request these IPs via MetalLBLoadBalancerIPs, using a list for dual stack (ipv4/ipv6)
+	LoadBalancerIPs []string
 }
 
 // ExposeEndpoints - creates services, routes and returns a map of created openstack endpoint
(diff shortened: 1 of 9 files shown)


stuggi:
example usage in https://github.com/openstack-k8s-operators/keystone-operator/pull/175

stuggi:
# openstack endpoint list
+----------------------------------+-----------+--------------+--------------+---------+-----------+-----------------------------------------------------------------+
| ID                               | Region    | Service Name | Service Type | Enabled | Interface | URL                                                             |
+----------------------------------+-----------+--------------+--------------+---------+-----------+-----------------------------------------------------------------+
| 040fa92602564b2c9421bd6d375683d4 | regionOne | keystone     | identity     | True    | internal  | http://keystone-internal.openstack.svc:5000                     |
| 6947e88973714c20a42266eb49d41d8c | regionOne | keystone     | identity     | True    | public    | http://keystone-public-openstack.apps.ostest.test.metalkube.org |
+----------------------------------+-----------+--------------+--------------+---------+-----------+-----------------------------------------------------------------+

gibizer:
The code looks good to me but I have limited knowledge about metallb

abays:
> The code looks good to me but I have limited knowledge about metallb

Same here

stuggi:
> /hold . will submit PRs for the services I have integrated so far. when have general agreement, we'll merge this PR and update them

ah wait we don't have automation to merge here :)

modules/common/service/service.go
```diff
@@ -105,6 +151,16 @@ func (s *Service) CreateOrPatch(
 		h.GetLogger().Info(fmt.Sprintf("Service %s - %s", service.Name, op))
 	}
 
+	// update the service instance with the ip/host information
+	s.clusterIP = service.Spec.ClusterIP
+	s.serviceHostname = fmt.Sprintf("%s.%s.svc", service.Name, service.GetNamespace())
+
+	if service.Spec.Type == corev1.ServiceTypeLoadBalancer {
+		if len(service.Status.LoadBalancer.Ingress) > 0 {
+			s.externalIP = service.Status.LoadBalancer.Ingress[0].IP
```
olliewalsh, line 160: can this have multiple IPs?

modules/common/service/service.go
```diff
@@ -70,6 +90,32 @@ func GenericService(svcInfo *GenericServiceDetails) *corev1.Service {
...
+				{
+					Name: svcInfo.Port.Name,
+					Port: svcInfo.Port.Port,
+					// corev1.ProtocolTCP/ corev1.ProtocolUDP/ corev1.ProtocolSCTP
+					// - https://pkg.go.dev/k8s.io/api@v0.23.6/core/v1#Protocol
+					Protocol: svcInfo.Port.Protocol,
+				},
+			},
+			Type:           corev1.ServiceTypeLoadBalancer,
+			LoadBalancerIP: svcInfo.LoadBalancerIP,
```
stuggi, line 114: the `spec.LoadBalancerIP` is deprecated https://github.com/kubernetes/kubernetes/pull/107235 . we should switch to use the `metallb.universe.tf/loadBalancerIPs` instead. will update the PR

modules/common/service/service.go
stuggi: I guess in case of dualstack https://metallb.universe.tf/usage/#ipv6-and-dual-stack-services

stuggi: probably should make this externalPs for this case

abays (commented):
Are we still holding this?  If you've tested it and it works as expected, +1 from me

stuggi:
@abays I think this is good to go now. could you please have another look.

modules/common/networkattachment/networkattachment.go
```diff
@@ -0,0 +1,125 @@
...
+	serviceLabels map[string]string,
+	readyCount int32,
+) (bool, map[string][]string, error) {
+
+	networkReady := true
+	networkAttachmentStatus := map[string][]string{}
+	if len(networkAttachments) > 0 {
+		podList, err := pod.GetPodListWithLabel(ctx, helper, helper.GetBeforeObject().GetNamespace(), serviceLabels)
+		if err != nil {
+			return networkReady, networkAttachmentStatus, err
```
abays, line 104: Should this maybe return false instead of the default `networkReady = true` in the case of an error?  I guess it doesn't really matter though since the caller would probably be checking the error first before even considering the boolean.

modules/common/networkattachment/networkattachment_test.go
```diff
@@ -0,0 +1,141 @@
...
+func TestCreateNetworksAnnotation(t *testing.T) {
+
+	tests := []struct {
+		name      string
+		networks  []string
+		namespace string
+		want      map[string]string
+	}{
+		{
+			name:      "Single network",
```
abays, line 36: Should this be perhaps "no network" instead?  Just trying to differentiate from the next test labelled `Single network`.

modules/common/service/service.go
```diff
@@ -46,6 +46,26 @@ func NewService(
 	}
 }
 
+// GetClusterIP - returns the cluster IP of the created service
+func (s *Service) GetClusterIP() []string {
```
abays, line 50: Should the name be plural here since it returns a string list of IPs?

modules/common/networkattachment/networkattachment.go
stuggi: done

modules/common/networkattachment/networkattachment.go
```diff
@@ -0,0 +1,125 @@
...
+				return networkReady, networkAttachmentStatus, err
+			}
+			for _, netStat := range netsStatus {
+				networkAttachmentStatus[netStat.Name] = append(networkAttachmentStatus[netStat.Name], netStat.IPs...)
+			}
+		}
+
+		for _, net := range networkAttachmentStatus {
+			if len(net) < int(readyCount) {
+				networkReady = false
```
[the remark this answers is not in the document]
stuggi: done

modules/common/networkattachment/networkattachment_test.go
stuggi: done

modules/common/service/service.go
stuggi: done

abays (approved):
/lgtm

modules/common/endpoint/endpoint.go
```diff
@@ -70,58 +82,102 @@ func ExposeEndpoints(
...
+		var hostname string
+		if data.MetalLB != nil {
+			annotations := map[string]string{
+				service.MetalLBAddressPoolAnnotation: data.MetalLB.IPAddressPool,
+			}
+			if len(data.MetalLB.LoadBalancerIPs) > 0 {
+				annotations[service.MetalLBLoadBalancerIPs] = strings.Join(data.MetalLB.LoadBalancerIPs, ",")
+			}
+			if data.MetalLB.SharedIP {
+				annotations[service.MetalLBAllowSharedIPAnnotation] = data.MetalLB.IPAddressPool + "-vip"
```
olliewalsh, line 96: Why append `-vip` to the pool name?

modules/common/service/service.go
```diff
@@ -105,6 +150,18 @@ func (s *Service) CreateOrPatch(
 		h.GetLogger().Info(fmt.Sprintf("Service %s - %s", service.Name, op))
 	}
 
+	// update the service instance with the ip/host information
+	s.clusterIPs = service.Spec.ClusterIPs
+	s.serviceHostname = fmt.Sprintf("%s.%s.svc", service.Name, service.GetNamespace())
+
+	if service.Spec.Type == corev1.ServiceTypeLoadBalancer {
+		if len(service.Status.LoadBalancer.Ingress) > 0 {
```
olliewalsh, line 158: Can it take some time for the service status to update?

modules/common/endpoint/endpoint.go
```diff
@@ -70,58 +82,102 @@ func ExposeEndpoints(
...
+				return endpointMap, ctrlResult, err
+			} else if (ctrlResult != ctrl.Result{}) {
+				return endpointMap, ctrlResult, nil
+			}
+			// create service - end
+
+			hostname = svc.GetServiceHostnamePort()
+		} else {
+
+			// Create the service if none exists
```
olliewalsh, line 126: Is that correct? Looks like it will patch an existing service too

stuggi: will update the comment

modules/common/service/service.go
stuggi: it probably could, right now we are not using GetExternalIPs(). if we see there are issues, GetExternalIPs() could be updated to query the service.

modules/common/endpoint/endpoint.go
```diff
@@ -70,58 +86,107 @@ func ExposeEndpoints(
...
+		var hostname string
+		if data.MetalLB != nil {
+			annotations := map[string]string{
+				service.MetalLBAddressPoolAnnotation: data.MetalLB.IPAddressPool,
+			}
+			if len(data.MetalLB.LoadBalancerIPs) > 0 {
+				annotations[service.MetalLBLoadBalancerIPs] = strings.Join(data.MetalLB.LoadBalancerIPs, ",")
+			}
+			if data.MetalLB.SharedIP {
+				if data.MetalLB.SharedIPKey == "" {
```
olliewalsh, line 100: should that be `!=`?

olliewalsh (changes requested):

olliewalsh (approved):
/lgtm

PR #165, opened 2023-02-08
Add finalizer on MariaDB instance

rabi:
This would ensure that MariaDB instance can't
be deleted before the services. Adds dbInstance
field to Database struct and then adds/removes
finalizer on it for the service as required.

### DIFF
--- a/modules/database/funcs.go
+++ b/modules/database/funcs.go
@@ -69,10 +69,8 @@ func NewDatabaseWithNamespace(
 	}
 }
 
-//
 // setDatabaseHostname - set the service name of the DB as the databaseHostname
 // by looking up the Service via the name of the MariaDB CR which provides it.
-//
 func (d *Database) setDatabaseHostname(
 	ctx context.Context,
 	h *helper.Helper,
@@ -114,26 +112,20 @@ func (d *Database) setDatabaseHostname(
 	return nil
 }
 
-//
 // GetDatabaseHostname - returns the DB hostname which host the DB
-//
 func (d *Database) GetDatabaseHostname() string {
 	return d.databaseHostname
 }
 
-//
 // GetDatabase - returns the DB
-//
 func (d *Database) GetDatabase() *mariadbv1.MariaDBDatabase {
 	return d.database
 }
 
-//
 // CreateOrPatchDB - create or patch the service DB instance
 // Deprecated. Use CreateOrPatchDBByName instead. If you want to use the
 // default the DB service instance of the deployment then pass "openstack" as
 // the name.
-//
 func (d *Database) CreateOrPatchDB(
 	ctx context.Context,
 	h *helper.Helper,
@@ -141,11 +133,9 @@ func (d *Database) CreateOrPatchDB(
 	return d.CreateOrPatchDBByName(ctx, h, "openstack")
 }
 
-//
 // CreateOrPatchDBByName - create or patch the service DB instance on
 // the DB service. The DB service is selected by the name of the MariaDB CR
 // providing the service.
-//
 func (d *Database) CreateOrPatchDBByName(
 	ctx context.Context,
 	h *helper.Helper,
@@ -192,6 +182,15 @@ func (d *Database) CreateOrPatchDBByName(
 			return err
 		}
 
+		err = d.getDBInstance(ctx, h)
+		if err != nil {
+			return err
+		}
+
+		controllerutil.AddFinalizer(d.dbInstance, fmt.Sprintf("%s-%s", h.GetFinalizer(), db.Name))
+		if err := h.GetClient().Update(ctx, d.dbInstance); err != nil && !k8s_errors.IsNotFound(err) {
+			return err
+		}
 		// If the service object doesn't have our finalizer, add it.
 		controllerutil.AddFinalizer(db, h.GetFinalizer())
 
@@ -222,9 +221,7 @@ func (d *Database) CreateOrPatchDBByName(
 	return ctrl.Result{}, nil
 }
 
-//
 // WaitForDBCreatedWithTimeout - wait until the MariaDBDatabase is initialized and reports Status.Completed == true
-//
 func (d *Database) WaitForDBCreatedWithTimeout(
 	ctx context.Context,
 	h *helper.Helper,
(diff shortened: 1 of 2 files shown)


modules/database/funcs.go
```diff
@@ -262,9 +258,44 @@ func (d *Database) WaitForDBCreated(
...
+		ctx,
+		types.NamespacedName{
+			Namespace: namespace,
+			Name:      "openstack",
+		},
+		dbInstance)
+	if err != nil {
+		if k8s_errors.IsNotFound(err) {
+			return util.WrapErrorForObject(
+				fmt.Sprintf("Failed to get db instance in %s", namespace),
```
gibizer, line 282: I would call this MariaDB instance as `db` is pretty overloaded term.

modules/database/funcs.go
```diff
@@ -262,9 +258,44 @@ func (d *Database) WaitForDBCreated(
...
+				h.GetBeforeObject(),
+				err,
+			)
+		}
+		return util.WrapErrorForObject(
+			fmt.Sprintf("Failed to get db instance in %s", namespace),
```
gibizer, line 288: ditto

modules/database/funcs.go
```diff
@@ -262,9 +258,44 @@ func (d *Database) WaitForDBCreated(
...
+	dbInstance := &mariadbv1.MariaDB{}
+	err := h.GetClient().Get(
+		ctx,
+		types.NamespacedName{
+			Namespace: namespace,
+			Name:      "openstack",
+		},
+		dbInstance)
+	if err != nil {
+		if k8s_errors.IsNotFound(err) {
```
gibizer, line 280: It seems the code in the if branch is the same as the code outside of the if branch so you can merge them.

gibizer (changes requested):

modules/database/funcs.go
rabi: AFAIK db instance vs db/database is not overloaded. We're not using the 'instance' in the programming sense here but as used with databases.

MariaaDB - db instance (database instance)
MariaDBDatabase - db, database

abays:
I think this approach can work, but I am curious as to why we don't use the actual `MariaDBDatabase` controller to add the finalizer to the `MariaDB` instance.  We acquire the associated `MariaDB` instance during `MariaDBDatabase` reconcile [1], so we could either add or remove the finalizer based on the `MariaDBDatabase.DeletionTimestamp.IsZero()` flag.  Anyhow, it's just a thought to consider.

[1] https://github.com/openstack-k8s-operators/mariadb-operator/blob/4970c77d3d00a6667d91ab66e8f55ee2b697f46d/controllers/mariadbdatabase_controller.go#L84-L97

gibizer:
> I think this approach can work, but I am curious as to why we don't use the actual `MariaDBDatabase` controller to add the finalizer to the `MariaDB` instance. We acquire the associated `MariaDB` instance during `MariaDBDatabase` reconcile [1], so we could either add or remove the finalizer based on the `MariaDBDatabase.DeletionTimestamp.IsZero()` flag. Anyhow, it's just a thought to consider.
> 
> [1] https://github.com/openstack-k8s-operators/mariadb-operator/blob/4970c77d3d00a6667d91ab66e8f55ee2b697f46d/controllers/mariadbdatabase_controller.go#L84-L97

Good point. I think ti is better to do it in the mariadbdatabase_controller.

rabi:
> but I am curious as to why we don't use the actual MariaDBDatabase controller to add the finalizer to the MariaDB instance

That does not work. Initially i did think about that and tried it. I guess it's probably because during KeystoneAPI deleteReconcile() we don't make an explicit API call to delete the database[1]. Removing the custom finalizer from MariaDB based on MariaDBDatabase.DeletionTimestamp.IsZero() does not work.

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/master/controllers/keystoneapi_controller.go#L202

rabi:
> Unless the k8s-automated delete-by-owner-ref just doesn't set the deletion timestamp for some reason.

reconcile() isn't even called for MariaDBDatabase which confused me too. Also, I thought doing in in deleteFinalizer() for the database is better as that is for "deleting all finalizers" on that object.

abays:
> > reconcile() isn't even called for MariaDBDatabase which confused me too
> 
> I feel this is a better way to implement. But if you guys think otherwise let me know and I can re-verify that again and see if I was doing something wrong.

@rabi @gibizer Here is an approach to doing this via `MariaDBDatabase` controller [1].  I tested it with Keystone and it seemed to work.

[1] https://github.com/openstack-k8s-operators/mariadb-operator/compare/master...abays:finalizers?expand=1

rabi:
> I tested it with Keystone and it seemed to work.

Yeah seems to work. I was noticing the instance.DeletionTimestamp.IsZero() returning true during delete reconciliation via ownerref, but this seems to work (probably missing something). I'm fine if we want to go with this as it won't require changes to other operators for rbac.

gibizer:
> > > reconcile() isn't even called for MariaDBDatabase which confused me too
> > 
> > 
> > I feel this is a better way to implement. But if you guys think otherwise let me know and I can re-verify that again and see if I was doing something wrong.
> 
> @rabi @gibizer Here is an approach to doing this via `MariaDBDatabase` controller [1]. I tested it with Keystone and it seemed to work.
> 
> [1] https://github.com/openstack-k8s-operators/mariadb-operator/compare/master...abays:finalizers?expand=1

I would go with this ^^. Could you open a PR for it?

abays:
> > > > reconcile() isn't even called for MariaDBDatabase which confused me too
> > > 
> > > 
> > > I feel this is a better way to implement. But if you guys think otherwise let me know and I can re-verify that again and see if I was doing something wrong.
> > 
> > 
> > @rabi @gibizer Here is an approach to doing this via `MariaDBDatabase` controller [1]. I tested it with Keystone and it seemed to work.
> > [1] https://github.com/openstack-k8s-operators/mariadb-operator/compare/master...abays:finalizers?expand=1
> 
> I would go with this ^^. Could you open a PR for it?

Sure, I'll do that shortly

rabi:
We're going with  the mariadb-operator change proposed by @abays.

PR #168, opened 2023-02-09
Fix util pkg use of path.Join()

abays:
We were mistakenly concating the `template` path with the `filename` before passing it to `path.Join()`, which resulted in errors like so (discovered by running MariaDB operator locally via `OPERATOR_TEMPLATES=./templates make run`):

```
"error": "open templatesdatabase.sh: no such file or directory"
```

Instead the `template` and `filename` args should be passed as separate items.


stuggi (approved):
/lgtm

PR #172, opened 2023-02-15
Add utility to generate random string

kajinamit:
This introduces a utility function to generate a random string. This can be used to generate some hash keys such as horizon secret key.

### DIFF
--- a/modules/common/util/string.go
+++ b/modules/common/util/string.go
@@ -16,6 +16,11 @@ limitations under the License.
 
 package util
 
+import (
+	"math/rand"
+	"time"
+)
+
 // StringInSlice - is string in slice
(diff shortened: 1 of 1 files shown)


fao89 (approved):

modules/common/util/string.go
```diff
@@ -25,3 +30,15 @@ func StringInSlice(a string, list []string) bool {
 	}
 	return false
 }
+
+// RandomString - generate a random string of given length
+func RandomString(length int) []byte {
```
stuggi, line 35: could also just use https://pkg.go.dev/k8s.io/apimachinery/pkg/util/rand#String ?

kajinamit: Ah, that makes sense. Let me try it.

PR #174, opened 2023-02-16
Run go work sync during make gowork

gibizer:
This will ensure that the go.mod files in the different modules will not have different versions from the same dependency. Such divergence known to cause troubles when generating manifests in different environments and while lib-common does not provide any CRDs we might have some common structs here in the future.


fao89 (approved):

modules/archive/go.mod
```diff
@@ -1,55 +1,52 @@
 module github.com/openstack-k8s-operators/lib-common/modules/archive
 
-go 1.19
```
bogdando, line 3: this doesn't look expected

bogdando (changes requested):

modules/storage/go.mod
```diff
@@ -1,28 +1,30 @@
 module github.com/openstack-k8s-operators/lib-common/modules/storage
 
-go 1.19
```
bogdando, line 3: ditto

stuggi (approved):
/lgtm

PR #176, opened 2023-02-16
Allow setting of deployment strategy

abays:
Currently needed for upcoming changes to `openstack-baremetal-operator`


stuggi (approved):
/lgtm

modules/common/deployment/deployment.go
```diff
@@ -64,6 +64,7 @@ func (d *Deployment) CreateOrPatch(
 		deployment.Labels = util.MergeStringMaps(deployment.Labels, d.deployment.Labels)
 		deployment.Spec.Template = d.deployment.Spec.Template
 		deployment.Spec.Replicas = d.deployment.Spec.Replicas
+		deployment.Spec.Strategy = d.deployment.Spec.Strategy
```
gibizer, line 67: It is so strange the we need to do field copy one by one. But yeah I needed to do similar for StatefulSet as well so I see the reason.

gibizer (approved):

PR #180, opened 2023-02-17
Move ansible module from AEE

fao89:
1 - https://github.com/openstack-k8s-operators/lib-common/pull/180
2 - https://github.com/openstack-k8s-operators/dataplane-operator/pull/39
3 - https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/89

### DIFF
--- a/modules/ansible/go.mod
+++ b/modules/ansible/go.mod
@@ -1,0 +1,5 @@
+module github.com/openstack-k8s-operators/lib-common/ansible
+
+go 1.19
+
+require gopkg.in/yaml.v3 v3.0.1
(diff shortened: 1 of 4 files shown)


fultonj:
Is the idea to put the ansible module in lib-common since other operators besides openstack-ansibleee-operator will need it?

Originally I thought they wouldn't since we could just create a `OpenStackAnsibleEE` CR from any operator. However, we do have cases of using the ansible module outside of that interface, e.g. calling the ansible module libraries to build an inventory as we see here:

  https://github.com/openstack-k8s-operators/dataplane-operator/pull/24

For that reason, I'd be OK with this. Any other reasons for this change?

fao89:
when you need to import `ansible` and `api` modules 
```
github.com/openstack-k8s-operators/openstack-ansibleee-operator
github.com/openstack-k8s-operators/openstack-ansibleee-operator/api
```
https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/go.mod#L14-L15
the first one tries to get the local `api`:
https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/blob/main/go.mod#L5-L13
and it breaks `make gowork`:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_dataplane-operator/37/pull-ci-openstack-k8s-operators-dataplane-operator-main-precommit-check/1626614873850908672#1:build-log.txt%3A116

slagle (approved):

PR #185, opened 2023-02-20
remove unused MetalLBServiceDetails.LoadBalancerIPs

stuggi:


gibizer (approved):
As we discussed on slack this is a leftover from an earlier solution but unused now.

PR #187, opened 2023-02-20
improve VerifyNetworkStatusFromAnnotation to check against spec.networkAttachments

stuggi:


gibizer (approved):
Tested it with the nova-operator envtest coverage for the networkAttachments and it works as expected.

PR #189, opened 2023-02-20
Add renovate.json / Remove dependabot.yml

dprince:


renovate.json
```diff
@@ -0,0 +1,44 @@
...
+    },
+    {
+      "groupName": "k8s.io",
+      "matchPackagePatterns": ["^k8s.io", "^sigs.k8s.io"],
+      "allowedVersions": "< 1.0.0",
+      "enabled": true
+    },
+    {
+      "groupName": "misc",
+      "matchPackagePatterns": ["^github.com/operator-framework/api", "^github.com/ghodss", "^github.com/go-logr/logr", "^github.com/imdario/mergo", "^go.uber.org/zap"],
```
gibizer, line 35: Is there a specific reason we group these together?

dprince: just similar to what we did in other repos. Could remove it, or not

gibizer: I guess it is as good as any other arbitrary grouping. So lets land this and we can improve on it later if we found a better one.

gibizer (approved):

PR #200, opened 2023-02-27
Fix local testing without $OPERATOR_TEMPLATES

rabi:
Currently it does not work (probably since migrating these to common-lib modules). GetTemplatesPath() function signature has been changed. Looks like it's not called externally yet, hence there would be no impact.


stuggi (approved):
/lgtm

PR #201, opened 2023-02-28
Ensures that the Secret object exists and the expected fields

fao89:


modules/common/condition/types.go
```diff
@@ -84,3 +84,9 @@ type conditionGroup struct {
 	severity   Severity
 	conditions Conditions
 }
+
+// Updater - generic type for helping to update conditions
+type Updater interface {
+	Set(c *Condition)
+	MarkTrue(t Type, messageFormat string, messageArgs ...interface{})
+}
```
gibizer, line 92: This is the point where my golang knowledge is lacking. Can we somehow make our Condition type somehow closer / aligned with the interface we need here to manipulate that Condition type?

modules/common/secret/secret.go
```diff
@@ -385,3 +386,70 @@ func GetDataFromSecret(
...
+	reader client.Reader,
+	Updater condition.Updater,
+	requeueTimeout time.Duration,
+) (string, ctrl.Result, error) {
+	secret := &corev1.Secret{}
+	err := reader.Get(ctx, secretName, secret)
+	if err != nil {
+		if k8s_errors.IsNotFound(err) {
+			// TODO(gibi): Change the message to state which input
+			// (i.e. Secret with a given name) is missing
```
gibizer, line 405: Lets resolve this TODO. I have the resolution in nova as part of a longer draft series.  https://github.com/openstack-k8s-operators/nova-operator/pull/259/commits/a6d3d2df1a278a28d293896ce40bbb87b5afe780#diff-da55647121285e1a60376c19fef7ce508bd29edddabac15bc6c8781be47565ee 

However it needs to change condition.InputReadyWaitingMessage to allow incorporating the missing input to the message. And that is a breaking change if done in lib-common

modules/common/secret/secret.go
```diff
@@ -385,3 +386,70 @@ func GetDataFromSecret(
...
+				condition.ErrorReason,
+				condition.SeverityWarning,
+				condition.InputReadyErrorMessage,
+				err.Error()))
+			return "", ctrl.Result{}, err
+		}
+		values = append(values, val)
+	}
+
+	// TODO(gibi): Do we need to watch the Secret for changes?
```
gibizer, line 441: While it was still in nova only I was OK to keep this TODO in place, but as soon as we move this to lib-common this becomes complicated. The goal is valid to eventually watch the Secret for changes so that we can pick up when the missing field is eventually provided. But we cannot easily add such watch via this lib-common function. So the TODO will not be resolved here but probably in every individual operator.

So at least I would change this TODO here to express this.

gibizer (commented):
thanks for proposing this

modules/common/secret/secret.go
SeanMooney: this does not really match or naming convention by the way.

we use ensure to mean validate and create/update if not correct.
this is a bug in the nova operator naming of this function.

this is actually VerifySecret

since this checks that the secret exists and that the required field are present but it will never ensure the value are set to a specific value.

can we rename it to VerifySecret

modules/common/secret/secret.go
SeanMooney: so in nova yes we do but this is out os the scope fo this function

we should have a separate function to do this if needed.

the comment should be removed i think.

modules/common/secret/secret.go
```diff
@@ -385,3 +386,70 @@ func GetDataFromSecret(
...
 	return data, ctrl.Result{}, nil
 }
+
+// EnsureSecretData - ensures that the Secret object exists and the expected fields
+// are in the Secret. It returns a hash of the values of the expected fields.
+func EnsureSecretData(
+	ctx context.Context,
+	secretName types.NamespacedName,
+	expectedFields []string,
+	reader client.Reader,
```
stuggi, line 396: could just pass in a helper `h *helper.Helper`  which has the client.

stuggi: it also has a logger

modules/common/secret/secret.go
```diff
@@ -385,3 +386,70 @@ func GetDataFromSecret(
...
+	Updater condition.Updater,
+	requeueTimeout time.Duration,
+) (string, ctrl.Result, error) {
+	secret := &corev1.Secret{}
+	err := reader.Get(ctx, secretName, secret)
+	if err != nil {
+		if k8s_errors.IsNotFound(err) {
+			// TODO(gibi): Change the message to state which input
+			// (i.e. Secret with a given name) is missing
+			Updater.Set(condition.FalseCondition(
```
stuggi, line 406: I am not sure if we want to update the condition in the lib-common funcs. It might better/easier to just return the error message and to the condition update in the caller, but want to hear other peoples opinions.

We did something like that in the OSP Director Operator where we set condition on a more fine level, but need to be careful to not have to detailed and maybe changing in information in the message. This can then result in unnecessary object updates as the status changes without the need.

modules/common/secret/secret.go
gibizer: It is fine by me to rename it.

modules/common/secret/secret.go
gibizer: But if we remove the TODO comment here then we loose the TODO without resolving it. This is why I'm torn on this.

modules/common/secret/secret.go
gibizer: > could just pass in a helper `h *helper.Helper` which has the client.

I would rather pass the interface than the full Helper. The Helper is bigger surface than what we need here.

gibizer: > it also has a logger

I would rater use logger.FromContext(ctx) to get a logger. As we discussed before that logger has everything we need.

modules/common/secret/secret.go
gibizer: Good point. I think this would be a first time moving condition updating to lib-common. So we need to think about it.

modules/common/secret/secret.go
stuggi: agree with rename it to VerifySecret

modules/common/secret/secret.go
stuggi: right, I did not mean to use the logging funcs from lib-common. Just that the helper already has an initialized logger from the reconcile instance, which could be `logger.FromContext(ctx)` . Also just wanted to mention that the helper has the client and the logger which gets passed via dedicated params, so in general its possible to pass the helper instead of list of dedicated params.

modules/common/secret/secret.go
stuggi: I think we should have a convention base on labels for CM/Secret objects to watch which one does not own. Is this what is meant by this comment? The creator would have to set the label/labels, could be something like `global-watcher` which all operators watch, `glance-watcher`, `nova-watcher` for service specific

SeanMooney: we can keep the todo in the nova code when we adapt to using it form lib common.

@stuggi 
well we could do that or we can just watch the secrets/configmaps that are provided as inputs to the service crs.

it would be nicer if we could subscribe to the one we cared about from the input set rather then requiring the OpenStack operator to set those labels.

fao89: I don't know exactly how to do it in go, but maybe we can turn this into an abstract class where people would need to implement the method for watching the secret

modules/common/secret/secret.go
fao89: should I remove the conditions?

modules/common/secret/secret.go
fao89: found this: https://golangbyexample.com/go-abstract-class/

modules/common/secret/secret.go
gibizer: In general I don't like the Helper pattern as it couples unrelated things. Passing around the Helper means the function gets coupled to things that it does not need. I think the general good pattern would be to define these function parameters as interfaces. That will allow passing any object that matches the interface. So if it is called from a place that has just a client than it can pass that, but if it is called from a place that has a Helper it can still call the function by passing the client from the Helper to it.

modules/common/secret/secret.go
gibizer: I'm torn. Half of the logic we generalizing here it the actual condition setting logic. So by removing that we force the callers to duplicate condition handling logic in each operator. For me the question is do we know that this condition setting logic is generic for all operators that depends on Secrets? If so then I think it is fine to put the logic into lib-common.

modules/common/secret/secret.go
SeanMooney: i agree with gibi here.

i would prefer to have an interface not use the helper stuct.
that applies to all our apis i don't think the helper stuct should really be in them in general even if that is what we pass.

on of the advantages of the interface apptoch si it will acutly be possible to implement a mock object that conform to the  interface and unit test the code.

that is much harder to do with concrete types.

modules/common/secret/secret.go
gibizer: > I think we should have a convention base on labels for CM/Secret objects to watch which one does not own. Is this what is meant by this comment? The creator would have to set the label/labels, could be something like `global-watcher` which all operators watch, `glance-watcher`, `nova-watcher` for service specific

I haven't had time to dig into solving the TODO itself so I don't know what would be a good solution. I assume that as the given operator gets the name of the Secret via the CRD field the easiest way would be to watch the Secret by name. But I don't know if that is possible or practical.

modules/common/secret/secret.go
SeanMooney: i would remove the condtion updating for now and just pass back the error as martin suggested.

we can always creat a generic function that will take a condition and a function to call and set the condition if an error is return or something like that later. and its not hard to refactor nova to just do the condition setting on our side so we can defer that to the caller.

stuggi: I'd remove setting the condition and set them, like we do it right now in the controller of the operator as you might not want to get a condition set in all the cases you'd use this func.

modules/common/secret/secret.go
gibizer: @stuggi thanks, that shows some downside of our CRD designs. So if it is hard to watch secrets by name as the name is dynamic and the watch is easiest to set statically in `SetupWithManager` then I think we should reconsider passing Secret by name to our CRDs and come up with something better suited for watching for changes.

modules/common/secret/secret.go
gibizer: OK, I'm fine with removing the condition settings from here.

modules/common/secret/secret.go
stuggi: ah you can also do it with a spec field like here https://book.kubebuilder.io/reference/watching-resources/externally-managed.html

gibizer: @stuggi cool then we can keep the Secret name field on our CRD and set up a watcher still.

Back to the original topic. I think I'm OK with removing the TODO from here and adding a sentence to the doc of this function stating that it is the responsibility of the caller to make sure it sets up a watcher for the Secret to get informed about any changes.

SeanMooney (commented):
i am ok with this version
lets add the same for configmaps but that can be a separate pr if we want to proceed with this as is

fao89:
> i am ok with this version lets add the same for configmaps but that can be a separate pr if we want to proceed with this as is

I started this one: https://github.com/openstack-k8s-operators/lib-common/pull/202

modules/common/secret/secret.go
```diff
@@ -385,3 +385,42 @@ func GetDataFromSecret(
...
+	reader client.Reader,
+	requeueTimeout time.Duration,
+) (string, ctrl.Result, error) {
+	secret := &corev1.Secret{}
+	err := reader.Get(ctx, secretName, secret)
+	if err != nil {
+		if k8s_errors.IsNotFound(err) {
+			return "",
+				ctrl.Result{RequeueAfter: requeueTimeout},
+				fmt.Errorf("OpenStack secret %s not found", secretName)
```
gibizer, line 404: Sorry for noticing just now, but can we please remove OpenStack from it. This is here just because in the nova version we used this method first to check for the osp-secret. But this can be any secret now.

gibizer (changes requested):

modules/common/secret/secret.go
```diff
@@ -385,3 +385,42 @@ func GetDataFromSecret(
...
+		}
+		return "", ctrl.Result{}, err
```
stuggi, line 406: while it might be not that important in this small func, I think we should wrap the errors we return to be easier to identify when debugging: `fmt.Errorf("get secret %s failed: %w", secretName, err)`. using `%w` makes it available to errors.Is and errors.As

modules/common/secret/secret.go
```diff
@@ -385,3 +385,42 @@ func GetDataFromSecret(
...
+		if !ok {
+			err := fmt.Errorf("field %s not found in Secret %s", field, secretName)
+			return "", ctrl.Result{}, err
+		}
+		values = append(values, val)
+	}
+
+	hash, err := util.ObjectHash(values)
+	if err != nil {
+		return "", ctrl.Result{}, err
```
stuggi, line 422: `util.ObjectHash()` already wraps this. we should just update that func to use `%w` at some point

modules/common/secret/secret.go
gibizer: +1 lets wrap the errors

gibizer (approved):
looks good to me

SeanMooney (approved):

stuggi (approved):
/lgtm Thanks!

PR #203, opened 2023-03-02
Validate if secretName is empty

fao89:


fao89:
this isn't needed
https://cs.github.com/?scopeName=All+repos&scope=&q=org%3Akubernetes+%22name+must+be+specified%22

PR #204, opened 2023-03-06
Mirror order conditions as False, Unknown, True

gibizer:
This is a breaking change on Conditions.Mirror(). Originally this call ordered conditions as False, True, Unknown then picked the first of that list to reflect. However all the use cases (we see) assume that if there is reasourceA that depends on the status of resourceB then resourceA reconciler calls:
```golang
  resourceB.Status.Conditions.Mirror(ResourceBSpecificSubConditionOfA)
```
and expects that the call returns the most sever non True condition if any or return a True condition otherwise. So the original behavior that returned True even if there was Unknown condition was unexpected.

This patch changes the order of the status to False, Unknown, True. But keeps the original logic that if the ReadyCondition is True then the rest of the conditions are ignored even if there are False conditions in the list.

### DIFF
--- a/modules/common/condition/funcs_test.go
+++ b/modules/common/condition/funcs_test.go
@@ -386,11 +386,11 @@ func TestMirror(t *testing.T) {
 	conditions.Set(trueA)
 	g.Expect(conditions).To(haveSameConditionsOf(CreateList(unknownReady, trueA)))
 	targetCondition = conditions.Mirror("targetConditon")
-	// expect to be mirrored trueA
-	g.Expect(targetCondition.Status).To(BeIdenticalTo(trueA.Status))
-	g.Expect(targetCondition.Severity).To(BeIdenticalTo(trueA.Severity))
-	g.Expect(targetCondition.Reason).To(BeIdenticalTo(trueA.Reason))
-	g.Expect(targetCondition.Message).To(BeIdenticalTo(trueA.Message))
+	// expect to be mirrored unknownReady
+	g.Expect(targetCondition.Status).To(BeIdenticalTo(unknownReady.Status))
+	g.Expect(targetCondition.Severity).To(BeIdenticalTo(unknownReady.Severity))
+	g.Expect(targetCondition.Reason).To(BeIdenticalTo(unknownReady.Reason))
+	g.Expect(targetCondition.Message).To(BeIdenticalTo(unknownReady.Message))
 
 	conditions.Set(falseB)
 	g.Expect(conditions).To(haveSameConditionsOf(CreateList(unknownReady, trueA, falseB)))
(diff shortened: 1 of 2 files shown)


gibizer:
/hold 
I want to see that the behavior change I noticed in nova-operator after this lib-common change are OK https://github.com/openstack-k8s-operators/nova-operator/pull/283

gibizer:
/unhold
https://github.com/openstack-k8s-operators/nova-operator/pull/283 now shows that nova-operator actually have a bug before this fix. And after this fix the test needs to be adapted as it made wrong assumptions.

modules/common/condition/funcs_test.go
```diff
@@ -387,10 +387,10 @@ func TestMirror(t *testing.T) {
 	g.Expect(conditions).To(haveSameConditionsOf(CreateList(unknownReady, trueA)))
 	targetCondition = conditions.Mirror("targetConditon")
 	// expect to be mirrored trueA
```
stuggi, line 389: this comment should be updated to unknownReady, right?

modules/common/condition/funcs_test.go
```diff
@@ -422,6 +422,45 @@ func TestMirror(t *testing.T) {
 	g.Expect(targetCondition.Message).To(BeIdenticalTo(trueReady.Message))
 }
 
+func TestMirrorReadyTrueAndFalse(t *testing.T) {
```
stuggi, line 425: I am not against adding this, but if we remove line 415 `conditions.MarkTrue(trueB.Type, trueB.Message)` it would be basically this test. then there would be `falseBError` and `ReadyCondition` is True

stuggi (commented):
looks good, as we discussed on slack, just two comments on the tests

modules/common/condition/funcs_test.go
gibizer: you are right, I need to fix this

modules/common/condition/funcs_test.go
```diff
@@ -422,6 +422,45 @@ func TestMirror(t *testing.T) {
...
+	conditions := Conditions{}
+	invalidStatusA := Condition{
+		Type:     "a",
+		Status:   "FooBar",
+		Reason:   "",
+		Severity: SeverityNone,
+		Message:  "",
+	}
+	conditions.Init(&Conditions{invalidStatusA})
+	// NOTE(gibi): we always have a ReadyCondition added automatically so it
```
bogdando, line 454: if ReadyConfition is added automatically (upon initialization?) we probably should not give it a maximum weight, when mirroring it as Ready as well?..

modules/common/condition/funcs_test.go
gibizer: True, I can do that.

modules/common/condition/funcs_test.go
gibizer: We only give it a maximum weight if it is ReadyCondition=True.

There is an assumption that ReadyCondition=True means that the overall resource is ready. We give this higher precedence than to the case when ReadyCondition=True + FooCondition=Unknown (or FooCondition=False) as we consider those as either software faults (i.e. inconsistency in ReadyCondition handling) or an edge case of operators wanting to represent an optional and currently non requested feature of the resource with a FooCondition=Unknown status.

modules/common/condition/funcs_test.go
```diff
@@ -422,6 +422,45 @@ func TestMirror(t *testing.T) {
 	g.Expect(targetCondition.Message).To(BeIdenticalTo(trueReady.Message))
 }
 
+func TestMirrorReadyTrueAndFalse(t *testing.T) {
+	g := NewWithT(t)
+
+	conditions := Conditions{}
+	// We expect that ReadyCondition True means the overall status of the
```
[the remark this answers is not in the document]
gibizer: done below and also hopefully in the doctext of Mirror too

modules/common/condition/funcs_test.go
gibizer: I agree that this error condition is not easy to trigger. If the caller never explicitly removes ReadyCondition from the conditions list then an condition with invalid status will always be lower prio than the ReadyCondition itself and therefore be ignored. So this is a highly speculative unit test to cover the case that is more clear for the reader of Mirror implementation than the user of the that implementation. Still I'd like to keep this unit test as otherwise we have a panic case in Mirror that is not covered.

PR #213, opened 2023-03-09
Add conditions for cronjobs

kajinamit:


kajinamit:
I'm submitting this based on the discussion in https://github.com/openstack-k8s-operators/keystone-operator/pull/207

modules/common/condition/conditions.go
```diff
@@ -210,4 +210,19 @@ const (
...
 	NetworkAttachmentsReadyErrorMessage = "NetworkAttachments error occured %s"
+
+	//
+	// CronjobReady condition messages
+	//
+	// CronjobReadyInitMessage
+	CronjobReadyInitMessage = "Cronjob not started"
+
+	// CronjobReadyMessage
+	CronjobReadyMessage = "Cronnjob completed"
```
stuggi, line 221: nit Cronn -> Cron

stuggi (approved):
/lgtm

modules/common/condition/conditions.go
kajinamit: Fixed. Good catch !

kajinamit (commented):
Oops. I forgot adding condition name. Will fix it by https://github.com/openstack-k8s-operators/lib-common/pull/216 .

PR #214, opened 2023-03-09
Set InterfaceRequest to get an expected ifname in the pod

stuggi:
For some services its required to identify its IP on a given network specified via an NAD name. Since it is not guaranteed which interface is used for this network the network interface name can be requested via the NetworkAttachmentAnnot annotation.

GetNetworkIFName returns the interface name base on the NAD name the interface name in Linux must not be longer then 15 chars.


olliewalsh (approved):
/lgtm

PR #215, opened 2023-03-11
Fix omitempty + default value

gibizer:
Having a CRD field with omitempty tag with a kubebuilder default value defined can cause unexpected behavior. See https://github.com/gibizer/operator-lint/blob/main/linters/crd/C003 for details.

So this patch remove the problematic definitions and bumps the operator-lint version to 0.3.0 that has a check to cover this.


stuggi (approved):
/lgtm

fmount (approved):

PR #216, opened 2023-03-13
Add conditions for cronjobs (again)

kajinamit:
The previous attempt added only status names and the condition name is still missing. This fixes that.

Also, this replaces 'Cronjob' by 'CronJob' to make it consistent with the class name.


stuggi (approved):
/lgtm

PR #218, opened 2023-03-13
Add common function to lookup Headless service pods

bshephar:
As memcached is likely to be used by multiple services, this change adds some common functions likely necessary to interact with the memcached CR. Adding these to lib-common will reduce the overhead of updating multiple service operators if the interface with memcached and it's associated CR are to change in the future.

### DIFF
--- a/modules/common/pod/pod.go
+++ b/modules/common/pod/pod.go
@@ -41,9 +41,30 @@ func GetPodListWithLabel(
 	// otherwise we hit "Error listing pods for labels: map[ ... ] - unable to get: default because of unknown namespace for the cache"
 	podList, err := h.GetKClient().CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{LabelSelector: labelSelectorString})
 	if err != nil {
-		err = fmt.Errorf("Error listing pods for labels: %v - %w", labelSelectorMap, err)
+		err = fmt.Errorf("error listing pods for labels: %v - %w", labelSelectorMap, err)
 		return nil, err
 	}
 
 	return podList, nil
+}
+
+// GetPodFQDNList gets a list of pods matching the labels provided and returns a slice of pod FQDNs.
+func GetPodFQDNList(ctx context.Context, h *helper.Helper, namespace string, labelSelector map[string]string) ([]string, error) {
+	var podSvcNames []string
+	var podList *corev1.PodList
+
+	podList, err := GetPodListWithLabel(ctx, h, namespace, labelSelector)
+	if err != nil {
+		return nil, fmt.Errorf("error getting list of pods: %w", err)
+	}
+
+	for _, pod := range podList.Items {
+		// Check for pod.Spec.Hostname and Subdomain
+		if pod.Spec.Hostname == "" || pod.Spec.Subdomain == "" {
+			return nil, fmt.Errorf("Pod does not have the required Spec Hostname and Subdomain details to accurately form a FQDN")
+		}
+		podSvcNames = append(podSvcNames, fmt.Sprintf("%s.%s", pod.Spec.Hostname, pod.Spec.Subdomain))
+	}
+
+	return podSvcNames, nil
 }


modules/common/memcached/memcached.go
```diff
@@ -0,0 +1,44 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+			Namespace: namespace,
+		},
+	}
+}
+
+// GetMemcachedServiceName function searches for all memcached services matching the provided memcachedName
+func GetMemcachedServiceName(ctx context.Context, memcachedName string, h *helper.Helper, namespace string) (svc string, err error) {
+	labelSelector := k8s_labels.Set{"app": "memcached"}
```
dciabrin, line 32: Currently the memcached controller creates a series of pod for a memcached CR, each with unique labels to distinguis between two CR, e.g.:
```oc get pod/memcached-0 -o json | jq .metadata.labels
{
  "app": "memcached",
  "controller-revision-hash": "memcached-67b5546d47",
  "cr": "memcached-memcached",
  "memcached/name": "memcached",
  "memcached/namespace": "openstack",
  "memcached/uid": "96adb626-96a4-40d1-a1b9-f3047f0137d7",
  "owner": "infra-operator",
  "statefulset.kubernetes.io/pod-name": "memcached-0"
}
```
So the labelSelector should also filter for memcached/name: memcachedName.

modules/common/memcached/memcached.go
```diff
@@ -0,0 +1,44 @@
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+			Namespace: namespace,
+		},
+	}
+}
+
+// GetMemcachedServiceName function searches for all memcached services matching the provided memcachedName
+func GetMemcachedServiceName(ctx context.Context, memcachedName string, h *helper.Helper, namespace string) (svc string, err error) {
```
dciabrin, line 31: This helper func should return ([]string, err) to account for all the pods that can be created for a single memcached CR.

modules/common/memcached/memcached.go
```diff
@@ -0,0 +1,44 @@
...
+	labelSelector := k8s_labels.Set{"app": "memcached"}
+	svcList, err := oko_svc.GetServicesListWithLabel(ctx, h, namespace, labelSelector)
+
+	if err != nil {
+		return "", err
+	}
+	for _, service := range svcList.Items {
+		if service.Name == memcachedName {
```
dciabrin, line 39: See above, this can't be used as is because of  "statefulset.kubernetes.io/pod-name": "memcached-0"

modules/common/memcached/memcached.go
```diff
@@ -0,0 +1,44 @@
...
+
+// GetMemcachedServiceName function searches for all memcached services matching the provided memcachedName
+func GetMemcachedServiceName(ctx context.Context, memcachedName string, h *helper.Helper, namespace string) (svc string, err error) {
+	labelSelector := k8s_labels.Set{"app": "memcached"}
+	svcList, err := oko_svc.GetServicesListWithLabel(ctx, h, namespace, labelSelector)
+
+	if err != nil {
+		return "", err
+	}
+	for _, service := range svcList.Items {
```
dciabrin, line 38: I wonder if rather than iterating among pods, we shouldn't create a static list out of a memcached CR.
Reason being, oslo.db is explicitely passed a list of memcached servers to control the hashing of keys that are stored across the replicas. So ultimately we expect to hash across a list of <memcached-cr>.spec.replicas, so I think we don't want to rely on whatever pod is available at the time we run this function.

dciabrin (changes requested):
Hey Brendan, this is to match what the memcached reconciler currently does and how it's used in https://github.com/openstack-k8s-operators/openstack-operator/pull/231

modules/common/memcached/memcached.go
bshephar: So, to be clear. This function is just getting the headless service. Which there will only ever be one of regardless of the number of replicas:

```
❯ oc get memcached horizon-memcached -o jsonpath='{.spec.replicas}' | jq
3

❯ oc get po -l app=memcached
NAME                  READY   STATUS              RESTARTS   AGE
horizon-memcached-0   1/1     Running             0          65s
horizon-memcached-1   1/1     Running             0          25s
horizon-memcached-2   0/1     ContainerCreating   0          19s

❯ oc get svc -l app=memcached
NAME                TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)     AGE
horizon-memcached   ClusterIP   None         <none>        11211/TCP   71s

```

So, I think the best course of action would be for me to add a new function that gets the list of pods, where then I will need to return a slice of strings like `[]String{'memcached-0', 'memcached-1', 'memcached-2'}`. But I think I should do that in a seperate function and leave this here in case we have a use for getting just the SVC like this. (Which btw is currently how I'm doing it in Horizon. I just add the single SVC and rely on whatever mechanism the k8s service uses to send traffic to the pods.

modules/common/memcached/memcached.go
dciabrin: hmm resolved from Brendan's previous comment. This was not the intent of this function.

modules/common/memcached/memcached.go
dciabrin: resolved from Brendan's previous comment. This was not the intent of this function.

modules/common/memcached/memcached.go
bshephar: That is a good point. Do we anticipate using the service at all then?

Maybe I should scrap the function entirely if it isn't as useful as I first thought. I can just work on getting the pod list instead if that's the case.

dciabrin: Based on https://github.com/openstack-k8s-operators/openstack-operator/pull/231, we just instantiate a headless service for getting DNS entries for the pods. So yes, we don't plan on accessing memcached via a service directly.

bshephar: I think we can close this one off too since I removed this function. But thanks again for the pointer on the headless service name resolution. Makes it much easier.

kajinamit:
I'm not 100% sure about this but should we put this to infra-operator rather than lib-common ?
If we add this interface to lib-common then it means we have circular dependencies between lib-common and infra-operator which sounds a bit tricky. If we can move this out to infra-operator then we expect strait dependency like `libcommon -> infra-opeeator -> foo-operator`.

modules/common/memcached/memcached.go
```diff
@@ -0,0 +1,52 @@
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
+}
+
+// GetMemcachedPodList gets a list of pods matching the labels provided and returns a slice of pod names.
+func GetMemcachedPodList(ctx context.Context, h *helper.Helper, memcachedName string, namespace string) ([]string, error) {
```
kajinamit, line 30: If we name this PodList then IMO this should return PodList instead of list of strings.
If we return these strings then we probably have to use a different naming which better represents what these strings are

modules/common/memcached/memcached.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// Define the label selector to filter on
+	labelSelector := metav1.LabelSelector{
+		MatchLabels: map[string]string{
+			"app":            "memcached",
+			"memcached/name": memcachedName,
+		},
+	}
+
+	memcachedPodList, err := h.GetKClient().CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{LabelSelector: metav1.FormatLabelSelector(&labelSelector)})
```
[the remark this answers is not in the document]
stuggi: I think we'd not need this any more as our operators are now cluster scoped. if we'd run the operator restricted mode with a multi namespace cache and you'd request pods from a not cached namespace you will run into this. https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/cache/multi_namespace_cache.go#L228

modules/common/memcached/memcached.go
bshephar: Fair point. I'll change the name

modules/common/memcached/memcached.go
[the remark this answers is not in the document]
bshephar: Ah yeah, I see the GetPodListWithLabel. So I can re-use that. I guess I could use that from the operator directly, but it might be nice to have some common interface operators can use to get the memcached pods in the right format to be used. Since it's a little different than just getting the pod.Name, we also need to append the Headless service name as well. 

I can re-use the GetPodListWithLabel, it would only really change this line:
https://github.com/openstack-k8s-operators/lib-common/pull/218/commits/1f140f4455ea282c9872984c7ea319a264b5610a#diff-ac4c9ae55fd0190f9e0ca09ce5b6d4bebd8409eed043bc85b119dc29a82459b8R42

And as for why I used the client-go library compared to controller-util. I had previously used the services.go library to get services and noticed it was also using the client-go get() function rather than controller-util. So I just kept with the established standard that I found. There may be other examples of controller-util, I didn't put much extra effort into looking once I found that though, I just made the assumption we would always use that interface.

modules/common/go.mod
```diff
@@ -40,6 +40,7 @@ require (
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
+	github.com/openstack-k8s-operators/infra-operator/apis v0.0.0-20230313171323-3a588eef4569
```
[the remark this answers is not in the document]
bshephar: Yeah, that's a fair point. I'll see where we land with the general discussion about the `GetMemcachedPodList()` method. Maybe we don't use this at all and instead just add some useful methods to the infra-operator instead?

This import was only necessary if we want to create some lib-common methods to create a memcached instance. But if we prefer to avoid the dependencies, we can move some of this logic elsewhere.

bshephar:
This has been simplified to make it more generic, removing dependencies on infra-operator, and putting the labelSelector back on the operator implementing the lookup. Since the operator using this function will have the information readily available, it probably doesn't make sense to add any additional logic to do this. 

So the main intention of this patch is to simply add a function that will lookup pods behind a headless k8s service and return the names in a slice using the format `pod-name.service-name`.  As an example, In the case of memcached with 3 replicas, then this would return:

```
[]string{'horizon-memcached-0.horizon-memcached','horizon-memcached-1.horizon-memcached',horizon-memcached-2.horizon-memcached'}
```

modules/common/pod/pod.go
```diff
@@ -47,3 +47,21 @@ func GetPodListWithLabel(
 
 	return podList, nil
 }
+
+// GetHeadlessPodSvcName gets a list of pods matching the labels provided and returns a slice of pod and service names. This
+// is required in order to allow resolution of pods behind Headless services.
```
gibizer, line 52: So does the value returned here (e.g. `horizon-memcached-0.horizon-memcached`) can be used to connect to the pod? So is this the hostname of the Pod or the FQDN of the Pod? I'm asking this as I would like to find a name for this function that better describes what it returns.

bshephar: It's the FQDN of the pod. So, you're suggesting something like: `GetHeadlessPodFQDN()` ?

gibizer: `GetHeadlessPodFQDN` sounds good to me

modules/common/pod/pod.go
```diff
@@ -47,3 +47,21 @@ func GetPodListWithLabel(
...
+
+// GetHeadlessPodFQDN gets a list of pods matching the labels provided and returns a slice of pod FQDNs. This
+// is required in order to allow resolution of pods behind Headless services.
+func GetHeadlessPodFQDN(ctx context.Context, h *helper.Helper, serviceName string, namespace string, labelSelector map[string]string) ([]string, error) {
+	var podSvcNames []string
+	var podList *corev1.PodList
+
+	podList, err := GetPodListWithLabel(ctx, h, namespace, labelSelector)
+	if err != nil {
+		return []string{}, fmt.Errorf("Error getting list of pods: %v", err)
```
gibizer, line 59: please use %w to wrap the original error into the new error. https://go.dev/blog/go1.13-errors

gibizer (commented):
One more thing then I think this is good to go

modules/common/pod/pod.go
```diff
@@ -47,3 +47,21 @@ func GetPodListWithLabel(
...
+	}
+
+	for _, pod := range podList.Items {
+		podSvcNames = append(podSvcNames, fmt.Sprintf("%s.%s", pod.Name, serviceName))
```
stuggi, line 63: I think we could make this more generic and don't have to pass in the service name, just namespace and a  pod label selector. then check if `spec.hostname` and `spec.subdomain` is set properly on the pod.

stuggi: and it would remove the limitation on heatless pods

bshephar: Ah sorry, you have asked me to do that a couple of times. My bad, latest update implements this.

modules/common/pod/pod.go
bshephar: Done

gibizer (commented):
My comment was resolved. But holding the approval for @stuggi as he had comments too.

modules/common/pod/pod.go
```diff
@@ -47,3 +47,20 @@ func GetPodListWithLabel(
...
+	var podSvcNames []string
+	var podList *corev1.PodList
+
+	podList, err := GetPodListWithLabel(ctx, h, namespace, labelSelector)
+	if err != nil {
+		return []string{}, fmt.Errorf("Error getting list of pods: %w", err)
+	}
+
+	for _, pod := range podList.Items {
+		podSvcNames = append(podSvcNames, fmt.Sprintf("%s.%s", pod.Spec.Hostname, pod.Spec.Subdomain))
```
stuggi, line 62: I think we should return an error if Hostname or Subdomain is not set

stuggi (approved):
/lgtm

dciabrin (approved):

PR #220, opened 2023-03-14
Added AnsibleEEReady condition

jlarriba:
AnsibleEE objects will be called all around the place. It is useful to have conditions to check on their status from other services.

### DIFF
--- a/modules/common/condition/conditions.go
+++ b/modules/common/condition/conditions.go
@@ -55,6 +55,9 @@ const (
 
 	// CronJobReadyCondition Status=True condition when cron jobs created ok.
 	CronJobReadyCondition Type = "CronJobReady"
+
+	// AnsibleEECondition Status=True condition when the AnsibleEE run has been created ok
+	AnsibleEECondition Type = "AnsibleEEReady"
 )
 
 // Common Reasons used by API objects.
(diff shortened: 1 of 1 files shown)


fao89 (approved):

modules/common/condition/conditions.go
```diff
@@ -55,6 +55,9 @@ const (
 
 	// CronJobReadyCondition Status=True condition when cron jobs created ok.
 	CronJobReadyCondition Type = "CronJobReady"
+
+	// AnsibleEECondition Status=True condition when the AnsibleEE run has been created ok
+	AnsibleEECondition Type = "AnsibleEEReady"
```
abays, line 60: Is this a condition that is set/checked across multiple operators?  Just asking in case it isn't, in which case it would make more sense to just have it live in the operator that uses it.

jlarriba: For the moment it is used in just one operator (ceilometer-operator), but as ansibleee objects are going to be created by a number of different operators, it seems reasonable to expect that other operators also have this need.

abays: Ack

abays (approved):
/lgtm

PR #221, opened 2023-03-16
Add make gotidy target for renovate postupgradetask

stuggi:
renovate runs go mod tidy as postupgradetask, which does not run in the directory of the sub dir. As there is no go.mod in the main repo dir, this go mod tidy fails. This adds a make gotidy target which loops through the sub module dirs and runs go mod tidy in there.

Depends-On: https://github.com/openstack-k8s-operators/openstack-k8s-operators-ci/pull/74


abays (approved):
/lgtm

PR #224, opened 2023-03-23
Add functions for OpenStack domain management

bshephar:
This change adds a CreateDomain function that leverages gophercloud to create OpenStack Domains.

### DIFF
--- a/modules/openstack/domain.go
+++ b/modules/openstack/domain.go
@@ -1,0 +1,45 @@
+package openstack
+
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


stuggi:
probably not needed now, but I think we also want funcs to do delete a domain for cleanup? or do we consider for our use case a domain gets only created?

modules/openstack/domain.go
```diff
@@ -0,0 +1,45 @@
+package openstack
+
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/domains"
```
fmount, line 7: +1 on gophercloud usage

PR #225, opened 2023-03-24
Add ComponentSelector const and enhance VerifyNetworkStatusFromAnnotation

stuggi:
Adds expectedCount to VerifyNetworkStatusFromAnnotation
    
If there are networkAttachments specificed, but the replicas count of a deployment is 0, this func will return false. This adds an expectedCount parameter and if 0, the func returns true for networkReady.

### DIFF
--- a/modules/common/const.go
+++ b/modules/common/const.go
@@ -20,6 +20,8 @@ package common
 const (
 	// AppSelector - used by operators to specify labels
 	AppSelector = "service"
+	// ComponentSelector - used by operators to specify labels for a sub component
+	ComponentSelector = "component"
 	// CustomServiceConfigFileName - file name used to add the service customizations
 	CustomServiceConfigFileName = "custom.conf"
 	// DebugCommand - Default debug command for pods
--- a/modules/common/networkattachment/networkattachment.go
+++ b/modules/common/networkattachment/networkattachment.go
@@ -97,18 +97,20 @@ func GetNetworkStatusFromAnnotation(annotations map[string]string) ([]networkv1.
 
 // VerifyNetworkStatusFromAnnotation - verifies if NetworkStatus annotation for the pods of a deployment,
 // pods identified via the service label selector, matches the passed in network attachments and the number of
-// per network IPs the ready count of the deployment. Return true if count matches with the list of IPs per network.
+// per network IPs the ready count of the deployment. Return true if count matches with the list of IPs per network,
+// no networkAttachments provided or expectedCount == 0
 func VerifyNetworkStatusFromAnnotation(
 	ctx context.Context,
 	helper *helper.Helper,
 	networkAttachments []string,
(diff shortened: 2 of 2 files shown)


stuggi:
this requires follow ups to all the consuming operators ...

modules/common/const.go
```diff
@@ -20,6 +20,8 @@ package common
 const (
 	// AppSelector - used by operators to specify labels
 	AppSelector = "service"
+	// ComponentSelector - used by operators to specify labels for a sub component
+	ComponentSelector = "component"
```
gibizer, line 24: How this relates to the VerifyNetworkStatusFromAnnotation change?

modules/common/networkattachment/networkattachment.go
```diff
@@ -97,18 +97,20 @@ func GetNetworkStatusFromAnnotation(annotations map[string]string) ([]networkv1.
...
 	networkAttachments []string,
 	serviceLabels map[string]string,
 	readyCount int32,
+	expectedCount int32,
 ) (bool, map[string][]string, error) {
 
 	networkReady := true
 	networkAttachmentStatus := map[string][]string{}
-	if len(networkAttachments) > 0 {
+	if len(networkAttachments) > 0 && expectedCount > 0 {
```
gibizer, line 113: would it be simpler not to  call VerifyNetworkStatusFromAnnotation() from the operator if Replicas is 0?

modules/common/const.go
stuggi: its not. I can split that out.

modules/common/networkattachment/networkattachment.go
stuggi: that's an alternative, can set the `networkReady` in the caller to true if Replicas is 0. doing it in here is just a bit more generic?

stuggi: we just need to get an agreement on one of the other way doing it.

gibizer: currently `expectedCount` looks like a boolean flag that turns off the whole logic of `VerifyNetworkStatusFromAnnotation`. So if I refactor this to be a boolean flag 

```golang
func VerifyNetworkStatusFromAnnotation(
	ctx context.Context,
	helper *helper.Helper,
	networkAttachments []string,
	serviceLabels map[string]string,
	readyCount int32,
	skipVerify bool,
) (bool, map[string][]string, error) {

    if skipVerify {
        return true
    }
```
then I think it shows more clearly why I think this does not belong here.

modules/common/const.go
stuggi: https://github.com/openstack-k8s-operators/lib-common/pull/226

stuggi:
ack makes sense, will close this and will just not call it in that case in the operator

PR #230, opened 2023-03-27
Simplify golang workspace target

jpodivin:
Previous implementation used external loop to control workspace inclusion. The builtin `-r` argument performs the same function.


stuggi (approved):
/lgtm

PR #231, opened 2023-03-30
Fix condition messages

gibizer:
This patch fixes the message of DBReadyErrorMessage constant as it was referring to DBsync job but it is used for DB creation conditions.

While at it I fixed typos in the messages and discrepancies in the comments as well.


stuggi (approved):
/lgtm

PR #234, opened 2023-04-04
Don't return nil for pointer reference

bshephar:
Returning a nil value for a *user.Users{} pointer results in nil pointer dereferences. This change returns empty &user.Users{} instead of nil.


bshephar:
Seems this is fine actually, the nil is the correct return value in this case.
https://go.dev/doc/effective_go#errors

PR #240, opened 2023-04-20
Add ReconcileRbac function

dprince:
This function will act as the primary means to help reconcile common rbac resources for serviceaccounts, roles, and rolebindings.

Individual operators can define functions for the rbac.Reconciler interface within their CRD reconcilers and then make a simple call to reconcile all these resources in the controller like:

	rbacRules := []rbacv1.PolicyRule{
		{
			APIGroups: []string{""},
			Resources: []string{"pods"},
			Verbs:     []string{"get", "list", "watch"},
		},
	}
	rbacResult, err := common_rbac.ReconcileRbac(ctx, h, instance, rbacRules)
	if err != nil {
		return rbacResult, err
	} else if (rbacResult != ctrl.Result{}) {
		return rbacResult, nil
	}

### DIFF
--- a/modules/common/condition/conditions.go
+++ b/modules/common/condition/conditions.go
@@ -265,15 +265,24 @@ const (
 	// ServiceAccountCreatingMessage
 	ServiceAccountCreatingMessage = "ServiceAccount creation in progress"
 
+	// ServiceAccountReadyInitMessage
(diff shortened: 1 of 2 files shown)


abays (approved):
/lgtm

PR #241, opened 2023-04-20
Allow setting protocol used for metallb LoadBalancer service

stuggi:
This adds `Protocol` to the MetalLBData type to specify which one to use for the service. Defaults to TCP if non is specified.


abays (approved):
/lgtm

PR #245, opened 2023-04-27
Abort if template parameter is missing

kajinamit:
Currently in case a template parameter is missing, render does not fail immediately and the parameter is replaced by '<no value>' string. This makes it difficult to catch some problems like typos in parameter keys.

This makes the template rendering immediately fail in case any of the parameters are missing.


stuggi (approved):
/lgtm

stuggi:
as discussed on slack, lets try adding a unit test for this case

kajinamit:
> as discussed on slack, lets try adding a unit test for this case

The latest version includes the required unit tests I believe.

kajinamit:
Hmm seems I pulled in some changes in go.mod file probably caused by some local runs (`make fmt`, `make test`, etc).
I can revert these though these are likely re-added when people runs these jobs, so I'll leave these for now.

stuggi (approved):
thanks a lot for adding the tests

PR #246, opened 2023-04-27
Add test coverage for Job

gibizer:
These unit tests uses a hand rolled mock for the external dependencies to make common.Job testable. There are possible alternatives for this:
* envtest with a running API server see in https://github.com/openstack-k8s-operators/lib-common/pull/247
* controllerutils fake client (DEPRECATED) https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/fake
* using mock libs / generators

### DIFF
--- a/modules/common/helper/helper.go
+++ b/modules/common/helper/helper.go
@@ -27,8 +27,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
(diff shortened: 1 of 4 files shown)


stuggi:
since we use envtest in placement/nova and the plan is to extend this to the other operators as well, wouldn't it be good (and I planned) to also use envtest in lib-common so that we don't have too many different ways, or do you see an issue with using envtest in lib-common and this approach would be better?

gibizer:
> > since we use envtest in placement/nova and the plan is to extend this to the other operators as well, wouldn't it be good (and I planned) to also use envtest in lib-common so that we don't have too many different ways, or do you see an issue with using envtest in lib-common and this approach would be better?
> 
> I can try to put together an similar PR with envtest so we can compare... Stay tuned. ;)

See https://github.com/openstack-k8s-operators/lib-common/pull/247 with the same coverage but with Envtest + Ginkgo

gibizer:
closed in favor of https://github.com/openstack-k8s-operators/lib-common/pull/247

PR #248, opened 2023-05-02
Util func for getting OS env vars

abays:
Currently desired for getting env var for defaults used with webhooks.

### DIFF
--- a/modules/common/util/env.go
+++ b/modules/common/util/env.go
@@ -1,0 +1,27 @@
+/*
+Copyright 2023 Red Hat
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import "os"
+
+// GetEnvVar - Get the value associated with key from environment variables, but use baseDefault as a value in case the ENV variable is not defined.
+func GetEnvVar(key string, baseDefault string) string {
+	if value, ok := os.LookupEnv(key); ok {
+		return value
+	}
+	return baseDefault
+}
--- a/modules/common/util/env_test.go
+++ b/modules/common/util/env_test.go
@@ -1,0 +1,65 @@
+/*
+Copyright 2023 Red Hat
+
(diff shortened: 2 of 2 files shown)


modules/common/util/env_test.go
```diff
@@ -0,0 +1,70 @@
...
+		},
+		{
+			name: "Get env var where it is not actually present and there is a default",
+			data: []string{"some_absent_env_var", envVarValDefault},
+			want: envVarValDefault,
+		},
+		{
+			name: "Get env var where it is not actually present and there is no default",
+			data: []string{"some_absent_env_var", ""},
+			want: "",
```
stuggi, line 57: What's the use case for GetEnvVar ? could also return an err for this case. I guess the caller needs to validate for empty string now?

modules/common/util/env.go
```diff
@@ -0,0 +1,27 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import "os"
+
+// GetEnvVar - Get the value associated with key from environment variables, but use baseDefault as a value in the case of an empty string
```
gibizer, line 21: > but use baseDefault as a value in the case of an empty string

but use baseDefault as a value in case the ENV variable is not defined.

modules/common/util/env_test.go
gibizer: I think the use case is coming from here https://github.com/openstack-k8s-operators/nova-operator/pull/335/files#diff-8731911f4dc06ba38aa36b07378ab68d4252b6a44549bb10296e1f48d9687180R164-R172 

When the containerImage defaults are set up for the webhook we want to use the ENV value if the ENV is set, otherwise we want to use a hardcoded default value for the containerImage.

abays: Yes, the use case is in the context of defaults via webhooks as @gibizer mentioned.  We could have it return an error in certain circumstances, but we'd have to think about when to do so.  Would it be when `key` is not set in ENV and `baseDefault` was provided as `""`?

Maybe this should go in a separate `webhook` package so as to avoid confusion as to its purpose?

modules/common/util/env.go
abays: Ack, will change

modules/common/util/env_test.go
gibizer: I don't think we have to handle the case specially when baseDefault is provided as "" as that means the caller explicitly provided us "".

abays: Fair enough.  I will remove that test case.

stuggi (approved):
/lgtm

abays:
@gibizer Are you good with this as well?  If so, I will merge it.

gibizer (approved):
Looks good.

PR #250, opened 2023-05-03
Add conditions check if all sub resources are true

stuggi:


modules/common/condition/funcs_test.go
```diff
@@ -316,6 +316,65 @@ func TestIsMethods(t *testing.T) {
...
+	g := NewWithT(t)
+
+	conditions.Init(nil)
+	g.Expect(conditions).To(haveSameConditionsOf(CreateList(unknownReady)))
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			conditions.Set(tt.condition)
+			conditions.AllSubConditionIsTrue()
```
abays, line 371: Is this needed here since we're actually checking on line 373?

modules/common/condition/funcs_test.go
```diff
@@ -316,6 +316,63 @@ func TestIsMethods(t *testing.T) {
...
+			condition: nil,
+			want:      true,
+		},
+		{
+			name:      "Add unknownB condition",
+			condition: unknownB,
+			want:      false,
+		},
+		{
+			name:      "Add another condition unknownA, gets sorted",
```
gibizer, line 346: I'm not sure what "gets sorted" here means.

modules/common/condition/funcs.go
```diff
@@ -159,6 +159,20 @@ func (conditions *Conditions) IsUnknown(t Type) bool {
 	return true
 }
 
+// AllSubConditionIsTrue validates if all subconditions are True
+// It assumes that all conditions report success via the True status
+func (conditions *Conditions) AllSubConditionIsTrue() bool {
```
gibizer, line 164: Thanks for pulling this out from nova-operator.

gibizer (approved):

modules/common/condition/funcs_test.go
stuggi: oh thats a c/p error. I'll remove that

stuggi: done

abays (approved):
/lgtm

modules/common/condition/funcs_test.go
gibizer: thanks!

PR #252, opened 2023-05-03
Split test module into two modules

gibizer:
To avoid circular dependencies between lib-common code, lib-common test code, and service operator api modules the existing modules/test module split into two:
1) modules/test contains helpers for handling k8s generic resource
   like ConfigMaps
2) modules/test-operators contains helpers for handling
   openstack-k8s-operators specific resources like KeystoneAPI

This way test code in lib-common that covering generic k8s resource handling does not need to depend on openstack-k8s-operators related api structures.

The TestHelper struct is also split into two and the TestHelper in modules/test-operators wraps the generic TestHelper from modules/test. The current users of modules/test needs to be switched over to use module/test-operators as a dependency but otherwise the interface TestHelper interface is kept intact.


gibizer:
Nova-operator adaptation patch https://github.com/openstack-k8s-operators/nova-operator/pull/353

gibizer:
@karelyatin FYI, this lib-common change will need adaptation in neutron- and ovn-operator. But only the import of lib-common/modules/test needs to be changed to lib-common/modules/test-operators. No other issue is expected based on my trial of this in nova-operator.

stuggi (approved):
thanks, looks good to me!

PR #253, opened 2023-05-04
Update pre-commit and golangci-lint setting

gibizer:
Added golangci-lint execution in pre-commit and updated enabled linters. Fixed the ginkgolinter findings.

Also added a github action to run pre-commit. Eventually we want to run prow CI on lib-common but that needs more prep work.


stuggi (approved):
/lgtm

PR #254, opened 2023-05-04
[Job]Support re-running the job while previous job still exists

gibizer:
The job handling logic is changed to allow re-running the job if the
hash of the job changed while the previous job is still exists.

If the old job still not finished when DoJob is called with a new job
then the caller is requested to wait by returning requeue result
until the old job is finished (either succeeded or failed).

If the old job already finished when DoJob is called with a new job
then then the old job is deleted and the caller is requested to requeue.
So that when DoJob is called after this requeue the normal code path can
be taken, and a new job can be created as no job exists.

To be able to detect the if the hash of the job is changed while the old
job is still running the hash of the job is stored as an annotation on
the k8s job object. This way DoJob can compare the hash of the running
job and the hash of the expected job.

The hash of the job is modified to only consider the Job.Spec.Template
part. As fields outside of the Template (e.g. TTL) are not defining what
to be run, but just how to run it. This way TTL can he changed on the
job any time without affecting its hash.

Depends-on: https://github.com/openstack-k8s-operators/lib-common/pull/247

### DIFF
--- a/modules/common/job/job.go
+++ b/modules/common/job/job.go
@@ -43,14 +43,17 @@ func NewJob(
 	beforeHash string,
 ) *Job {
 
-	return &Job{
-		job:        job,
-		jobType:    jobType,
-		preserve:   preserve,
-		timeout:    timeout,
-		beforeHash: beforeHash,
-		changed:    false,
+	j := &Job{
+		expectedJob: job,
+		actualJob:   nil,
+		jobType:     jobType,
+		preserve:    preserve,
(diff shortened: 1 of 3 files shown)


stuggi (approved):
/lgtm Thanks for you work on the job PRs! As discussed on slack, one scenario e.g. for edpm we could add in a follow up is to allow cancel a running job if the definition (hash) changed. could add method to set allowcancel or something, which then could be used in DoJob

PR #255, opened 2023-05-04
[test-operators]Simulate internal endpoint for Keystone

gibizer:
The nova-operator only depends on the public endpoint of Keystone so the original helper copied from nova-operator only simulates that one. However other operators like neutron-operator depends on the internal endpoint too. The keystone-operator by default creates both so the test helper also modified to simulate both.

```
$ oc get KeystoneAPI/keystone -o yaml | yq ".status.apiEndpoint"
{
  "internal": "http://keystone-internal.openstack.svc:5000",
  "public": "http://keystone-public-openstack.apps-crc.testing"
}
```


stuggi (approved):
/lgtm

PR #256, opened 2023-05-05
[pre-commit]skip branch check in CI

gibizer:
We run all pre-commit on both push to branch and open pull request events. This means that we run it also against master after a PR is approved. In that case the no-commit-to-branch check, that enforces no direct commit to master branch fails, so we disable this check in CI. Locally it still make sense to keep it as it is good git hygiene to always work on a local topic branch.


gibizer:
See failing example in https://github.com/openstack-k8s-operators/lib-common/actions

stuggi (approved):
/lgtm

PR #257, opened 2023-05-05
[job]Switch to controller-runtime client

gibizer:
This patch removes the last usage of the direct k8s client from the job package. Now the controller-runtime client has all the capabilities we need.

Also simplified DeleteJob logic. There is no need to get the job first to be able the delete it. We just need to ignore if the Delete returns NotFound.


modules/common/test/functional/job_test.go
```diff
@@ -389,3 +389,39 @@ var _ = Describe("job.Job", func() {
...
+		// as namespaces cannot be deleted in a locally running envtest. See
+		// https://book.kubebuilder.io/reference/envtest.html#namespace-usage-limitation
+		namespace = uuid.New().String()
+		th.CreateNamespace(namespace)
+		// We still request the delete of the Namespace to properly cleanup if
+		// we run the test in an existing cluster.
+		DeferCleanup(th.DeleteNamespace, namespace)
+	})
+
+	It("DeleteJob deletes existing jobs", func() {
```
stuggi, line 407: just curious if there is a reason why you haven't added these tests to the above Describe?

gibizer: Technically I can add it without any change. I just looked at the existing Describe and it was talking about the job.Job struct. While DeleteJob is does not have a job.Job receiver. But I guess it is less noise if I merge the two describe and just rename the one Describe to talk about the whole package.

gibizer: Will fix it in a minute.

gibizer: Done.

stuggi (approved):
/lgtm

PR #258, opened 2023-05-05
[DNS] Set annotation on MetalLB services to register in dnsmasq

stuggi:


abays (approved):
/lgtm

modules/common/endpoint/endpoint.go
```diff
@@ -131,6 +133,8 @@ func ExposeEndpoints(
 				exportLabels,
 				timeout,
 			)
+			// add annotation to register service name in dnsmasq
+			svc.AddAnnotation(map[string]string{AnnotationHostnameKey: svc.GetServiceHostname()})
```
gibizer, line 137: Is there a technical reason why we add this annotation after NewService call while the rest of the annotations collected before the NewService call? If no specific reason then I would move this closer to the rest of the annotation gathering. Also the new AddAnnotation function could be used for the other annotations above too.

stuggi: its because the NewService initializes the serviceHostname. We rely on that information here as the value of the annotation.  On the 2nd point you are right, we first initialize the svc and then use AddAnnotation to add the  MetalLB annotations.

gibizer (approved):
looks good to me.

PR #263, opened 2023-05-10
Add volume module

konan-abhi:
Adds volume module with functions to manage
- ServiceCheck

E.g. Check whether particular cinder service is running or not.

### DIFF
--- a/modules/openstack/volume.go
+++ b/modules/openstack/volume.go
@@ -1,0 +1,51 @@
+/*
+Copyright 2023 Red Hat
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package openstack
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/go-logr/logr"
+	services "github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/services"
+)
+
+// VolumeServiceCheck - Check particular cinder service is enabled and running or not
+func (o *OpenStack) VolumeServiceCheck(
+	log logr.Logger,
+	serviceName string,
+) (bool, error) {
+	var serviceRunning bool
+	serviceRunning = false
+	log.Info(fmt.Sprintf("Checking %s service is running or not", serviceName))
+	allPages, err := services.List(o.osclient, services.ListOpts{}).AllPages()
+	if err != nil {
+		return serviceRunning, err
+	}
+	allServices, err := services.ExtractServices(allPages)
+	if err != nil {
+		return serviceRunning, err
+	}
+	for _, service := range allServices {
+		if strings.Contains(service.Binary, serviceName) && service.State == "up" && service.Status == "enabled" {
+			serviceRunning = true
+			break
+		}
+	}
+	return serviceRunning, nil
+
+}


stuggi:
@konan-abhi shouldn't this be generic as part of `service.go`, then could use GetService() from https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/openstack/service.go#L79  and add a func to check an ExpectedState of the service we pass to that func?

konan-abhi:
> @konan-abhi shouldn't this be generic as part of `service.go`, then could use GetService() from https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/openstack/service.go#L79 and add a func to check an ExpectedState of the service we pass to that func?

Hi @stuggi AFIAK the mentioned service module will return the service and we can check whether it is enabled or not but it does not tell us whether it is up or not. This is why I have added this piece of code here.

The goal is from glance-operator we want to check whether cinder volume service is enabled and running or not.

stuggi:
> > @konan-abhi shouldn't this be generic as part of `service.go`, then could use GetService() from https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/openstack/service.go#L79 and add a func to check an ExpectedState of the service we pass to that func?
> 
> Hi @stuggi AFIAK the mentioned service module will return the service and we can check whether it is enabled or not but it does not tell us whether it is up or not. This is why I have added this piece of code here.
> 
> The goal is from glance-operator we want to check whether cinder volume service is enabled and running or not.

ah right, have not closely looked at the imported gophercloud services module, sorry for that :)

konan-abhi:
> > > @konan-abhi shouldn't this be generic as part of `service.go`, then could use GetService() from https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/openstack/service.go#L79 and add a func to check an ExpectedState of the service we pass to that func?
> > 
> > 
> > Hi @stuggi AFIAK the mentioned service module will return the service and we can check whether it is enabled or not but it does not tell us whether it is up or not. This is why I have added this piece of code here.
> > The goal is from glance-operator we want to check whether cinder volume service is enabled and running or not.
> 
> ah right, have not closely looked at the imported gophercloud services module, sorry for that :)

No problem :D

konan-abhi:
PR is ready for review!

fmount:
So @stuggi the idea behind this change is to query (using gophercloud) an existing `ctlplane` where cinder is *already* deployed: this way we can reconfigure glance to point to cinder and enable the backend (we have multistore, we can define as many backends as we want and eventually switch the default rolling out new pods).
However, instead of blindly enabling cinder, the glance-operator can have a quick check returning the status of cinder-volume, and keep this status to evaluate and raise potential issues when the cinder is used.
I think it makes sense having a call like this instead of watching/querying the cinder CR, because here we're getting a ospclient and query the service.

modules/openstack/volume.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/go-logr/logr"
+	services "github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/services"
+)
+
+// ServiceCheck - Check particular cinder service is enabled and running or not
+func (o *OpenStack) ServiceCheck(
```
fmount, line 28: @stuggi @konan-abhi wondering if this function can be considered generic enough to be moved in `service.go` : we'
re creating `volume.go`, but the function accepts a `serviceName string` as an argument, so it would make sense to me having it as part of service.go.

konan-abhi: IMO services module is related to keystone which is used for service related CRUD operations, we are using cinder extension here, so I think having this separate module makes more sense. Lets see what @stuggi has to say here.

fmount: oh I see your point, "service" is an extension of `blockstorage`, therefore it justifies having a `volume.go` here.
I'm ok w/ this implementation then, and we can add more related to "volumes" as long as we need it.

modules/openstack/volume.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/go-logr/logr"
+	services "github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/services"
+)
+
+// VolumeServiceCheck - Check particular cinder service is enabled and running or not
+func (o *OpenStack) VolumeServiceCheck(
```
fmount, line 28: +1

fmount (approved):

stuggi (approved):
/lgtm

PR #271, opened 2023-05-23
[test]Use Eventually for all k8s Update calls

gibizer:
We have couple of envtest helpers which call Update on existing k8s resources, for example to simulate a status condition. These updates can race with updates from the controller under test. In such case k8s will return 409. So this can cause random test instability. To avoid this this PR moves all Update calls into an Eventually block so the test will re-query the resource, reapply the change and redo the Update in a limited retry loop.

The only exception is CreateKeystoneAPI. That call creates an KeystoneAPI resource and updates its Status in a second Update call as the Status cannot be set during create as it needs a different client. In this case we don't expect anybody else to touch the KeystoneAPI resource.


stuggi (approved):
/lgtm

PR #272, opened 2023-05-23
delete archive module

stuggi:


stuggi:
I think its time that we can drop the archive

gibizer (approved):
Looks OK. We can always tag the last commit that has the archive available if somebody needs it.

abays (approved):
/lgtm

PR #273, opened 2023-05-25
Bump controller-runtime

hjensas:
In controller-runtime 0.15.x and later webhook.Validator expect both error and warnings returned. This change bumps controller- runtime


bshephar:
Looks like we will need to update the method signature of `ValidateCreate()` in all of these web hooks at the same time as this version bump:

https://github.com/openstack-k8s-operators/lib-common/actions/runs/5079078182/jobs/9124373908?pr=273#step:4:121
```
	*Galera does not implement admission.Validator (wrong type for ValidateCreate method)
		have ValidateCreate() error
		want ValidateCreate() (warnings admission.Warnings, err error)
```

bshephar:
Ah, I see you already did it:
https://github.com/openstack-k8s-operators/mariadb-operator/pull/110/files

PR #274, opened 2023-05-26
Simulate Deployment and StatefulSet Ready with Pods

fmount:
In the effort of adding `envTest` across the storage operators, we've realized that many operators
should take advantage of a lot of functions if defined in `lib-common`.
This patch moves the common functions (used in also in [1]) to the helpers module.

[1] https://github.com/openstack-k8s-operators/manila-operator/pull/67/


gibizer (approved):
Looks good

PR #277, opened 2023-05-30
[test]Handle PVC in StatefulSet and Deployment

gibizer:
The helper simulating Pods for StatefulSets and Deployments are changed to support PVCs. Before this patch if PVC was present the Pod creation failed with missing volumes. Now the helper simply drops the volumes and mounts from the Pod it creates. Later on if we need a better solution we can translate PVCs from to volumes when creating the Pod, but I believe we will not need this in our tests.


stuggi (approved):
/lgtm

PR #278, opened 2023-06-01
Use consistent message for all conditions

kajinamit:
This updates condition messages so we use the same format for all conditions.

### DIFF
--- a/modules/common/condition/conditions.go
+++ b/modules/common/condition/conditions.go
@@ -111,7 +111,7 @@ const (
 	ReadyInitMessage = "Setup started"
 
 	// ReadyMessage
-	ReadyMessage = "Setup complete"
+	ReadyMessage = "Setup completed"
 
 	//
 	// InputReady condition messages
@@ -120,7 +120,7 @@ const (
 	InputReadyInitMessage = "Input data not checked"
 
 	// InputReadyMessage
-	InputReadyMessage = "Input data complete"
+	InputReadyMessage = "Input data checked"
 
 	// InputReadyWaiting
 	InputReadyWaitingMessage = "Input data resources missing"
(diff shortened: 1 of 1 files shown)


modules/common/condition/conditions.go
```diff
@@ -219,10 +219,10 @@ const (
 	// NetworkAttachmentsReady condition messages
 	//
 	// NetworkAttachmentsReadyInitMessage
-	NetworkAttachmentsReadyInitMessage = "NetworkAttachments not started"
+	NetworkAttachmentsReadyInitMessage = "NetworkAttachments checked"
```
stuggi, line 222: `NetworkAttachments not checked` ?

modules/common/condition/conditions.go
```diff
@@ -219,10 +219,10 @@ const (
...
 
 	// NetworkAttachmentsReadyMessage
-	NetworkAttachmentsReadyMessage = "NetworkAttachments completed"
+	NetworkAttachmentsReadyMessage = "NetworkAttachments not checked"
```
stuggi, line 225: `NetworkAttachments checked` ?

stuggi (commented):
this will most likely required updates to kuttl assert s ?

modules/common/condition/conditions.go
kajinamit: Fixed. Good catch !

modules/common/condition/conditions.go
kajinamit: Fixed in the latest version.

kajinamit:
> this will most likely required updates to kuttl assert s ?

A few operators ( placement-operator ) assert condition messages so we have update these, but I generally prefer not asserting these messages in kuttl, because strictly asserting the messages are not very much useful while it requires more frequent update caused by underlying change.

stuggi (approved):
/lgtm

stuggi:
> > this will most likely required updates to kuttl assert s ?
> 
> A few operators ( placement-operator ) assert condition messages so we have update these, but I generally prefer not asserting these messages in kuttl, because strictly asserting the messages are not very much useful while it requires more frequent update caused by underlying change.

I think it makes sense what you said. do you plan to follow up on this?

kajinamit:
> I think it makes sense what you said. do you plan to follow up on this?

Quickly searching the repositories it seems only mariadb-operator and placement-operator assert every single condition (including messages) so I've submitted PR to update these.

https://github.com/openstack-k8s-operators/placement-operator/pull/18
https://github.com/openstack-k8s-operators/mariadb-operator/pull/118

kajinamit:
We later learned we have to reconsider the implementation to assert conditions (see https://github.com/openstack-k8s-operators/placement-operator/pull/18 ) so this may later require changes in kuttl tests.

PR #279, opened 2023-06-01
Remove unused CronJobReadyRunningMessage

kajinamit:
This condition is not really used, because we only check whether the resource is created.


stuggi (approved):
/lgtm

PR #280, opened 2023-06-05
Template: Allow customizing missingkey

kajinamit:
... to allow developers to use a different option(such as zero) if needed.

### DIFF
--- a/modules/common/util/template_util.go
+++ b/modules/common/util/template_util.go
@@ -57,6 +57,7 @@ type Template struct {
 	ConfigOptions      map[string]interface{} // map of parameters as input data to render the templates
 	SkipSetOwner       bool                   // skip setting ownership on the associated configmap
 	Version            string                 // optional version string to separate templates inside the InstanceType/Type directory. E.g. placementapi/config/18.0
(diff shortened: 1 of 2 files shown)


modules/common/util/template_util.go
```diff
@@ -57,6 +57,7 @@ type Template struct {
 	ConfigOptions      map[string]interface{} // map of parameters as input data to render the templates
 	SkipSetOwner       bool                   // skip setting ownership on the associated configmap
 	Version            string                 // optional version string to separate templates inside the InstanceType/Type directory. E.g. placementapi/config/18.0
+	MissingKey         string                 // Response to indexing a map with a key that is not present.
```
rabi, line 60: should we constrain the value like template module using an enum?

kajinamit: I checked the codes in the template module but we can't use the constants defined there. We have to provide these options as a string format and it parses the input to that constant internally. So I decided to avoid implementing type check here in this layer.

https://go.dev/src/text/template/option.go

rabi: Maybe something like this[1]? We just need two things with default 'error'.

[1] https://go.dev/play/p/z1VvfIOPrW5

PR #281, opened 2023-06-05
Use Endpoint type for endpoint map key

kajinamit:
The ExposeEndpoints function returns map of endpoints but we can't use endpoint type constants, because golang differentiate Endpoint type and string type and causes type mismatch failure.

This fixes the key type so that we can use the constants instead of hard-coding the same strings in each operator.


kajinamit:
This would demonstrate the current problem: https://github.com/openstack-k8s-operators/horizon-operator/pull/137

bogdando (approved):

stuggi:
ExposeEndpoints is deprecated as we moved to route management via the openstack-operator. Closing this PR

PR #282, opened 2023-06-05
Service: Add interface to obtain IPFamilies

kajinamit:
Some config options require special format according to the IP version used by a service. For example memcache server list used by python-memcached requires inet6 prefix in case memcached listens at an IPv6 address.
    
This change introduces the interface to obtain spec.IPFamilies of Service resource so that we can determine which IP version is used by a service.


gibizer (approved):
Looks good to me

PR #283, opened 2023-06-08
Accept multiple labels for affinity

kajinamit:
Some operators deploy multiple components and use the additional selector label (ComponentSelector) to distinguish components.


stuggi:
do we need to pass multiple labels, or wouldn't it be enough that the caller uses a component selector instead of a global service selector, like `glance-api-external` or `glance-api-internal`?

kajinamit:
> 
Some operators are already using ComponentSelector but these put only component name (eg. api) instead of the full process name(eg. heat-api).
We can adjust that but I tend to follow that pattern because having full service name along with ServiceSelector sounds redundant.

bogdando:
lgtm

PR #284, opened 2023-06-10
Get gophercloud system wide client

fmount:
The `NewOpenStack` function is only used by `keystone` to get a an admin, system wide token that can be used by other operators through keystone service to register the associated resources. However, without `scope:system` `Glance` is not able to register the global limits defined in its main CR and it will fail configuring limits [1].
This patch ensures we have a system wide admin client that is able to act globally against the existing ctlplane.

[1] https://docs.openstack.org/glance/latest/admin/quotas.html#configuring-glance-for-per-tenant-quotas

### DIFF
--- a/modules/openstack/openstack.go
+++ b/modules/openstack/openstack.go
@@ -40,6 +40,7 @@ type AuthOpts struct {
 	TenantName string
 	DomainName string
 	Region     string
+	Scope      *gophercloud.AuthScope
 }
 
 // NewOpenStack creates a new new instance of the openstack struct from a config struct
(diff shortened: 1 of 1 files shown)


modules/openstack/openstack.go
```diff
@@ -47,12 +47,16 @@ func NewOpenStack(
...
 ) (*OpenStack, error) {
+
 	opts := gophercloud.AuthOptions{
 		IdentityEndpoint: cfg.AuthURL,
 		Username:         cfg.Username,
 		Password:         cfg.Password,
 		TenantName:       cfg.TenantName,
 		DomainName:       cfg.DomainName,
+		Scope: &gophercloud.AuthScope{
+			System: true,
```
konan-abhi, line 58: Shouldn't this be based on some condition or do we always want this client with system scope?

konan-abhi: Something like this;
if os.Getenv("OS_SYSTEM_SCOPE") != "all"
     System: false

stuggi: As you said, I am not sure if we need it, but you could add a Scope pointer to the above `AuthOpts` and if it is not nil you can set it in `NewOpenStack`. with this there would be no need to the keystone-operator

konan-abhi (approved):
Looks good, Thank you!

stuggi (approved):
/lgtm

PR #293, opened 2023-07-04
Allow customize route via route.Override

stuggi:
Adds an OverridSpec to the Route which allows to customize metadata.Annotations, metadata.Labels and spec of a route. The override values get merged into the object definition created by the operator. This allows e.g. to add custom labels, configure the route via annotations as in [1], or set TLS parameters.

[1] https://docs.openshift.com/container-platform/4.13/networking/routes/route-configuration.html#nw-route-specific-annotations_route-configuration

Jira: [OSP-21715](https://issues.redhat.com//browse/OSP-21715)
Jira: [OSP-26299](https://issues.redhat.com//browse/OSP-26299)


stuggi:
need to add more tests

stuggi:
example on service operator usage in https://github.com/openstack-k8s-operators/keystone-operator/pull/269

bshephar (commented):
I like the approach, it makes the interface within lib-common much more flexible. It does put some of the complexity burden back onto the user as they need to understand how to set these options in their yaml files. But, ultimately the added extensibility and maintainability of this is nice.

I'm not sure if we need to literally copy the entire RouteSpec though. Could we maybe import and just set `Spec` to `RouteSpec`?

modules/common/route/types.go
```diff
@@ -38,3 +41,112 @@ type GenericRouteDetails struct {
...
+	// path that the router watches for, to route traffic for to the service. Optional
+	//
+	// +optional
+	// +kubebuilder:validation:Pattern=`^/`
+	Path string `json:"path,omitempty" protobuf:"bytes,2,opt,name=path"`
+
+	// to is an object the route should use as the primary backend. Only the Service kind
+	// is allowed, and it will be defaulted to Service. If the weight field (0-256 default 100)
+	// is set to zero, no traffic will be sent to this backend.
+	// +optional
```
stuggi, line 128: This is the only difference to the org RouteSpec to make To optional. otherwise you can not override e.g. only TLS related things

modules/common/route/types.go
```diff
@@ -38,3 +41,112 @@ type GenericRouteDetails struct {
...
+// and return a standard 503 response.
+//
+// The `tls` field is optional and allows specific certificates or behavior for the
+// route. Routers typically configure a default certificate on a wildcard domain to
+// terminate routes without explicit certificates, but custom hostnames usually must
+// choose passthrough (send traffic directly to the backend via the TLS Server-Name-
+// Indication field) or provide a certificate.
+//
+// Copy of RouteSpec in https://github.com/openshift/api/blob/master/route/v1/types.go, all
+// parameters set to optional.
```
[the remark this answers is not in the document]
stuggi: see bellow the only change is to make `To` optional

modules/common/route/types.go
```diff
@@ -38,3 +41,112 @@ type GenericRouteDetails struct {
...
+type OverrideSpec struct {
+	// +optional
+	*EmbeddedLabelsAnnotations `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+	// Spec defines the behavior of a Route.
+	// https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
+	// +optional
+	Spec *Spec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
+}
+
+// EmbeddedLabelsAnnotations is an embedded subset of the fields included in k8s.io/apimachinery/pkg/apis/meta/v1.ObjectMeta.
```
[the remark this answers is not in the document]
stuggi: sure, could rename it

stuggi:
> I like the approach, it makes the interface within lib-common much more flexible. It does however put some of the complexity burden back on to the user. As they need to understand how to set these options in their yaml files. But, ultimately the added extensibility and maintainability of this is nice.
> 
> I'm not sure if we need to literally copy the entire RouteSpec though. Could we maybe import and just set `Spec` to `RouteSpec`?

to be able to override a subset of the spec, we have to make sure that all parameters in the override spec are optional and not required as in the original spec definition. thats why we can not just use it.

bshephar:
> > I like the approach, it makes the interface within lib-common much more flexible. It does however put some of the complexity burden back on to the user. As they need to understand how to set these options in their yaml files. But, ultimately the added extensibility and maintainability of this is nice.
> > I'm not sure if we need to literally copy the entire RouteSpec though. Could we maybe import and just set `Spec` to `RouteSpec`?
> 
> to be able to override a subset of the spec, we have to make sure that all parameters in the override spec are optional and not required as in the original spec definition. thats why we can not just use it.

Right yeah, sorry I see Gibi mentioned the same thing and you responded to his comment. Thanks for the clarification.

modules/common/route/route.go
```diff
@@ -88,8 +92,43 @@ func (r *Route) CreateOrPatch(
...
+			if err != nil {
+				return ctrl.Result{}, fmt.Errorf("error patching Route Spec: %w", err)
+			}
+
+			patchedSpec := routev1.RouteSpec{}
+			err = json.Unmarshal(patchedJSON, &patchedSpec)
+			if err != nil {
+				return ctrl.Result{}, fmt.Errorf("error unmarshalling patched Route Spec: %w", err)
+			}
+			r.route.Spec = patchedSpec
```
gibizer, line 126: OK so this defines what the override does. Still we need to define for the end user what that means. See my comment in the keystone patch with example questions.

modules/common/route/route.go
```diff
@@ -88,8 +92,43 @@ func (r *Route) CreateOrPatch(
...
+		}
+	}
+
 	op, err := controllerutil.CreateOrPatch(ctx, h.GetClient(), route, func() error {
-		route.Labels = util.MergeStringMaps(route.Labels, r.route.Labels)
```
gibizer, line 92: Does the merging happen during strategicpatch.StrategicMergePatch now?

modules/common/route/types.go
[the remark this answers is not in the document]
gibizer: Thanks!

modules/common/go.mod
```diff
@@ -79,3 +79,7 @@ require (
 	github.com/kr/pretty v0.3.1 // indirect
 	github.com/rogpeppe/go-internal v1.10.0 // indirect
 )
+
+// mschuppert: map to latest commit from release-4.13 tag
+// must consistent with common module
+replace github.com/openshift/api => github.com/openshift/api v0.0.0-20230414143018-3367bc7e6ac7
```
gibizer, line 85: Do we have significant difference between the old and the new version. Do we need to be careful bumping to this in the service operators? At least the keystone-operator bump seems to cause no issues so maybe this is not dangerous.

modules/common/route/route.go
```diff
@@ -37,10 +39,12 @@ func NewRoute(
 	route *routev1.Route,
 	labels map[string]string,
 	timeout time.Duration,
+	override *OverrideSpec,
```
gibizer, line 42: NewRoute is not used outside of lib-common so this parameter addition will not break the service operators.

modules/common/route/route.go
gibizer: OK the label merging logic now part of the overall merge above.

gibizer (commented):
Looks good to me. Extra thanks for the documentation and the test coverage.

modules/common/route/types.go
```diff
@@ -38,3 +41,146 @@ type GenericRouteDetails struct {
...
+	//
+	// +kubebuilder:validation:Enum=None;Subdomain;""
+	WildcardPolicy routev1.WildcardPolicyType `json:"wildcardPolicy,omitempty" protobuf:"bytes,7,opt,name=wildcardPolicy"`
+}
+
+// TargetReference specifies the target that resolve into endpoints. Only the 'Service'
+// kind is allowed. Use 'weight' field to emphasize one over others.
+// Copy of RouteTargetReference in https://github.com/openshift/api/blob/master/route/v1/types.go,
+// parameters set to be optional, have omitempty, and no default.
+type TargetReference struct {
```
stuggi, line 167: I also had to create a copy of `RouteTargetReference` setting omitempty and removing defaults. Otherwise the defaulting would have overwritten values set by the operator

modules/common/go.mod
stuggi: there should be no issues, but we should update this. right now we use a really old OCP api version and there are no later releases, so we have to pin to a release tag. for the future we probably want to have versions/branches of our operators for each OCP release and then bump this to match the OCP release.

modules/common/route/route.go
stuggi: right. so far all operators just use the ExposeEndpoints, so NewRoute is kind of internal right now. but might change

modules/common/go.mod
stuggi: and we have https://issues.redhat.com/browse/OSP-26293 to bump it anyway

abays (approved):
/lgtm

modules/common/go.mod
gibizer: OK with me.

stuggi:
rebased

stuggi:
@gibizer @abays @bshephar are we good to merge this now that lib-common is tagged for 0.1.0?

gibizer:
> @gibizer @abays @bshephar are we good to merge this now that lib-common is tagged for 0.1.0?

I'm OK to land this now. My comments were answered and we tagged lib-common. This PR does not create a backward incompatible changes for our service operators so if we need to fix something else in lib-common for the dev preview we can do that on main still after this PR.

abays (approved):
/lgtm

PR #294, opened 2023-07-05
Add helpers to assert a resource does not exist

kajinamit:
There are some cases where we want to ensure a resource is not yet created by the operator (eg. No resources should be created in case the global secret does not exist). This adds the helper functions to assert that.

This also removes the unused AssertConfigMapExists. This function can be replaced by GetConfigMap.

### DIFF
--- a/modules/common/test/helpers/configmap.go
+++ b/modules/common/test/helpers/configmap.go
@@ -19,7 +19,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
-
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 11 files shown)


stuggi:
this needs rebase after we fixed the circular dependency with https://github.com/openstack-k8s-operators/lib-common/pull/301

modules/common/test/helpers/configmap.go
```diff
@@ -96,11 +95,11 @@ func (tc *TestHelper) CreateConfigMap(name types.NamespacedName, data map[string
 	return tc.CreateUnstructured(raw)
 }
 
-// AssertConfigMapExists used only in lib-common tests
-func (tc *TestHelper) AssertConfigMapExists(name types.NamespacedName) *corev1.ConfigMap {
+// AssertConfigMapDoesNotExist ensures the ConfigMap resource does not exist in a k8s cluster.
```
stuggi, line 98: is it expected that you changed the current AssertConfigMapExists and not created a copy to have both, AssertConfigMapExists and AssertConfigMapDoesNotExist? AssertConfigMapExists might be used in some checkes?

kajinamit: This is intentional (see the commit message). We can use GetConfigMap instead and as far as I checked this is not yet used. Alternatively we can add `Assert<resource>MapExists` to all for consistency but I feel like that is redundant.

stuggi: seems it is only used in the infra-operator https://github.com/openstack-k8s-operators/infra-operator/blob/main/tests/functional/dnsmasq_controller_test.go#L226 . need to updated it there.

gibizer: I favor simplicity so lets use GetConfigMap if we want to assert that the CM exists.

gibizer (approved):
Looks good.

kajinamit:
I've pushed https://github.com/openstack-k8s-operators/infra-operator/pull/86 to drop usage of AssertConfigMapExists from infra-operator(which can be merge independently)

stuggi (approved):
/lgtm

PR #295, opened 2023-07-05
Add helper functions to assert CronJob resources

kajinamit:
This introduces a few initial helper functions to assert CronJob resources in functional tests. Now the resource are all created by operators, this does not implement the delete interface.

### DIFF
--- a/modules/test/helpers/cronjob.go
+++ b/modules/test/helpers/cronjob.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2022 Red Hat
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


kajinamit:
Note: This PR includes the function to assert the CronJob does not exist, which is widely added by https://github.com/openstack-k8s-operators/lib-common/pull/294

modules/test/helpers/cronjob.go
```diff
@@ -0,0 +1,45 @@
+/*
+Copyright 2022 Red Hat
```
abays, line 2: nit: `s/2022/2023`

stuggi (changes requested):
this should now go into modules/common/test/helpers after we landed https://github.com/openstack-k8s-operators/lib-common/pull/301

PR #296, opened 2023-07-05
Add new Degraded condition status

kajinamit:
This adds the new "Degraded" condition status to the existing "True", "False" and "Unkown". This status is similar to "True" and represents that the resources are ready to serve requests, but degraded. This is useful to expose status of deployments which might contain pods less than requested repolicas.

### DIFF
--- a/modules/common/condition/conditions.go
+++ b/modules/common/condition/conditions.go
@@ -18,6 +18,16 @@ limitations under the License.
 
 package condition
 
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Common Condition Statuses used by API objects
+const (
+	// ConditionDegraded defines the status where a resource is ready but degraded (eg deployment have at least one pod but available pods are less than requested numbers)
+	ConditionDegraded corev1.ConditionStatus = "Degraded"
+)
+
 // Common Condition Types used by API objects.
 const (
 	// ReadyCondition defines the Ready condition type that summarizes the operational state of an API object.
@@ -109,6 +119,9 @@ const (
 	//
 	// ReadyInitMessage
 	ReadyInitMessage = "Setup started"
+
+	// ReadyDegraded
+	ReadyDegraded = "Setup degraded"
 
 	// ReadyMessage
 	ReadyMessage = "Setup complete"
--- a/modules/common/condition/funcs.go
+++ b/modules/common/condition/funcs.go
@@ -132,6 +132,11 @@ func (conditions *Conditions) MarkUnknown(t Type, reason Reason, messageFormat s
 	conditions.Set(UnknownCondition(t, reason, messageFormat, messageArgs...))
 }
 
+// MarkDegraded sets Status=Degraded for the condition with the given type.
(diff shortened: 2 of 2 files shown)


kajinamit:
Example usage can be found it https://github.com/kajinamit/keystone-operator/commit/b8944a41214d412997e92189e65d616c14ce3d47 .

modules/common/condition/conditions.go
```diff
@@ -18,6 +18,16 @@ limitations under the License.
...
 package condition
 
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Common Condition Statuses used by API objects
+const (
+	// ConditionDegraded defines the status where a resource is ready but degraded (eg deployment have at least one pod but available pods are less than requested numbers)
+	ConditionDegraded corev1.ConditionStatus = "Degraded"
```
gibizer, line 28: I don't think his will result in a valid Condition according to the type definition in https://github.com/kubernetes/apimachinery/blob/77401902abdff140aad8d3369ea748f68aaa5810/pkg/apis/meta/v1/types.go#L1459-L1463
```golang
type Condition struct {
	// type of condition in CamelCase or in foo.example.com/CamelCase.
	// ---
	// Many .condition.type values are consistent across resources like Available, but because arbitrary conditions can be
	// useful (see .node.status.conditions), the ability to deconflict is important.
	// The regex it matches is (dns1123SubdomainFmt/)?(qualifiedNameFmt)
	// +required
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:Pattern=`^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])$`
	// +kubebuilder:validation:MaxLength=316
	Type string `json:"type" protobuf:"bytes,1,opt,name=type"`
	// status of the condition, one of True, False, Unknown.
	// +required
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:Enum=True;False;Unknown
	Status ConditionStatus `json:"status" protobuf:"bytes,2,opt,name=status"`
...
```
It seems that Status should be an enum of True/False/Unknown

kajinamit: I believe we are not using it but ouw own definition implemented here https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/condition/types.go#L50 so that type validation does not exist for us.

gibizer: Ahh, then technically it is possible to define a different struct with a different enum. Still I might advise against breaking the convention of that enum based on:

>  Conditions should follow the standard schema included in [k8s.io/apimachinery/pkg/apis/meta/v1/types.go](http://k8s.io/apimachinery/pkg/apis/meta/v1/types.go). It should be included as a top level element in status, similar to

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md

PR #297, opened 2023-07-06
Store the IngressController CRD for EnvTest

gibizer:


gibizer:
This is still WIP. This will be needed in ironic-operator env test. But we need other steps first:

* to be able to import the IngressController type from [github.com/openshift/api](http://github.com/openshift/api) we would need to bump the version of it from v3.9.0 to something newer. But for that we need to do the bump in lib-common first.
* such bump will happen as part of https://github.com/openstack-k8s-operators/lib-common/pull/293/files#diff-109b7e7cb3e1a1797ee9d007daf0a314e6c927e31351eb1b82a4ca20a451d7b7R73

gibizer:
It seem ironic does not need this.

PR #299, opened 2023-07-09
Set proper podnames on simulated statefulset with pods

hjensas:
Use the statfuleset name.Name + replica index to set podName for simulated statefulset pods.

Also remove volume mounts from all InitContainers as well.


modules/common/test/helpers/statefulset.go
```diff
@@ -82,6 +83,9 @@ func (tc *TestHelper) SimulateStatefulSetReplicaReadyWithPods(name types.Namespa
 		for i := range pod.Spec.Containers {
 			pod.Spec.Containers[i].VolumeMounts = []corev1.VolumeMount{}
 		}
+		for i := range pod.Spec.InitContainers {
+			pod.Spec.InitContainers[i].VolumeMounts = []corev1.VolumeMount{}
+		}
```
gibizer, line 88: could you please do this in SimulateDeploymentReadyWithPods as well? Then I can drop https://github.com/openstack-k8s-operators/lib-common/pull/302

modules/common/test/helpers/statefulset.go
```diff
@@ -70,7 +71,7 @@ func (tc *TestHelper) SimulateStatefulSetReplicaReadyWithPods(name types.Namespa
 			Spec:       ss.Spec.Template.Spec,
 		}
 		pod.ObjectMeta.Namespace = name.Namespace
-		pod.ObjectMeta.GenerateName = name.Name
+		pod.ObjectMeta.Name = fmt.Sprintf("%s-%d", name.Name, i)
```
gibizer, line 74: +1

gibizer (approved):
Looks good. Thanks @hjensas

PR #300, opened 2023-07-10
Add common ready conditions/messages for Memcached/TransportURL

kajinamit:
Now we define individual ready conditions and messages for Memcached and TransportURL but this is redundant because these implementations are almost same. This adds these into the lib-common module so that we can reuse the same definition.


stuggi (approved):
/lgtm

PR #301, opened 2023-07-11
Remove circular import common <-> test module

stuggi:
To remove circular import, this moves modules/test/helpers into common/test/helpers.
Also integrates the modules/test-operators into modules/test and removes the need for the modules/test-operators module.


gibizer (approved):
While the change looks extensive it is just moving functions around and dissolving test-operators in the process. @stuggi did the placement adaptation, I did the nova adaptation PRs and both is fairly simple and passing the envtest job. So I'm confident in this change.

PR #306, opened 2023-07-19
Pin ocp api

stuggi:


modules/common/go.mod
```diff
@@ -79,3 +79,7 @@ require (
 	github.com/kr/pretty v0.3.1 // indirect
 	github.com/rogpeppe/go-internal v1.10.0 // indirect
 )
+
+// mschuppert: map to latest commit from release-4.13 tag
+// must consistent within modules and service operators
+replace github.com/openshift/api => github.com/openshift/api v0.0.0-20230414143018-3367bc7e6ac7
```
gibizer, line 85: Is there a reason to do this as a replace instead of updating L11 directly?

Makefile
```diff
@@ -17,10 +17,10 @@ CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
 ENVTEST ?= $(LOCALBIN)/setup-envtest
 
 ## Tool Versions
-CONTROLLER_TOOLS_VERSION ?= v0.9.2
+CONTROLLER_TOOLS_VERSION ?= v0.10.0
```
gibizer, line 20: that is what we use in nova-operator

Makefile
```diff
@@ -17,10 +17,10 @@ CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
...
 
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.24
+ENVTEST_K8S_VERSION = 1.25
```
gibizer, line 23: that is what we use in nova-operator

modules/common/go.mod
abays: It might be because various `go` commands automatically find `github.com/openshift/api v3.9.0+incompatible` and overwrite what you may have manually added to the `require` section.  At least I feel I remember seeing that happen to me in the past.

stuggi: we could for common module, but where its indirect you'd have to use replace. I thoughts its more obvious then to do it the same everywhere.

gibizer: Thanks. It seems the indirect dep in modules/test coming from some service operator /api module. so we cannot avoid that.

gibizer (approved):

PR #307, opened 2023-07-20
fix memcached replica change int32 -> *int32

stuggi:


gibizer (approved):
The fix is valid but the pre-commit failure seems relevant.

PR #313, opened 2023-07-26
Allow customize service via service.Override and handle endpoint protocol

stuggi:
- Allow customize service via service.Override

    
Adds an OverridSpec to the Service which allows to customize metadata.Annotations, metadata.Labels and spec of a service. The override values get merged into the object definition created by the operator. This is part of moving the external connection config into the openstack-operator and allows to override the default service created by an operatior with a MetalLB LoadBalancer service spec.

Jira: OSP-26690

- [TLS] Add Protocol to endpoint

Adds Protocol to endpoint Data struct with funcs to get the protocol based on the type. If the type is nil, http is considered to match the current behavior.
    
Jira: OSP-26845


stuggi:
need rebase when https://github.com/openstack-k8s-operators/lib-common/pull/293 merged

stuggi:
This requires follow up changes to ovn-operator, infra-operator (memcached, redis) and swift operator due to the change in NewService()

modules/common/service/service.go
```diff
@@ -39,12 +41,50 @@ func NewService(
 	service *corev1.Service,
 	labels map[string]string,
 	timeout time.Duration,
-) *Service {
-	return &Service{
+	override *OverrideSpec,
```
stuggi, line 44: This requires follow up changes to ovn-operator, infra-operator (memcached, redis) and swift operator due to the change in NewService()

modules/common/service/service.go
```diff
@@ -280,3 +353,17 @@ func GetServicesPortDetails(
...
+	endpt Endpoint,
+) *OverrideSpec {
+	for _, o := range overrides {
+		if endpt == o.Endpoint {
+			return &o
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
stuggi: yes, I had this initially. the only downside with this is that in the CRD we can not limit the allowed keys using annotations, like it is now with the separate Endpoint parameter. we'd have to do it via a validations webhook then.

modules/common/service/service.go
```diff
@@ -67,42 +108,143 @@ func (s *Service) GetServiceHostname() string {
...
+		var endptURL string
+		if protocol != nil &&
+			((*protocol == ProtocolHTTP && port == "80") ||
+				(*protocol == ProtocolHTTPS && port == "443")) {
+			endptURL = fmt.Sprintf("%s%s", EndptProtocol(protocol), hostname)
+		} else {
+			endptURL = fmt.Sprintf("%s%s:%s", EndptProtocol(protocol), hostname, port)
+		}
+
+		// Update instance status with service endpoint url from route host information
```
abays, line 170: Is this an old copy-pasted comment?  We're not really updating the instance status here, right?

stuggi: yes thats an outdated comment. will fix that, thanks!

stuggi:
closing in favor of 332

PR #314, opened 2023-07-28
Give specific domain id for user/project

kajinamit:
Keystone supports "domain" and users, groups and projects belong to a specific domain, and it's possible that multiple entries with the same name exist in different domains.

This introduces the interface to specify the domain where the user or the project should exist, to avoid unexpected behavior caused by such "duplicate" entities.

### DIFF
--- a/modules/openstack/user.go
+++ b/modules/openstack/user.go
@@ -95,6 +95,8 @@ func (o *OpenStack) GetUser(
 
 	if len(allUsers) == 0 {
 		return nil, fmt.Errorf(fmt.Sprintf("%s %s", userName, UserNotFound))
+	} else if len(allUsers) > 1 {
+		return nil, fmt.Errorf(fmt.Sprintf("multiple users named \"%s\" found", userName))
 	}
 
 	return &allUsers[0], nil
--- a/modules/openstack/user.go
+++ b/modules/openstack/user.go
@@ -32,6 +32,7 @@ type User struct {
 	Name      string
 	Password  string
 	ProjectID string
+	DomainID  string
 }
 
 // CreateUser - creates user with userName, password and default project projectID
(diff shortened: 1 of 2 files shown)


modules/openstack/user.go
```diff
@@ -76,8 +82,9 @@ func (o *OpenStack) CreateUser(
 func (o *OpenStack) GetUser(
 	log logr.Logger,
 	userName string,
+	domainID string,
```
bshephar, line 85: Maybe if we need to change the function signature here anyway, we could pass in the User struct here? Which might improve the maintainability of this in the future if we need to add other parameters to it?

modules/openstack/user.go
```diff
@@ -97,18 +106,20 @@ func (o *OpenStack) GetUser(
 func (o *OpenStack) DeleteUser(
 	log logr.Logger,
 	userName string,
+	domainID string,
```
bshephar, line 109: Same here. Passing in a User struct might be more suitable moving forward.

modules/openstack/user.go
kajinamit: We can update the sigunature but that would eventually affect all of the functions we have in this openstack module, which can affect a few more operators. So I tend to leave it now but am open for opinion from the others.

stuggi: yes, passing in a user struct would help to easy extend it in future without the immediate change of the consumers. if my grep was correct, only keystone, glance and heat use this module direct. but I guess we could also do it later

stuggi (approved):
/lgtm requires a change to at least keystone

kajinamit:
I'll leave the suggested interface change because;

* We don't use the instance obtained by GetUser when calling DeleteUser so defining a struct might be redundant
* Changing the interface would need consistent update in the all resources
* I don't expect we need to update the interface later. We may need it when Keystone updates data structure but I assume it's very rare.

PR #320, opened 2023-08-11
Add EventRecorder to Helper

fao89:
https://www.cncf.io/blog/2023/03/13/how-to-use-kubernetes-events-for-effective-alerting-and-monitoring/

https://www.cncf.io/blog/2021/12/21/extracting-value-from-the-kubernetes-events-feed/

https://www.atlassian.com/engineering/how-to-export-kubernetes-events-for-observability-and-alerting

Events: https://gist.githubusercontent.com/fao89/f25d02317adfebee8943055be23a4553/raw/cf40e9197b6507a67722f334722ac5fa6c382270/events.log

### DIFF
--- a/modules/common/configmap/configmap.go
+++ b/modules/common/configmap/configmap.go
@@ -105,7 +105,11 @@ func createOrPatchConfigMap(
 		return nil
 	})
 	if err != nil {
+		h.GetRecorder().Event(obj, corev1.EventTypeWarning, "ConfigMapError", fmt.Sprintf("error create/updating configmap: %s", cm.Name))
 		return "", op, fmt.Errorf("error create/updating configmap: %w", err)
+	}
+	if op == controllerutil.OperationResultCreated {
+		h.GetRecorder().Event(obj, corev1.EventTypeNormal, "ConfigMapCreated", fmt.Sprintf("configmap %s created", cm.Name))
 	}
 
 	configMapHash, err := Hash(configMap)
--- a/modules/common/daemonset/daemonset.go
+++ b/modules/common/daemonset/daemonset.go
@@ -24,6 +24,7 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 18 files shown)


fao89:
@stuggi @gibizer what do you think about this?

gibizer:
> @stuggi @gibizer what do you think about this?

I don't have much context about events in k8s. I think it worth to create a doc where you propose what will be the use of the events in openstack-k8s-operators. I.e. what step during a reconciliation worth emitting an event. What information an event should always contain, etc. A bit similar to the definition of the condition usage https://github.com/openstack-k8s-operators/docs/blob/main/conditions.md

stuggi:
> > @stuggi @gibizer what do you think about this?
> 
> I don't have much context about events in k8s. I think it worth to create a doc where you propose what will be the use of the events in openstack-k8s-operators. I.e. what step during a reconciliation worth emitting an event. What information an event should always contain, etc. A bit similar to the definition of the condition usage https://github.com/openstack-k8s-operators/docs/blob/main/conditions.md

I agree with @gibizer , I think we should not just create events, first we'd have to agree/define when there should be one and the content.

fao89:
I think this can help us to detect race conditions, we can sort events and see the chronological order things were added/updated in the cluster

modules/common/configmap/configmap.go
```diff
@@ -105,8 +105,12 @@ func createOrPatchConfigMap(
 		return nil
 	})
 	if err != nil {
+		h.GetRecorder().Event(obj, corev1.EventTypeWarning, "ConfigMapError", fmt.Sprintf("error create/updating configmap: %s", cm.Name))
```
stuggi, line 108: re `ConfigMapError` and other events, its probably good that we create a central reason collection which then gets used in all the operators. similar to how we do it with the conditions. Otherwise I guess if events are used in the service operators, we'll see different reasons used. I could see two ways. Either we create an events pkg and have the consts for reasons in there, or they are part of the already existing pkgs, like ConfigMap reason in configmap. but then how to handle reasons for stuff we do not yet have. create extra pkgs, or better the one events pkg. thoughts?

modules/common/daemonset/daemonset.go
```diff
@@ -76,8 +77,12 @@ func (d *DaemonSet) CreateOrPatch(
...
 			return ctrl.Result{RequeueAfter: d.timeout}, nil
 		}
+		h.GetRecorder().Event(h.GetBeforeObject(), corev1.EventTypeWarning, "DaemonSetError", fmt.Sprintf("error create/updating daemonset: %s", d.daemonset.Name))
 		return ctrl.Result{}, err
 	}
+	if op == controllerutil.OperationResultCreated {
+		h.GetRecorder().Event(h.GetBeforeObject(), corev1.EventTypeNormal, "DaemonSetCreated", fmt.Sprintf("daemonset %s created", d.daemonset.Name))
+	}
 	if op != controllerutil.OperationResultNone {
 		util.LogForObject(h, fmt.Sprintf("DaemonSet: %s", op), daemonset)
```
stuggi, line 87: should there be an updated event when the object got updated?

gibizer:
> > @stuggi @gibizer what do you think about this?
> 
> I don't have much context about events in k8s. I think it worth to create a doc where you propose what will be the use of the events in openstack-k8s-operators. I.e. what step during a reconciliation worth emitting an event. What information an event should always contain, etc. A bit similar to the definition of the condition usage https://github.com/openstack-k8s-operators/docs/blob/main/conditions.md

I'm worried that this PR is moving forward without agreeing on the scope of it.

stuggi:
> > > @stuggi @gibizer what do you think about this?
> > 
> > 
> > I don't have much context about events in k8s. I think it worth to create a doc where you propose what will be the use of the events in openstack-k8s-operators. I.e. what step during a reconciliation worth emitting an event. What information an event should always contain, etc. A bit similar to the definition of the condition usage https://github.com/openstack-k8s-operators/docs/blob/main/conditions.md
> 
> I'm worried that this PR is moving forward without agreeing on the scope of it.

agreed, @fao89 do you plan to work on some definition doc as we have for conditions?

fao89:
by the community guide:
```
Events are expected to provide important insights for the application developer/operator on the state
of their application. Events relevant to cluster administrators are acceptable, as well, though they usually
also have the option of looking at component logs. Events are much more expensive than logs, thus they're not
expected to provide in-depth system debugging information. Instead concentrate on things that are important 
from the application developer's perspective. Events need to be either actionable, or be useful to understand 
past or future system's behavior. Events are not intended to drive automation. Watching resource status should 
be sufficient for controllers.
```
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/event-style-guide.md#when-to-emit-an-event

I think we can close this PR

PR #332, opened 2023-09-07
Allow customize service via service.Override and handle endpoint protocol 

stuggi:
* Allow customize service via service.Override
Adds an OverrideSpec to the Service which allows to customize metadata.Annotations, metadata.Labels and spec of a service. The override values get merged into the object definition created by the operator. This is part of moving the external connection config into the openstack-operator and allows to override the default service created by an operatior with a MetalLB LoadBalancer service spec.

Jira: OSP-26690

* [TLS] Add Protocol to endpoint
Adds Protocol to endpoint Data struct with funcs to get the protocol based on the type. If the type is nil, http is considered to match the current behavior.

* Add annotations for ingress control
Adds AnnotationIngressCreateKey . Setting this annotation on the service will control the openstack-operator to create a route .

Jira: OSP-26845


modules/common/route/types.go
```diff
@@ -22,16 +22,20 @@ import (
 	"time"
 
 	routev1 "github.com/openshift/api/route/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // Route -
+// +kubebuilder:object:generate:=false
```
bogdando, line 29: why don't we want these in zz_generated?

stuggi: it is not a struct which we use in the API

modules/common/service/types.go
```diff
@@ -67,3 +110,136 @@ const (
...
 	MetalLBLoadBalancerIPs = "metallb.universe.tf/loadBalancerIPs"
 )
+
+// OverrideSpec - service override configuration for the Service created to serve traffic to the cluster.
+// Allows for the manifest of the created Service to be overwritten with custom configuration.
+type OverrideSpec struct {
+	// EndpointURL to be used to register the service in keystone.
+	// +optional
+	EndpointURL *string `json:"endpointURL,omitempty"`
+
```
gibizer, line 120: I think we are mixing two use cases.

1. Expose a service via a Route created by the openstack-operator
    This to work we need both the Service override for the openstack-operator to pass down a label selector to the service operator, and the endpointURL for the openstack-operator to pass down the URL of the created Route.
2. Allow the end user to specify extra configuration for the k8s Service created by a service operator.
    For this we only need the Service override but not the endpointURL

Do we have services that only need use case 2) but not 1) and therefore does no need the endpointURL? 

At least I see that in nova-operator nova-metadata will have a service override (as it will have a k8s Service) but openstack-operator will not create a Route for it (as this API is not for external consumption) and therefore the endpointURL won't ever be set.

Can we decouple the two use cases? The use case 1) could use the current struct while the use case 2) could use a different struct that only has EmbeddedLabelsAnnotations and OverrideServiceSpec fields but not EndpointURL (as that would be unused).

We could even avoid duplicated definition with
```golang
type ServiceOverrideSpec struct {
    *EmbeddedLabelsAnnotations `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
	Spec *OverrideServiceSpec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
}

type RoutedServiceOverrideSpec struct {
    ServiceOverrideSpec
    EndpointURL *string `json:"endpointURL,omitempty"`
}
```

stuggi: yes that is correct, we have some services which don't require the EndpointURL, without checking its, as you said NovaMetadata, and DNSmasq. maybe there are more, or there will be a new service. I'll update the PR to support that situation.

stuggi: done

gibizer (commented):
I don't have any open question left in this PR. So I'm +1 on this. I expect that other service operator maintainers will take a look too.

stuggi:
rebased

modules/common/service/service.go
```diff
@@ -67,42 +108,143 @@ func (s *Service) GetServiceHostname() string {
...
-// GetServiceHostnamePort - returns the service hostname with port
-func (s *Service) GetServiceHostnamePort() string {
-	return fmt.Sprintf("%s:%d", s.GetServiceHostname(), GetServicesPortDetails(s.service, s.service.Name).Port)
+// GetServiceHostnamePort - returns the service hostname with port if service port
+// is not nil, otherwise returns GetServiceHostname()
+func (s *Service) GetServiceHostnamePort() (string, string) {
+	servicePort := GetServicesPortDetails(s.service, s.service.Name)
+	if servicePort != nil {
+		return s.GetServiceHostname(),
+			strconv.FormatInt(int64(GetServicesPortDetails(s.service, s.service.Name).Port), 10)
```
abays, line 117: Since we acquired `servicePort` on L112, couldn't we use it here instead of calling `GetServicesPortDetails(s.service, s.service.Name)` again?

stuggi: right, will change that

stuggi: done

abays (approved):
/lgtm

modules/common/endpoint/endpoint.go
```diff
@@ -73,12 +82,14 @@ type MetalLBData struct {
 }
 
 // ExposeEndpoints - creates services, routes and returns a map of created openstack endpoint
+// NOTE: (mschuppert) deprecated, can be removed when external endpoint creation moved to openstack-operator
+// and ExposeEndpoints() can be removed
 func ExposeEndpoints(
```
fmount, line 87: ack, we still have this here but after the whole chain is merged we can do some cleanup.

modules/common/endpoint/endpoint.go
```diff
@@ -184,7 +203,6 @@ func ExposeEndpoints(
 						ServiceName:    endpointName,
 						TargetPortName: endpointName,
 					}),
-					exportLabels,
 					timeout,
 					data.RouteOverride,
```
fmount, line 207: @stuggi quick question here: when a `GenericRoute` is created, in case `tls` is used we're terminating the ssl on the Route, w/o re-encrypting the second segment (from the `Route` to the `Service`). 
Is my understanding correct? 
This would be a good behavior to keep and provide parameters via `RouteOverride` to change the behavior to `re-encrypt`. 
For instance Glance atm doesn't support tls, and in TripleO we used to deploy an httpd in front of GlanceAPI to properly handle tls.
This is something I'm going to work on in the next few weeks, but wanted to check that we actually have the ability to terminate the connection on the `Route` until that part is done.

stuggi:
rebased

modules/common/endpoint/endpoint.go
fmount: yes, I'll wait to see the first patch series landed, and when doing tls we can sync up and discuss about the Glance use case. We might need to have an additional sidecar container w/ httpd or make it default like it is today in Manila/Cinder.

PR #333, opened 2023-09-08
Add the TLS struct support for service operators

Deydra71:
The service operators will have TLS struct type to pass in the reference to the certificates secrets as well as the option to add multiple CA certs.

Jira: [OSP-26846](https://issues.redhat.com/browse/OSP-26846)

### DIFF
--- a/modules/common/tls/tls.go
+++ b/modules/common/tls/tls.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 Red Hat
+Copyright 2023 Red Hat
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -18,17 +18,27 @@ limitations under the License.
 
 package tls
 
+import (
+	"context"
+	"fmt"
+
+	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
+	corev1 "k8s.io/api/core/v1"
+)
+
 // Service contains server-specific TLS secret
 type Service struct {
-	// Server-specific settings
-	SecretName             string `json:"secretName,omitempty"`
-	DisableNonTLSListeners bool   `json:"disableNonTLSListeners,omitempty"`
+	// +kubebuilder:validation:Optional
+	SecretName string `json:"secretName,omitempty"`
+	// +kubebuilder:validation:Optional
+	DisableNonTLSListeners bool `json:"disableNonTLSListeners,omitempty"`
 }
 
 // Ca contains CA-specific settings, which could be used both by services (to define their own CA certificates)
 // and by clients (to verify the server's certificate)
 type Ca struct {
-	// CA-specific settings
+	// +kubebuilder:validation:Optional
 	CaSecretName string `json:"caSecretName,omitempty"`
 }
 
@@ -36,4 +46,82 @@ type Ca struct {
 type TLS struct {
 	Service *Service `json:"service"`
 	Ca      *Ca      `json:"ca"`
+}
+
+// NewTLS - initialize and return a TLS struct
+func NewTLS(ctx context.Context, h *helper.Helper, namespace string, service *Service, ca *Ca) (*TLS, error) {
+
+	// Ensure service SecretName exists or return an error
+	if service != nil && service.SecretName != "" {
+		secretData, _, err := secret.GetSecret(ctx, h, service.SecretName, namespace)
+		if err != nil {
+			return nil, fmt.Errorf("error ensuring secret %s exists: %w", service.SecretName, err)
+		}
+
+		_, keyOk := secretData.Data["tls.key"]
+		_, certOk := secretData.Data["tls.crt"]
+		if !keyOk || !certOk {
+			return nil, fmt.Errorf("secret %s does not contain both tls.key and tls.crt", service.SecretName)
+		}
+	}
+
+	return &TLS{
+		Service: service,
+		Ca:      ca,
+	}, nil
+}
+
+// CreateVolumeMounts - add volume mount for TLS certificate and CA certificates, this counts on openstack-operator providing CA certs with unique names
+func (t *TLS) CreateVolumeMounts() []corev1.VolumeMount {
+	var volumeMounts []corev1.VolumeMount
+
+	if t.Service != nil && t.Service.SecretName != "" {
+		volumeMounts = append(volumeMounts, corev1.VolumeMount{
+			Name:      "tls-certs",
+			MountPath: "/var/lib/config-data/tls-certificates",
+			ReadOnly:  true,
+		})
+	}
+
+	if t.Ca != nil && t.Ca.CaSecretName != "" {
+		volumeMounts = append(volumeMounts, corev1.VolumeMount{
+			Name:      "ca-certs",
+			MountPath: "/var/lib/config-data/ca-certificates",
+			ReadOnly:  true,
+		})
+	}
+
+	return volumeMounts
+}
+
+// CreateVolumes - add volume for TLS certificate and CA certificates
+func (t *TLS) CreateVolumes() []corev1.Volume {
+	var volumes []corev1.Volume
+	mode := int32(0400)
+
+	if t.Service != nil && t.Service.SecretName != "" {
+		volumes = append(volumes, corev1.Volume{
+			Name: "tls-certs",
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName:  t.Service.SecretName,
+					DefaultMode: &mode,
+				},
+			},
+		})
+	}
+
+	if t.Ca != nil && t.Ca.CaSecretName != "" {
+		volumes = append(volumes, corev1.Volume{
+			Name: "ca-certs",
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName:  t.Ca.CaSecretName,
+					DefaultMode: &mode,
+				},
+			},
+		})
+	}
+
+	return volumes
 }
--- a/modules/common/tls/tls_test.go
+++ b/modules/common/tls/tls_test.go
@@ -1,0 +1,109 @@
+/*
+Copyright 2023 Red Hat
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tls
+
+import (
+	"testing"
+)
+
+func TestCreateVolumeMounts(t *testing.T) {
+	tests := []struct {
+		name          string
+		service       *Service
+		ca            *Ca
+		wantMountsLen int
+	}{
+		{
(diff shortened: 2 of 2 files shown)


modules/common/tls/tls.go
```diff
@@ -0,0 +1,235 @@
...
+			})
+	}
+
+	if tls.Ca.CaSecretName != "" {
+		for i := 0; i < MaxNumberOfCAs; i++ {
+			volumeMounts = append(volumeMounts, corev1.VolumeMount{
+				Name:      fmt.Sprintf("ca-cert-%d", i),
+				MountPath: fmt.Sprintf("/etc/ca-certs/%d", i),
+				ReadOnly:  true,
+				SubPath:   fmt.Sprintf("ca-%d.crt", i),
```
dciabrin, line 128: I'm not sure why, but on my env this volume mount would never expose the secret's certificate when SubPath was used. Maybe let's get rid of it and only  rely on volumes to expose a single kay "ca.crt"?

modules/common/tls/tls.go
```diff
@@ -0,0 +1,235 @@
...
+	if tls.Ca.CaSecretName != "" {
+		for i := 0; i < MaxNumberOfCAs; i++ {
+			volumes = append(volumes, corev1.Volume{
+				Name: fmt.Sprintf("ca-cert-%d", i),
+				VolumeSource: corev1.VolumeSource{
+					Secret: &corev1.SecretVolumeSource{
+						SecretName: tls.Ca.CaSecretName,
+						Items: []corev1.KeyToPath{
+							{
+								Key:  fmt.Sprintf("ca-%d.crt", i),
```
dciabrin, line 170: This assumes that a secret will be stored in a key with a non-standard name, while CA are usually referenced by key "ca.crt". This is the case for e.g. certs created by cert-manager.
We have to support that naming convention and we should assume the CA would always be referenced by key  "ca.crt"

modules/common/tls/tls.go
```diff
@@ -0,0 +1,235 @@
...
+							Key:  "tls.key",
+							Path: "tls.key",
+						},
+					},
+				},
+			},
+		})
+	}
+
+	if tls.Ca.CaSecretName != "" {
```
dciabrin, line 161: My understanding is that we plan to support two high level use cases:
 1. we create certificates for the control plane services, so we need one CA cert to verify all those certificates  when doing TLS communication between pods.
 2. we need to support additional CA in case specific services require to connect to endpoints for which we don't create certificate.

Given that we always need 1. for TLS, and 2. seems to be unbounded, could we expose that in the signature of CreateVolumes, like maybe e.g.
```
CreateVolumes(tls *TLSService, ca []TLSCa)
``` 

This way we wouldn't need MaxNumberOfCAs

modules/common/tls/tls.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// CreateVolumeMounts - add volume mount for TLS certificate and CA certificates
+func CreateVolumeMounts(tls *TLS) []corev1.VolumeMount {
+	var volumeMounts []corev1.VolumeMount
+
+	if tls.Service.SecretName != "" {
+		volumeMounts = append(volumeMounts, corev1.VolumeMount{
+			Name:      "tls-certs",
+			MountPath: "/etc/tls-certs",
```
dciabrin, line 110: I have some doubts on directly bind-mounting certs in /etc, because then the permission of the certs/ private keys and whatnot must be managed by this volumeMount.
We could bind-mount it under /var/{something}, and let Kolla config copy it in /etc/tls-certs and specify the  permissions in the Kolla config file? (for the record, this is how we use to do it with TripleO)

modules/common/tls/tls.go
Deydra71: Thanks for pointing this one. I didn't realize that when we use different mounthPath for each ca.crt, we don't have to use different names for individual CA cert.

modules/common/tls/tls.go
Deydra71: Yes, in general we need multiple CA cert - one for validate the certs used between the pods, meaning the tls.crt certs. Other CA certs can be used to validate certificates provided by external clients.

So yes, you concluded it correctly in points 1. and 2. And there's no issue with using:
`CreateVolumes(tls *TLSService, ca []TLSCa)`

modules/common/tls/tls.go
Deydra71: Agree, changed to `/var/tls-certs` and /`var/ca-certs` instead.

modules/common/tls/tls.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	return &TLS{
+		Service: service,
+		Ca:      ca,
+	}, nil
+}
+
+// EnsureOrCreateSecret - checks if a secret with the given name exists in the specified namespace
+func EnsureOrCreateSecret(clientset *kubernetes.Clientset, namespace, secretName string, data map[string][]byte) error {
```
stuggi, line 77: lib-common common module has funcs to manage secret. don't think we need an extra one? also I am not sure if we want to create secrets. we just return an error if the referenced secret with the name does not exist.

modules/common/tls/tls.go
```diff
@@ -0,0 +1,198 @@
...
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName:  tls.SecretName,
+					DefaultMode: &mode,
+				},
+			},
+		})
+	}
+
+	for i, caCert := range ca {
```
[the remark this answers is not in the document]
Deydra71: > this now expects that we have a secret per CA where the ca is referenced by "ca.crt", right?
Yes, that's correct and I agree that it's not practical

> The service operators than would only have to deal with two secrets, one for the service cert (which might also have the ca. crt + the combined CA secret)
I am little bit confused with this sentence. By the service cert you mean tls.crt and tls.key, witch is under the TLSService struct, right? Then we have the TLSCa struct that consists of CA certs: 

1. Our internal CA cert
2. The CA certs provided by the end user

And you are saying that the openstack-operator will create one secret that combines all the CA certs. So ultimately we will have two secrets - one for TLS service and one for all the CA certs. Can you please tell me if I understand correctly?

Then yes, it makes more sense to use unique CA cert file names rather then have unique secrets for each one of them named ca.crt. WHich would be incompatible with what openstack-operator creates. In other words when deciding between **unique CA cert file names vs. unique secrets**, we go with **unique CA cert file names**, is this correct?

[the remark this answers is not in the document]
Deydra71: You're right @stuggi , I made a mistake and did not realize that I am iterating over multiple TLSCa secrets. Also if I understood you correctly, the actual CA secret will be provided by the openstack-operator, so the tls package doesn't have to take care of creating them. Same applies for the TLSService secret.

Deydra71:
Pending changes when #345  is merged.

stuggi:
I think we should also squash all the commits into one

modules/common/tls/tls.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// TLS - a generic type, which encapsulates both the service and CA configurations
+type TLS struct {
+	Service Service `json:"service"`
+	Ca      Ca      `json:"ca"`
+}
+
+// NewTLS - initialize and return a TLS struct
+func NewTLS(cr client.Object, clientset *kubernetes.Clientset, scheme *runtime.Scheme, namespace string, service Service, ca Ca) (*TLS, error) {
```
stuggi, line 55: should also pass in the context, instead of use `context.TODO()`

stuggi: how should we handle case where it is not a service, and instead only requires the CA volume mount. could use ptr to Service and CA  for and when service is nil we only return the volume mounts for the CA?

modules/common/tls/tls.go
```diff
@@ -0,0 +1,192 @@
...
+			Name:      "tls-certs",
+			MountPath: "/var/tls-certs",
+			ReadOnly:  true,
+		})
+	}
+
+	if ca.CaSecretName != "" {
+		volumeMounts = append(volumeMounts, corev1.VolumeMount{
+			Name:      "ca-certs",
+			MountPath: "/var/ca-certs",
```
stuggi, line 111: lets put them in `/var/lib/config-data/ca-certificates`, thats where I set to expect them in https://github.com/openstack-k8s-operators/tcib/pull/82

modules/common/tls/tls.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// CreateVolumeMounts - add volume mount for TLS certificate and CA certificates, this counts on openstack-operator providing CA certs with unique names
+func CreateVolumeMounts(tls *Service, ca *Ca) []corev1.VolumeMount {
+	var volumeMounts []corev1.VolumeMount
+
+	if tls.SecretName != "" {
+		volumeMounts = append(volumeMounts, corev1.VolumeMount{
+			Name:      "tls-certs",
+			MountPath: "/var/tls-certs",
```
stuggi, line 103: lets put them in `/var/lib/config-data/tls-certificates`

modules/common/tls/tls.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	return &TLS{
+		Service: service,
+		Ca:      ca,
+	}, nil
+}
+
+// CreateVolumeMounts - add volume mount for TLS certificate and CA certificates, this counts on openstack-operator providing CA certs with unique names
+func CreateVolumeMounts(tls *Service, ca *Ca) []corev1.VolumeMount {
```
stuggi, line 97: should this be `func (t *TLS) CreateVolumeMounts()`

modules/common/tls/tls.go
```diff
@@ -0,0 +1,192 @@
...
+			MountPath: "/var/ca-certs",
+			ReadOnly:  true,
+		})
+	}
+
+	return volumeMounts
+}
+
+// CreateVolumes - add volume for TLS certificate and CA certificates
+func CreateVolumes(tls *Service, ca *Ca) []corev1.Volume {
```
stuggi, line 120: same as above, should this be `func (t *TLS) CreateVolumes()`

modules/common/tls/tls.go
```diff
@@ -0,0 +1,192 @@
...
+				},
+			},
+		})
+	}
+
+	return volumes
+}
+
+// SetProbes sets the liveness and readiness probes based on the TLS configuration, disable HTTPS for debugging purposes
+func SetProbes(port int, disableNonTLSListeners bool) (livenessProbe *corev1.Probe, readinessProbe *corev1.Probe, err error) {
```
stuggi, line 152: this Probes func might be better located in a probes package instead of the tls here.

stuggi: we'd also need to pass in the the path v3 is not used everywhere

stuggi: maybe just a full `corev1.Probe` the timeouts can be different depending on the service. there might be also a difference needed between liveness and readiness ...

stuggi (approved):
/lgtm

modules/common/tls/tls.go
```diff
@@ -37,3 +50,81 @@ type TLS struct {
 	Service *Service `json:"service"`
 	Ca      *Ca      `json:"ca"`
 }
+
+// NewTLS - initialize and return a TLS struct
+func NewTLS(ctx context.Context, cr client.Object, clientset *kubernetes.Clientset, scheme *runtime.Scheme, namespace string, helperInstance *helper.Helper, service *Service, ca *Ca) (*TLS, error) {
```
stuggi, line 55: we don't need to pass in the cr, clientset and scheme its not used, right?

```
func NewTLS(ctx context.Context, namespace string, helperInstance *helper.Helper, service *Service, ca *Ca) (*TLS, error) {
```

Deydra71: You're right, they're remnants from the previous definition. Pushing correction in a moment.

PR #337, opened 2023-09-18
[envtest]Add UpdateSecret helper

gibizer:
This is pulled out from nova-operator as it is generic enough to be reused in other repos like in placement-operator


stuggi (approved):
/lgtm

PR #338, opened 2023-09-21
Helper function for TLS support in database access

dciabrin:
New function to generate the mysql flags read by oslo.db to connect to a mysql database via TLS.
This commit also returns a fqdn for the database hostname, as it is required to validate the mysql database's certificate.


dciabrin:
This PR depends on the new TLS types defined in https://github.com/openstack-k8s-operators/lib-common/pull/333

PR #343, opened 2023-09-27
Merge Annotations and Labels in CreateOrPatch

stuggi:
Identified in OCP4.13 where metallb adds an annotation to the service when handling it, it resulted in reconcile loop since the current service CreateOrPatch did not expect that annotations will be altered outside the owner.

This changes the different functions to merge Labels and Annotations in CreateOr Patch instead of assigning.


abays (approved):
/lgtm

gibizer (commented):
I agree that we needed to do this but I wan't to note that we lost one use case by this change. Before this change we could remove a label or annotation from our CR and then it was removed from the underlying secret/cm/service resource. Now it is not possible to remove labels or annotations as Merge can always add but never remove entity.

PR #351, opened 2023-09-29
Allow route overrides list in NewRoute() and add EnsureCert()

stuggi:
* updates NewRoute() to support a list of overrides. Later overrides from the list will override parameters from previous. This allows to patch  custom provided overrides with the one for the TLS public endpoint by the openstack-operator.
* Adds EnsureCert() to request a certificate from an issuer and returns the a ptr to the secret holding the cert.

Jira: [OSP-26299](https://issues.redhat.com//browse/OSP-26299)

### DIFF
--- a/modules/certmanager/certificate.go
+++ b/modules/certmanager/certificate.go
@@ -22,13 +22,18 @@ import (
 	"time"
 
 	certmgrv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
+	certmgrmetav1 "github.com/cert-manager/cert-manager/pkg/apis/meta/v1"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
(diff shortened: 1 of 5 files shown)


modules/certmanager/certificate.go
```diff
@@ -128,3 +133,97 @@ func (c *Certificate) Delete(
...
+
+	err := helper.GetClient().Get(ctx, types.NamespacedName{Name: issuerName, Namespace: namespace}, issuer)
+	if err != nil {
+		err = fmt.Errorf("Error getting issuer %s/%s - %w", issuerName, namespace, err)
+
+		return nil, ctrl.Result{}, err
+	}
+
+	// create certificate adding fqdn, but also short names as DNSNames
+	// TODO is that something we want to do, or just the passed in hostnames?
```
stuggi, line 174: @olliewalsh @Deydra71 do we want to do this or just the fqdn?

stuggi: probably the fqdn is enough as thats what the service gets configured for

stuggi: removed it

Deydra71 (approved):

PR #355, opened 2023-10-04
Addin specialized utility function for genereting mock ssh secrets

jpodivin:
These secrets are used in a considerable number of tests. Implementing separate function for generating them in each repo is suboptimal.

At this point this would see use in the dataplane-operator repo, but I expect that in the future a similar fixture would be useful for tests of other operators that have to use ssh connections.


stuggi (approved):
/lgtm

PR #357, opened 2023-10-05
Helper function for TLS support in database access

dciabrin:
New function to generate the mysql flags read by oslo.db to connect to a mysql database via TLS.
This commit also returns a fqdn for the database hostname, as it is required to validate the mysql database's certificate.

### DIFF
--- a/modules/common/tls/tls.go
+++ b/modules/common/tls/tls.go
@@ -21,6 +21,7 @@ package tls
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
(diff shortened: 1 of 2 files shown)


modules/common/tls/tls.go
```diff
@@ -125,3 +126,27 @@ func (t *TLS) CreateVolumes() []corev1.Volume {
...
+// CreateDatabaseClientConfig - connection flags for the MySQL client
+// Configures TLS connections for clients that use TLS certificates
+// returns a string of mysql config statements
+func (t *TLS) CreateDatabaseClientConfig() string {
+	conn := []string{}
+	// This assumes certificates are always injected in
+	// a common directory for all services
+	if t.Service.SecretName != "" {
+		conn = append(conn,
+			"ssl-cert=/etc/pki/tls/certs/tls.crt",
```
stuggi, line 139: is this for when client certs are used?

modules/common/tls/tls.go
```diff
@@ -125,3 +126,27 @@ func (t *TLS) CreateVolumes() []corev1.Volume {
...
+			"ssl-key=/etc/pki/tls/private/tls.key")
+	}
+	// Client uses a CA certificate that will get merged
+	// into the pod's CA bundle by the pod's init container
```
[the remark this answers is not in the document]
dciabrin: It's copied during kolla_start, but the idea was that this file was merged with 'update-ca-trust' during the init container.

https://github.com/openstack-k8s-operators/keystone-operator/blob/1e32653ddc594eda8428fc0335608b7befbb1a14/templates/keystoneapi/bin/init.sh#L44

modules/common/tls/tls.go
dciabrin: correct, it might be useful in the future if we enable client cert-verification in mariadb.
The idea is that this name is generic, and the real certificate for the service is copied by kolla_start into this generic location.
https://github.com/openstack-k8s-operators/keystone-operator/blob/1e32653ddc594eda8428fc0335608b7befbb1a14/templates/keystoneapi/config/keystone-api-config.json#L35

modules/common/tls/tls.go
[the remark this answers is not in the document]
stuggi: but we don't have to do this in the init container as you can see in the tcib change. the deployment get the cacert mounted and kolla does it already.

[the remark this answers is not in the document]
stuggi: I think the plan is to get rid of any init container where possible. for client cert we then should mount the cert and use the kolla config to copy it to the right place. the CA update is handled when we get new images from the above PR. if you need test images, I have built some at quay.io/mschuppe/...

modules/common/tls/tls.go
dciabrin: We might want to use a similar copy mechanism in kolla_start, and agree on a cert name to use here, rather than updating all service's kolla config to copy them manually?

stuggi: yes, we could do that and have pre-defined names for client/server cert

stuggi (approved):
/lgtm

PR #362, opened 2023-10-09
Removed duplicated conditional and improved logging message

jpodivin:
Simple change, so we don't ask the same question twice. 
The logging adjustment is intended to somewhat improve our situation if there is problem with job retrieval.


stuggi (approved):
/lgtm

PR #363, opened 2023-10-10
Restoring PROC_CMD var and normalizing test commands

jpodivin:
Minimal formatting change was introduced in the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,9 @@ CONTROLLER_TOOLS_VERSION ?= v0.10.0
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.25
 
+# Number of CPUs to be allocacted for testing
+PROCS?=$(shell expr $(shell nproc --ignore 2) / 2)
+PROC_CMD = --procs $(PROCS)
 
 .PHONY: all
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -23,6 +23,9 @@ CONTROLLER_TOOLS_VERSION ?= v0.10.0
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.25
 
+# Number of CPUs to be allocacted for testing
+PROCS?=$(shell expr $(shell nproc --ignore 2) / 2)
+PROC_CMD = --procs ${PROCS}
```
dciabrin, line 28: isn't it $(PROCS), otherwise shell will read it instead (and will probably get confused with the single $ ?)

jpodivin: AFAIK, the makefile syntax for vars works both ways. [0] In practice you can expand variable with `${}` or with `$()`
In case of this specific var I have decided to go with prevalent approach in the project
[0]https://www.gnu.org/software/make/manual/html_node/Reference.html#Reference

dciabrin: Thanks for the ref.

dciabrin (approved):

stuggi (approved):
/lgtm

PR #365, opened 2023-10-11
move certmanager test helper into certmanager module

stuggi:


gibizer (commented):
I think the TestHelper in modules/test is now unused so we can remove that as well.

gibizer (approved):
Thanks @stu

gibizer:
We need adaption patches to the service operators (sorry I did not noticed this before):
* nova
* ansibleee-operator
* infra
* keystone
* placement
* manila
* ovn
* neutron
* ironic
* heat
* horizon
* dataplane
* glance
* switft

PR #378, opened 2023-10-26
Add TLSInputReadyCondition

Deydra71:
Indicating the condition when required TLS sources are ready.


stuggi (approved):
/lgtm

stuggi (commented):
do we also want to add specific TLS messages or reuse https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/condition/conditions.go#L126 ? maybe we add custom, especially for the error case ?

PR #384, opened 2023-11-06
Add TLS support for multiple service endpoints

Deydra71:
Some services create multiple endpoints, therefore each endpoint needs to support TLS.


stuggi (commented):
@Deydra71 @olliewalsh wondering if this would be a more generic definition of the certs we need, also with the possibility to customize the mount paths. this might be useful for the edpm side where default cert location for libvirt/qemu is `/etc/pki/libvirt/` , `/etc/pki/libvirt/private/`, but could also use the config params `cert_file` `key_file` for this. https://wiki.libvirt.org/TLSCreateServerCerts.html#overriding-the-default-locations

```
// Service contains server-specific TLS secret
type Service struct {
	// +kubebuilder:validation:Optional
	// SecretName - holding the cert, key for the service
	SecretName string `json:"secretName,omitempty"`
	// +kubebuilder:validation:Optional
	// CertMount - dst location to mount the service tls.crt cert. Can be used to override the default location which is /etc/tls/<service key>/tls.crt
	CertMount *string `json:"certMount,omitempty"`
	// +kubebuilder:validation:Optional
	// KeyMount - dst location to mount the service tls.key  key. Can be used to override the default location which is /etc/tls/<service key>/tls.key
	KeyMount *string `json:"keyMount,omitempty"`
	// +kubebuilder:validation:Optional
	// CaMount - dst location to mount this cert CA ca.crt cert to. Can be used if the service CA cert should be mounted specifically, e.g. to be set in a service config for validation, instead of the env wide bundle.
	CaMount *string `json:"caMount,omitempty"`
	// +kubebuilder:validation:Optional
	// DisableNonTLSListeners - disable non TLS listeners of the service (if supported)
	DisableNonTLSListeners bool `json:"disableNonTLSListeners,omitempty"`
}

// Ca contains CA-specific settings, which could be used both by services (to define their own CA certificates)
// and by clients (to verify the server's certificate)
type Ca struct {
	// +kubebuilder:validation:Optional
	// CaBundleSecretName - dst location to mount the CA cert bundle
	CaBundleSecretName string `json:"caBundleSecretName,omitempty"`

	// +kubebuilder:validation:Optional
	// +kubebuilder:default="/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
	// CaBundleMount - dst location to mount the CA cert bundle
	CaBundleMount *string `json:"caBundleMount,omitempty"`
}

// TLS - a generic type, which encapsulates both the service and CA configurations
type TLS struct {
	// certificate configuration for API service certs
	APIService map[service.Endpoint]Service `json:"APIService"`
	// certificate configuration for additional arbitrary certs
	Service map[string]Service `json:"service"`
	// CA bundle configuration
	Ca      *Ca      `json:"ca"`
}
```

Deydra71:
I've implemented the changes to the structs and their fields as suggested by @stuggi ^^ Following these updates the functions and tests were revised within the tls package, so it should now be in place for operator testing.

modules/common/tls/tls.go
```diff
@@ -36,129 +37,232 @@ const (
...
+			certMountPath = *s.CertMount
+		}
+
+		keyMountPath := "/etc/pki/tls/private/tls.key"
+		if s.KeyMount != nil {
+			keyMountPath = *s.KeyMount
+		}
+
 		volumeMounts = append(volumeMounts, corev1.VolumeMount{
 			Name:      "tls-crt",
```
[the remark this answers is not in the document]
stuggi: broke with https://github.com/openstack-k8s-operators/lib-common/commit/7daa69cf58c70eb62832761183354a42e291542c

Deydra71:
Correcting right now, so the PR consists only of corresponding commits...

modules/common/tls/tls.go
```diff
@@ -20,145 +20,403 @@ package tls
...
+// APIService - API tls type which encapsulates for API services
+type APIService struct {
 	// +kubebuilder:validation:Optional
-	SecretName string `json:"secretName,omitempty"`
+	// Disabled TLS for the deployment of the service
+	Disabled *bool `json:"disabled,omitempty"`
+
+	// +kubebuilder:validation:optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// The key must be the endpoint type (public, internal)
```
stuggi, line 78: can you update the description. this is old

modules/common/tls/tls.go
```diff
@@ -20,145 +20,403 @@ package tls
...
+	Public GenericService `json:"public,omitempty"`
+
+	// +kubebuilder:validation:optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// The key must be the endpoint type (public, internal)
```
stuggi, line 83: same, can you update the description. this is old

modules/common/tls/tls.go
```diff
@@ -20,145 +20,403 @@ package tls
...
+// Enabled - returns true if tls is configured for the public and internal
+func (a *APIService) Enabled(endpt service.Endpoint) bool {
+	switch endpt {
+	case service.EndpointPublic:
+		return (a.Disabled == nil || (a.Disabled != nil && !*a.Disabled)) && a.Public.SecretName != nil
+	case service.EndpointInternal:
+		return (a.Disabled == nil || (a.Disabled != nil && !*a.Disabled)) && a.Internal.SecretName != nil
+	}
+
+	return false
```
vakwetu, line 126: I'm not convinced this is the best way to define the boolean that determines whether tls is enabled or not.

1. Its likely that - given our stance for not changing the tls status in adoption - we will have situations where tls is enabled only on the public interface.  I understand that this can be represented by disabled: false and setting the public secret name, but this also seems prone to error.  What if someone wanted tls (on either interface) but they forgot to provide the secret name.  In this scenario, the deployment would succeed but we would have no tls.  From a security point of view, I'd rather that the deployment failed, rather than accidentally exposing the interface.
2. A better implementation might be to have booleans for disable_public_tls and disable_internal_tls, which default to false and simply to check those booleans.  A separate check should also be done to confirm that the relevant secrets are set and if not, fail spectacularly.

vakwetu: Or maybe it would fail in ValidateCertSecrets

modules/common/tls/tls.go
```diff
@@ -20,145 +20,403 @@ package tls
...
+			}
+
+			svc = a.Public
+
+		case service.EndpointInternal:
+			if !a.Enabled(service.EndpointInternal) {
+				continue
+			}
+
+			svc = a.Public
```
vakwetu, line 151: Is this right?  Looks the same as the public endpoint.

modules/common/tls/tls.go
stuggi: yes, `ValidateCertSecrets` should catch the situation where the secret is missing, or the content of the secret is wrong. I think we could just move the disable into the `GenericService` struct type, which is iiuc what you'd prefer, right? I'll give that a try how this impacts the current patches

modules/common/tls/tls.go
stuggi: no this is wrong, good catch

modules/common/tls/tls.go
vakwetu: Agree.  Moving disable into GenericService would work.

Deydra71:
With changes TLS support is moved here ---> https://github.com/openstack-k8s-operators/lib-common/pull/428

PR #385, opened 2023-11-07
Add object package with generic ownerref func

stuggi:
Adds object package to the common module with initial functions to add additional non controller owner references to objects. This allows additional owners to watch resources

```
// watch for secrets we added ourselves as additional owners, NOT as controller
Watches(
    &source.Kind{Type: &corev1.Secret{}},
    &handler.EnqueueRequestForOwner{OwnerType: &clientv1.OpenStackClient{}, IsController: false}).
```

Note: Adding as an additional owner will not triggere a reconciliation when the object gets re-created from scratch, like deleting a secret.

### DIFF
--- a/modules/common/object/metadata.go
+++ b/modules/common/object/metadata.go
@@ -1,0 +1,116 @@
+/*
+Copyright 2023 Red Hat
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package object
+
+import (
+	"context"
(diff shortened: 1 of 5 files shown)


modules/common/object/metadata.go
```diff
@@ -0,0 +1,119 @@
...
+	owner client.Object,
+	object client.Object,
+) error {
+	// return if the owner is already in the owner ref list
+	if CheckOwnerRefExist(owner.GetUID(), object.GetOwnerReferences()) {
+		return nil
+	}
+
+	// create owner ref patch
+	patchDiff, patch, err := PatchOwnerRef(owner, object, h.GetScheme())
```
gibizer, line 104: I'm wondering why we need this check and set pattern. The SetOwnerReference handles if the ref already exists. Could we just check that the resulting patch is empty? Or you explicitly want to avoid updating the owner ref here if only the version of the owner changed? (See the last sentence of https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#SetOwnerReference )

stuggi: yes, I think you are right, my main goal was to only call patching the obj if the owner is not already in the list.  I'll remove the above check and check if the patch is empty

gibizer (approved):
looks good to me. Thanks!

abays (approved):
/lgtm

PR #386, opened 2023-11-07
Modify EnsureCert to support certs with IPs and usages

vakwetu:
Add support for usages and ip address SANs in the cert requests

### DIFF
--- a/modules/certmanager/certificate.go
+++ b/modules/certmanager/certificate.go
@@ -42,6 +42,18 @@ type Certificate struct {
 	timeout     time.Duration
 }
 
+// CertificateRequest -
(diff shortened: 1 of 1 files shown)


stuggi (commented):
in general looks good to me, just thinking now that the parameter list grows, maybe we should convert them into a struct type which allows it to be extended later more easy without impact the consumer. could do this in this PR or in a follow up, but then we'd have to update the consumers again. wdyt?

vakwetu:
I think that's a great idea.  I'll do it in this PR.

stuggi (approved):
/lgtm

PR #387, opened 2023-11-09
test: allow fetching CRD from a local replace statement

dciabrin:
Ginkgo test suites require fetching the CRDs of the custom resources used in the tests. Those are fetched from the go.mod and any replace statement is parsed and honoured.
Add the ability to process replace statements that point to a local directory as well, to ease the developement process when a project depends on a feature from another project that has not been merged yet.

### DIFF
--- a/modules/test/crd.go
+++ b/modules/test/crd.go
@@ -18,6 +18,7 @@ import (
 	"go/build"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"golang.org/x/mod/modfile"
 )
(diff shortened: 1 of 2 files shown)


modules/test/crd.go
gibizer, line 1: Could you please extend the test coverage of this in https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/test/crd_test.go with the new path?

dciabrin: Done now.

gibizer (approved):
Thanks!

PR #397, opened 2023-11-20
[util] add generic MergeMaps func

stuggi:


Deydra71 (approved):
This will be very useful, thanks for it.

PR #405, opened 2023-11-29
Refactor of using gophercloud

mrkisaolamb:


modules/openstack/openstack.go
```diff
@@ -117,7 +114,6 @@ func NewOpenStack(
 	if err != nil {
 		return nil, err
 	}
-
```
gibizer, line 120: nit: whitespace damage. Or is it go vet again? :)

modules/openstack/openstack.go
```diff
@@ -61,17 +61,14 @@ type TLSConfig struct {
...
+func GetOpenStackProvider(
 	cfg AuthOpts,
-) (*OpenStack, error) {
-
+) (*gophercloud.ProviderClient, error) {
 	opts := gophercloud.AuthOptions{
 		IdentityEndpoint: cfg.AuthURL,
 		Username:         cfg.Username,
 		Password:         cfg.Password,
-		TenantName:       cfg.TenantName,
```
gibizer, line 74: What is the reason you drop Tenant?

modules/test/apis/server.go
```diff
@@ -18,11 +18,12 @@ import (
 	"net/http/httptest"
 
 	"github.com/go-logr/logr"
+	"github.com/gorilla/mux"
```
gibizer, line 21: can we make this in a separate commit? (can be in a same PR if you need it)

modules/openstack/openstack.go
mrkisaolamb: yes it was go vet

modules/openstack/openstack.go
mrkisaolamb: done

modules/test/apis/server.go
mrkisaolamb: done

modules/openstack/openstack.go
gibizer: then ignore me and chalk up one more to the "random acts of govet"

gibizer (approved):
As far as I see this is a clean refactor of `NewOpenStack` so the users of NewOpenStack will not break.

PR #411, opened 2023-12-07
[testhelper] add test helpers to assert volume and volumeMounts

stuggi:


Deydra71 (approved):

modules/common/test/helpers/volumes.go
```diff
@@ -0,0 +1,60 @@
...
+//
+//	th.AssertVolumeExists("name", []corev1.Volume{...})
+func (tc *TestHelper) AssertVolumeExists(name string, volumes []corev1.Volume) bool {
+	exist := false
+	for _, v := range volumes {
+		if v.Name == name {
+			exist = true
+		}
+	}
+	gomega.Expect(exist).To(gomega.BeTrue())
```
gibizer, line 34: I think this can be "simplified" (make it more gomegalike):
```golang
gomega.Expect(volumes).ContainElement(MatchField("Name", name))
```

stuggi: cool , thanks @gibizer will push an update

Deydra71 (approved):

modules/common/test/helpers/volumes.go
gibizer: thanks

PR #428, opened 2024-01-05
 TLS support for multiple endpoints

Deydra71:
This change adds the TLS support for multiple endpoints, and adds default mounting paths for the CA certificates bundle file, as well for the service certs and keys.


stuggi (commented):
/lgtm

modules/common/tls/tls.go
```diff
@@ -45,133 +54,417 @@ const (
...
+	}
+	if s.SecretName != "" {
+		volume = corev1.Volume{
+			Name: serviceID + "-tls-certs",
 			VolumeSource: corev1.VolumeSource{
 				Secret: &corev1.SecretVolumeSource{
-					SecretName:  t.Service.SecretName,
-					DefaultMode: ptr.To[int32](0440),
+					SecretName:  s.SecretName,
+					DefaultMode: ptr.To[int32](0400),
```
olliewalsh, line 390: If we want to use this without root (using FSGroup) does this need to be 0440?

stuggi: yes, maybe. for the services we did so far, non uses non root without kolla. if needed we can introduce a `*DefaultMode` to the Service type so that it could be customized if needed

olliewalsh: ack, will leave this for follow-up patch if/when it's required

olliewalsh (approved):
/lgtm

PR #432, opened 2024-01-08
[service/route] allow updating labels and services

stuggi:
with 18d0ea2c06a85f0d9a329c8f7ffbff41afeb0f1c we introduced merging annotations and labels. It was required as other components can add annotations to a created service:

~~~
Author: Martin Schuppert <mschuppert@redhat.com>
Date:   Wed Sep 27 09:06:26 2023 +0200

    Merge Annotations and Labels in CreateOrPatch

    Identified in OCP4.13 where metallb adds an annotation to the
    service when handling it, it resulted in reconcile loop since
    the current service CreateOrPatch did not expect that annotations
    will be altered outside the owner.

    This changes the different functions to merge Labels and Annotations
    in CreateOr Patch instead of assigning.
~~~

With the current order of how the labels on routes and services get created does not allow to change a value of an existing key as how the merge func works.

This changes the order that existing labels and annotations of services and routes get merged so that those from an existing object gets merged into the one either provided via overrides or calculated by the operators. This still allows other components to add entries, and the operators/users to updated existing ones.

### DIFF
--- a/modules/common/route/route.go
+++ b/modules/common/route/route.go
@@ -133,8 +133,8 @@ func (r *Route) CreateOrPatch(
 	}
 
 	op, err := controllerutil.CreateOrPatch(ctx, h.GetClient(), route, func() error {
-		route.Labels = util.MergeStringMaps(route.Labels, r.route.Labels)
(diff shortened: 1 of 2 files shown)


olliewalsh (approved):
/lgtm

PR #442, opened 2024-01-24
[renovate]Do not bump common module within lib-common

gibizer:
Both openstack and certmanager module depends on the common module within the same repo and both go.mod files has a replace line to use the in repo version of that dependency. Therefore it is not necessary to bump the pseudoversion of the common module in those go.mod files. So this PR disables those renovate bumps to avoid unnecessary CI churn.


stuggi (approved):
/lgtm

PR #444, opened 2024-01-29
service: expose PublishNotReadyAddresses for GenericService

dciabrin:
There are times where one wants to create a service that enables endpoints for pods which are not ready as per their healthchecks. For example, for bootstrapping a clustered service such as Galera or during the failover of an A/P service such as Redis.

The type corev1.Service addresses this use case by providing option PublishNotReadyAddresses. Expose this option in libcommon type GenericServiceDetails to build the proper Service object when using GenericService().


stuggi (approved):
/lgtm

PR #445, opened 2024-02-02
Drop DebugCommand const

gibizer:
The service and job Debug feature is removed from the operators in favor of using oc debug directly.

Implements: [OSPRH-4290](https://issues.redhat.com/browse/OSPRH-4290)


gibizer:
Keeping it draft until the service operator changes are removed

stuggi (approved):
/lgtm

PR #447, opened 2024-02-05
Add AddLabel()/AddAnnotation() for Route

elvgarrui:
I think this might need testing, but looking at similar PR (https://github.com/openstack-k8s-operators/lib-common/pull/418) I did not fully understand how I should create something similar to what we see there in "modules/common/service/service_test.go" for the Route struct.

### DIFF
--- a/modules/common/route/route.go
+++ b/modules/common/route/route.go
@@ -189,3 +189,39 @@ func (r *Route) Delete(
 
 	return nil
 }
+
+// GetLabels - returns labels of the route
+func (r *Route) GetLabels() map[string]string {
+	return r.route.Labels
+}
+
+// GetAnnotations - returns annotations of the route
+func (r *Route) GetAnnotations() map[string]string {
+	return r.route.Annotations
+}
+
+// AddAnnotation - Adds annotation and merges it with the current set
+func (r *Route) AddAnnotation(anno map[string]string) {
+	r.route.Annotations = util.MergeStringMaps(r.route.Annotations, anno)
+}
+
+// AddLabel - Adds label and merges it with the current set
+func (r *Route) AddLabel(label map[string]string) {
+	r.route.Labels = util.MergeStringMaps(r.route.Labels, label)
+}
+
+// AddAnnotation - Adds annotation and merges it with the current set
+func (r *OverrideSpec) AddAnnotation(anno map[string]string) {
+	if r.EmbeddedLabelsAnnotations == nil {
+		r.EmbeddedLabelsAnnotations = &EmbeddedLabelsAnnotations{}
+	}
+	r.Annotations = util.MergeMaps(r.Annotations, anno)
+}
+
+// AddLabel - Adds annotation and merges it with the current set
+func (r *OverrideSpec) AddLabel(label map[string]string) {
+	if r.EmbeddedLabelsAnnotations == nil {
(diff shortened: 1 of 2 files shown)


stuggi (commented):
right, this needs tests as in https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/service/service_test.go#L479 . specify a default Route starting with no annotation, then:
- add a annotation
- add another annotation should have 2
- override an annotation (first one remains)

you could just make a copy of service.TestOverrideSpecAddAnnotation and change it for the Route use case.

modules/common/route/route.go
```diff
@@ -189,3 +189,20 @@ func (r *Route) Delete(
 
 	return nil
 }
+
+// AddAnnotation - Adds annotation and merges it with the current set
+func (r *Route) addAnnotation(anno map[string]string) {
```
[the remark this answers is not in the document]
elvgarrui: Thanks for pointing it out, fixed

modules/common/route/route.go
```diff
@@ -189,3 +189,20 @@ func (r *Route) Delete(
...
+       if r.EmbeddedLabelsAnnotations == nil {
+               r.EmbeddedLabelsAnnotations = &EmbededLabelsAnnotations{}
+       }
+       r.Annotations = util.MergeMaps(r.Annotations, anno)
+}
+
+
+// AddLabel - Adds annotation and merges it with the current set
+func (r *Route) addLabel(label map[string]string) {
```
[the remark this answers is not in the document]
elvgarrui: Done!

modules/common/route/route.go
```diff
@@ -189,3 +189,24 @@ func (r *Route) Delete(
 
 	return nil
 }
+
+// AddAnnotation - Adds annotation and merges it with the current set
+func (r *Route) AddAnnotation(anno map[string]string) {
+	r.route.Annotations = util.MergeStringMaps(r.route.Annotations, anno)
+}
+
```
stuggi, line 197: I guess we also want an `AddLabel` for `Route`?
```
// AddLabel - Adds label and merges it with the current set
func (r *Route) AddLabel(label map[string]string) {
	r.route.Labels = util.MergeStringMaps(r.route.Labels, label)
}
```

modules/common/route/route.go
```diff
@@ -189,3 +189,24 @@ func (r *Route) Delete(
...
+// AddAnnotation - Adds annotation and merges it with the current set
+func (r *OverrideSpec) AddAnnotation(anno map[string]string) {
+	if r.EmbeddedLabelsAnnotations == nil {
+		r.EmbeddedLabelsAnnotations = &EmbeddedLabelsAnnotations{}
+	}
+	r.Annotations = util.MergeMaps(r.Annotations, anno)
+}
+
+// AddLabel - Adds annotation and merges it with the current set
```
stuggi, line 206: nit, `Adds annotation` -> `Adds label`

modules/common/route/route.go
elvgarrui: Done!

modules/common/route/route.go
elvgarrui: thx!

stuggi (approved):
/lgtm thanks!

PR #448, opened 2024-02-07
FIPS: Add common/ocp module

Akrog:
For FIPS support we need to be able to tell when an OCP cluster has been deployed in FIPS mode (it's a day one operation).

This patch adds the `IsFipsCluster` function that checks if FIPS is enabled or not.

The way to do that is checking the install-config YAML that is stored in the cluster-config-v1 Config Map.

If we find the fips key, and it has the true value, then it's deployed in FIPS mode, otherwise it isn't.

Related patch: https://github.com/openstack-k8s-operators/openstack-operator/pull/657

Jira: #OSPRH-4666


stuggi (approved):
/lgtm

PR #449, opened 2024-02-08
Fixing error log formatting

jpodivin:


modules/common/job/job.go
```diff
@@ -291,7 +291,7 @@ func GetJobWithName(
 	job := &batchv1.Job{}
 	err := h.GetClient().Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, job)
 	if err != nil {
-		h.GetLogger().Info("GetJobWithName %s err: %w", name, err)
+		h.GetLogger().Info(fmt.Sprintf("GetJobWithName %s err: %s", name, err))
 		return job, err
```
stuggi, line 295: I think its better to use `return job, fmt.Errorf("GetJobWithName %s err: %w", name, err)`

jpodivin: I suppose.

jpodivin: Done.

stuggi (approved):
/lgtm

PR #451, opened 2024-02-12
Fix missing string formatting and return msg as an error

cschwede:
String formatting is missing if job is not found. Instead of logging an error, return the error with the message. Jobs need to be restarted to be executed again, but might be deleted already because of their TTL. In this case this is not an error and should be handled properly by the calling code.


cschwede:
Just noticed https://github.com/openstack-k8s-operators/lib-common/pull/449 fixes this as well. Closing.

PR #462, opened 2024-02-22
[CronJob]Support updating config

gibizer:
Until now the update on the CronJob Spec fields was ignored. So
for example Schedule could not be upgraded.

This is fixed know.


stuggi (approved):
/lgtm

PR #465, opened 2024-02-26
Add annotation for specifying ingress target port.

vyzigold:
In telemetry-operator, we use the cluster-observability-operator to manage prometheus. COO creates a prometheus service with a target port name "web", which is different from the name of the service. COO doesn't create routes. This annotation lets us specify the intended target port name and use the openstack-operator to create a route for us after adding this annotation to the COO created service.


stuggi (approved):
/lgtm

PR #466, opened 2024-02-26
Fix DeleteInstance helper to check UID

gibizer:
When the helper deletes a resource it checks that it is not found any
more (i.e.:finalizeres are not blocking the deletion). But it checks by
resource name so if the resource is recreated in the meantime with the
same name then the helper will think the deletion did not happened.
This patch makes sure that only the original resource is deleted by
checking its UID.


mrkisaolamb (approved):
lgtm

stuggi (approved):
/lgtm

PR #467, opened 2024-02-26
Added function GetIssuerByLabels

vakwetu:


olliewalsh (approved):
/lgtm

modules/certmanager/test/functional/certmanager_test.go
```diff
@@ -50,6 +50,16 @@ var _ = Describe("certmanager module", func() {
 			g.Expect(issuer.Spec.SelfSigned).NotTo(BeNil())
 			g.Expect(issuer.ObjectMeta.Labels["f"]).To(Equal("l"))
 		}, timeout, interval).Should(Succeed())
+		Eventually(func(g Gomega) {
```
stuggi, line 53: could also add tests for the `==0` and `>1`  case. we can do that in a follow up

stuggi (approved):
/lgtm

modules/certmanager/test/functional/certmanager_test.go
stuggi: https://github.com/openstack-k8s-operators/lib-common/pull/468

PR #468, opened 2024-02-27
[certmanager] enhance GetIssuerByLabels tests

stuggi:
cover tests where no issuer is found using label and multiple issuers got returned by the passed in label.


olliewalsh (approved):
/lgtm

PR #473, opened 2024-03-09
Fix dup job runs - hash was calculated on wrong root object

olliewalsh:
Spec.Template.Metadata.Labels being set by the job controller were triggering a 2nd run as this was included in the hash.
Hash should only consider Spec.Template.Spec.


olliewalsh:
E.g dup job run:
```$ diff -u foo2.yaml foo.yaml 
--- foo2.yaml	2024-03-09 01:08:22.463746059 +0000
+++ foo.yaml	2024-03-09 01:05:39.882073342 +0000
@@ -5,7 +5,7 @@
     batch.kubernetes.io/job-tracking: ""
     hash: n59fhbbh7h8dh674h66fh4h59ch545h584h54dhd4h6h689h665h648h559h587h56bh67h5ffh5bbhbbh75h5d6hfchfh594h544hd6h5d9h58dq
     k8s.v1.cni.cncf.io/networks: '[{"name":"ctlplane","namespace":"openstack","interface":"ctlplane"}]'
-  creationTimestamp: "2024-03-09T01:08:13Z"
+  creationTimestamp: "2024-03-09T01:05:14Z"
   generation: 1
   labels:
     app: openstackansibleee
@@ -23,8 +23,8 @@
     kind: OpenStackAnsibleEE
     name: bootstrap-openstack-edpm
     uid: 8acf94f5-fa2e-4d52-b38f-f2bf5d7c75b2
-  resourceVersion: "65060"
-  uid: 3de3bfba-0431-44f4-bdb6-ffbb8d1cfd97
+  resourceVersion: "62296"
+  uid: b751000e-e78a-4941-b43f-bca610197c40
 spec:
   backoffLimit: 6
   completionMode: NonIndexed
@@ -32,7 +32,7 @@
   parallelism: 1
   selector:
     matchLabels:
-      batch.kubernetes.io/controller-uid: 3de3bfba-0431-44f4-bdb6-ffbb8d1cfd97
+      batch.kubernetes.io/controller-uid: b751000e-e78a-4941-b43f-bca610197c40
   suspend: false
   template:
     metadata:
@@ -41,10 +41,10 @@
       creationTimestamp: null
       labels:
         app: openstackansibleee
-        batch.kubernetes.io/controller-uid: 3de3bfba-0431-44f4-bdb6-ffbb8d1cfd97
+        batch.kubernetes.io/controller-uid: b751000e-e78a-4941-b43f-bca610197c40
         batch.kubernetes.io/job-name: bootstrap-openstack-edpm
         bootstrap: 3ed5d1cd-78d1-416f-ac1b-24489af01f3e
-        controller-uid: 3de3bfba-0431-44f4-bdb6-ffbb8d1cfd97
+        controller-uid: b751000e-e78a-4941-b43f-bca610197c40
         job-name: bootstrap-openstack-edpm
         openstackansibleee_cr: bootstrap-openstack-edpm
         osaee: "true"
@@ -121,5 +121,5 @@
 status:
   active: 1
   ready: 1
-  startTime: "2024-03-09T01:08:13Z"
+  startTime: "2024-03-09T01:05:14Z"
   uncountedTerminatedPods: {}
```

stuggi:
e.g. for edpm job do we want the job to rerun if e.g. the annotations for network attachments change? then we might want to add the annotations to the hash?

modules/common/job/job.go
```diff
@@ -132,7 +132,7 @@ func (j *Job) DoJob(
 	// Fields outside of the PodTemplate, like TTL does not define what to run
 	// just how to run it. So changing such field should not trigger the re-run
 	// of the Job
-	j.hash, err = util.ObjectHash(j.expectedJob.Spec.Template)
+	j.hash, err = util.ObjectHash(j.expectedJob.Spec.Template.Spec)
```
gibizer, line 135: I would extend the comment above and indicate that we ignore the pod metadata as it is also not relevant to what to run.

gibizer (commented):
I agree not to rerun the Job if the pod metadata is changing due to automatic label management by k8s. If we have labels or annotations that are significant to the definition of  "what" to run, then we need a some special hashing logic for that data.

gibizer:
> e.g. for edpm job do we want the job to rerun if e.g. the annotations for network attachments change? then we might want to add the annotations to the hash?

Does the network attachment changes what to run? Could it be that a changed network attachment on and edpm job results in the job run against a different set of computes?

stuggi:
> > e.g. for edpm job do we want the job to rerun if e.g. the annotations for network attachments change? then we might want to add the annotations to the hash?
> 
> Does the network attachment changes what to run? Could it be that a changed network attachment on and edpm job results in the job run against a different set of computes?

Not what to run and against which computes, but how the pod gets connected and can reach the edpm nodes. like if one realized the network attachments were wrong and updates them, it would have to rerun the job, or am I mixing something ?

gibizer:
> > > e.g. for edpm job do we want the job to rerun if e.g. the annotations for network attachments change? then we might want to add the annotations to the hash?
> > 
> > 
> > Does the network attachment changes what to run? Could it be that a changed network attachment on and edpm job results in the job run against a different set of computes?
> 
> Not what to run and against which computes, but how the pod gets connected and can reach the edpm nodes. like if one realized the network attachments were wrong and updates them, it would have to rerun the job, or am I mixing something ?

If the network attachment was wrong then I assume the job is failed and therefore the hash is not supposed to be saved in the CRD status.  So when the attachment is corrected the job will be restarted as the empty hash stored in the CRD  will be different from the actual job hash.
Is there a scenario where a wrong network attachement result in a successful job run?

stuggi:
> > > > e.g. for edpm job do we want the job to rerun if e.g. the annotations for network attachments change? then we might want to add the annotations to the hash?
> > > 
> > > 
> > > Does the network attachment changes what to run? Could it be that a changed network attachment on and edpm job results in the job run against a different set of computes?
> > 
> > 
> > Not what to run and against which computes, but how the pod gets connected and can reach the edpm nodes. like if one realized the network attachments were wrong and updates them, it would have to rerun the job, or am I mixing something ?
> 
> If the network attachment was wrong then I assume the job is failed and therefore the hash is not supposed to be saved in the CRD status. So when the attachment is corrected the job will be restarted as the empty hash stored in the CRD will be different from the actual job hash. Is there a scenario where a wrong network attachement result in a successful job run?

only can think of a scenario right now where routing would also work via the pod network. but then the content of the job itself won't change, so we'd not have to rerun it. I am not sure if we use annotations as another information for jobs. Its probably safe to exclude it (at least for now).

gibizer (approved):

PR #480, opened 2024-03-18
Restoring TargetReference defaults

jpodivin:


jpodivin:
> why do you think we should to this change? as you can see from the comment at the top of the struct it was intentional set to optional with omitempty
> 
> ```
> // Copy of RouteSpec in https://github.com/openshift/api/blob/master/route/v1/types.go,
> // parameters set to be optional, have omitempty, and no default.
> ```

Yes, but it doesn't say why.

Omitempty doesn't really provide any value in this case. It only makes sense when you are dealing with struct pointers, or other types with meaningful empty value. Structs like `TargetReference` don't have that. So the field remains.

Consider following:
```
package main

import (
	"encoding/json"
	"fmt"
)

type MyStruct struct {
	ValueOfImport int `json:",omitempty"`
}

type SuperStruct struct {
	IntegerValue          int      `json:",omitempty"` // this makes sense
	NestedStruct          MyStruct `json:",omitempty"` // this doesn't
	PointerToIntegerValue *int     `json:",omitempty"` // neither does this
}

type AlternativeStruct struct {
	IntegerValue          int      `json:",omitempty"`
	NestedStruct          MyStruct `json:""`
	PointerToIntegerValue *int     `json:",omitempty"`
}

func main() {
	d := SuperStruct{
		IntegerValue: 0,
	}
	b, _ := json.Marshal(d)
	fmt.Println(string(b))
	x := AlternativeStruct{
		IntegerValue: 0,
	}
	b, _ = json.Marshal(x)
	fmt.Println(string(b))
}
```

When you marshall the `SuperStruct` the `IntegerValue` is set to `0` which is equivalent for empty of that type. Now the field won't be in the string representation of that object.

But the `NestedStruct` will be there. Even though I didn't supply any value for it.  Compare with the output for `AlternativeStruct`.

This is one of the reasons why some other projects have been pruning omitempty usage [0]. Because it's not that useful.

It doesn't hurt us. Not actively, but why should we deviate from what was already implemented for no benefit at all?

As for the `Weight`. I don't see how allowing potentially `nil` weight helps us.

[0]  https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/1729

stuggi:
> > why do you think we should to this change? as you can see from the comment at the top of the struct it was intentional set to optional with omitempty
> > ```
> > // Copy of RouteSpec in https://github.com/openshift/api/blob/master/route/v1/types.go,
> > // parameters set to be optional, have omitempty, and no default.
> > ```
> 
> Yes, but it doesn't say why.
> 
> Omitempty doesn't really provide any value in this case. It only makes sense when you are dealing with struct pointers, or other types with meaningful empty value. Structs like `TargetReference` don't have that. So the field remains.
> 
> Consider following:
> 
> ```
> package main
> 
> import (
> 	"encoding/json"
> 	"fmt"
> )
> 
> type MyStruct struct {
> 	ValueOfImport int `json:",omitempty"`
> }
> 
> type SuperStruct struct {
> 	IntegerValue          int      `json:",omitempty"` // this makes sense
> 	NestedStruct          MyStruct `json:",omitempty"` // this doesn't
> 	PointerToIntegerValue *int     `json:",omitempty"` 
> }
> 
> type AlternativeStruct struct {
> 	IntegerValue          int      `json:",omitempty"`
> 	NestedStruct          MyStruct `json:""`
> 	PointerToIntegerValue *int     `json:",omitempty"`
> }
> 
> func main() {
> 	d := SuperStruct{
> 		IntegerValue: 0,
> 	}
> 	b, _ := json.Marshal(d)
> 	fmt.Println(string(b))
> 	x := AlternativeStruct{
> 		IntegerValue: 0,
> 	}
> 	b, _ = json.Marshal(x)
> 	fmt.Println(string(b))
> }
> ```
> 
> When you marshall the `SuperStruct` the `IntegerValue` is set to `0` which is equivalent for empty of that type. Now the field won't be in the string representation of that object.
> 
> But the `NestedStruct` will be there. Even though I didn't supply any value for it. Compare with the output for `AlternativeStruct`.
> 
> This is one of the reasons why some other projects have been pruning omitempty usage [0]. Because it's not that useful.
> 
> It doesn't hurt us. Not actively, but why should we deviate from what was already implemented for no benefit at all?
> 
> As for the `Weight`. I don't see how allowing potentially `nil` weight helps us.

This is not true and required. Omitempty does make sense on ptr types to not have them included when nil. Since this OverrideSpec is used to be able to override specific values (not all of them), e.g. from your example, PointerToIntegerValue must not be there if not provided, so omitempty on Weight must be there. Again, this is an OverrideSpec, where all of them must be optional and not there if not specificed. To reflect this omitEmpty was added. Lets keep it simple and just keep them.

bogdando (commented):
> > > why do you think we should to this change? as you can see from the comment at the top of the struct it was intentional set to optional with omitempty
> > ```
> > // Copy of RouteSpec in https://github.com/openshift/api/blob/master/route/v1/types.go,
> > // parameters set to be optional, have omitempty, and no default.
> > ```
> 
> Yes, but it doesn't say why.
> 
> Omitempty doesn't really provide any value in this case. It only makes sense when you are dealing with struct pointers, or other types with meaningful empty value. Structs like `TargetReference` don't have that. So the field remains.
> 
> Consider following:
> 
> ```
> package main
> 
> import (
> 	"encoding/json"
> 	"fmt"
> )
> 
> type MyStruct struct {
> 	ValueOfImport int `json:",omitempty"`
> }
> 
> type SuperStruct struct {
> 	IntegerValue          int      `json:",omitempty"` // this makes sense
> 	NestedStruct          MyStruct `json:",omitempty"` // this doesn't
> 	PointerToIntegerValue *int     `json:",omitempty"` 
> }
> 
> type AlternativeStruct struct {
> 	IntegerValue          int      `json:",omitempty"`
> 	NestedStruct          MyStruct `json:""`
> 	PointerToIntegerValue *int     `json:",omitempty"`
> }
> 
> func main() {
> 	d := SuperStruct{
> 		IntegerValue: 0,
> 	}
> 	b, _ := json.Marshal(d)
> 	fmt.Println(string(b))
> 	x := AlternativeStruct{
> 		IntegerValue: 0,
> 	}
> 	b, _ = json.Marshal(x)
> 	fmt.Println(string(b))
> }
> ```
> 
> When you marshall the `SuperStruct` the `IntegerValue` is set to `0` which is equivalent for empty of that type. Now the field won't be in the string representation of that object.
> 
> But the `NestedStruct` will be there. Even though I didn't supply any value for it. Compare with the output for `AlternativeStruct`.
> 
> This is one of the reasons why some other projects have been pruning omitempty usage [0]. Because it's not that useful.
> 
> It doesn't hurt us. Not actively, but why should we deviate from what was already implemented for no benefit at all?
> 
> As for the `Weight`. I don't see how allowing potentially `nil` weight helps us.

> This is not true and required. Omitempty does make sense on ptr types to not have them included when nil. Since this OverrideSpec is used to be able to override specific values (not all of them), e.g. from your example, PointerToIntegerValue must not be there if not provided, so omitempty on Weight must be there. Again, this is an OverrideSpec, where all of them must be optional and not there if not specificed. To reflect this omitEmpty was added. Lets keep it simple and just keep them.

From the linked above CAPI issue "Omitempty only works for pointers to structs, not structs themselves". Perhaps , this change still makes sense for the omitempty structure, if not for the pointer to int?

bogdando (commented):
@mdbooth WDYT?

stuggi:
I am closing this for now. We do not want to do this change now before GA

PR #483, opened 2024-03-25
Turn conditions.hasSameState function public

fmount:
In order to consume `hasSameState` function in the service operators, we need to export this function from `common/condition` module.
This patch just turns it into an exported function that we can consume from the service operators.


fmount:
I need this first: https://github.com/openstack-k8s-operators/lib-common/pull/484

abays (approved):
/lgtm

mrkisaolamb (approved):
lgtm

PR #485, opened 2024-03-25
conditions: add RestoreLastTransitionTimes function

ASBishop:
The RestoreLastTransitionTimes function facilitates preserving a condition's LastTransitionTime if a controller reinitializes its conditions at the start of every reconcilation. Controllers can use the following sequence in their reconciliation code:

- Save a copy of the current conditions
- Reinitialize conditions to Unknown
- Reconcile, which updates the conditions to their new state
- Call RestoreLastTransitionTimes

For conditions whose state doesn't change, the LastTransitionTime is set back to the time recorded in the saved copy. This prevents the LastTransitionTime from changing unless the condition itself actually changes state.

### DIFF
--- a/modules/common/condition/funcs.go
+++ b/modules/common/condition/funcs.go
@@ -380,6 +380,17 @@ func (conditions *Conditions) Mirror(t Type) *Condition {
 	return mirrorCondition
 }
 
+// RestoreLastTransitionTimes - Updates each condition's LastTransitionTime when its state
+// matches the one in a list of "saved" conditions.
+func RestoreLastTransitionTimes(conditions *Conditions, savedConditions Conditions) {
+	for idx, cond := range *conditions {
+		savedCond := savedConditions.Get(cond.Type)
+		if savedCond != nil && HasSameState(&cond, savedCond) {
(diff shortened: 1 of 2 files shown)


mrkisaolamb (approved):
lgtm

fmount (approved):
Thank you @ASBishop for working on this!

modules/common/condition/funcs.go
```diff
@@ -380,6 +380,17 @@ func (conditions *Conditions) Mirror(t Type) *Condition {
 	return mirrorCondition
 }
 
+// RestoreLastTransitionTimes - Updates each condition's LastTransitionTime when its state
+// matches the one in a list of "saved" conditions.
+func RestoreLastTransitionTimes(conditions *Conditions, savedConditions Conditions) {
+	for idx, cond := range *conditions {
+		savedCond := savedConditions.Get(cond.Type)
+		if savedCond != nil && HasSameState(&cond, savedCond) {
```
gibizer, line 388: OK, so we went with the more fine grained transition detection. I.e. if only the message changes we bump the transition time. That is OK to me.

PR #487, opened 2024-04-02
[Job] Do not silent delete job when preserve is set to true

fao89:
It can make must-gather blind to failed jobs

### DIFF
--- a/modules/common/job/job.go
+++ b/modules/common/job/job.go
@@ -258,14 +258,20 @@ func (j *Job) waitOnJob(
 		return ctrl.Result{}, nil
 	} else if j.actualJob.Status.Failed > 0 {
 		if existingJobHash != j.hash {
-			h.GetLogger().Info(
-				"The hash of the job changed but the previous failed job still exists. " +
-					"Deleting old job and requeueing.")
-			err := DeleteJob(ctx, h, j.actualJob.Name, j.actualJob.Namespace)
-			if err != nil {
-				return ctrl.Result{}, err
+			if j.preserve {
+				h.GetLogger().Info(
+					"The hash of the job changed and the previous failed job still exists. " +
+						"Consider deleting the old job and requeue.")
+			} else {
+				h.GetLogger().Info(
+					"The hash of the job changed but the previous failed job still exists. " +
+						"Deleting old job and requeueing.")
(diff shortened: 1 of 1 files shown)


gibizer (commented):
could you please add a test case for this.

modules/common/job/job.go
```diff
@@ -258,14 +258,20 @@ func (j *Job) waitOnJob(
...
-			h.GetLogger().Info(
-				"The hash of the job changed but the previous failed job still exists. " +
-					"Deleting old job and requeueing.")
-			err := DeleteJob(ctx, h, j.actualJob.Name, j.actualJob.Namespace)
-			if err != nil {
-				return ctrl.Result{}, err
+			if j.preserve {
+				h.GetLogger().Info(
+					"The hash of the job changed and the previous failed job still exists. " +
+						"Consider deleting the old job and requeue.")
```
gibizer, line 264: This is problematic. This means if a job failed  then the operator requesting a different config for the job will not get applied but simply ignored.

gibizer:
> It can make must-gather blind to failed jobs

Could you be a bit more specific what leads to the missing jobs? If the job's input did not change then the job hash should not change either and therefore I don't expect that a new job is created leading to the old job's deletion.

modules/common/job/job.go
fao89: I agree, what I really wanted was a way to signal to the job owner that the job was deleted

fao89:
> > It can make must-gather blind to failed jobs
> 
> Could you be a bit more specific what leads to the missing jobs? If the job's input did not change then the job hash should not change either and therefore I don't expect that a new job is created leading to the old job's deletion.

my motivation came from: https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/341
AEE was wrongly changing hashes when handling a map field, it was hiding the failures on CI which led to people bumping the timeout. It took a lot of time to find what was hiding the AEE failures.

modules/common/job/job.go
fao89: OTOH, deleting a job means deleting pod logs, I wonder if could keep the failed logs somehow

fao89:
actually, AEE is the right place to make changes

PR #488, opened 2024-04-05
Add .ObservedGeneration to statefulset and deployments test module

fmount:
In order to be able to rollout the full `.ObservedGeneration` pattern we need `envTests` to pass if this is added as a condition to mark a `Deployment` or `StatefulSet` `Ready`. 
This change introduces this parameter for the test module so we can take it into account in the envTests when the `DeploymentReadyCondition` is computed.


gibizer (approved):
Thanks!

PR #490, opened 2024-04-09
[certmanager] add subject and common name to certrequest

stuggi:
Allow setting the subject of a certificate via the certrequest.


olliewalsh (approved):
/lgtm

PR #494, opened 2024-04-19
Add SimulateDeploymentAnyNumberReplicaReady

mrkisaolamb:
Add method that will simulate any number of replicas ready

### DIFF
--- a/modules/common/test/helpers/deployment.go
+++ b/modules/common/test/helpers/deployment.go
@@ -50,6 +50,24 @@ func (tc *TestHelper) GetDeployment(name types.NamespacedName) *appsv1.Deploymen
 	return deployment
 }
 
+// SimulateDeploymentAnyNumberReplicaReady function retrieves the Deployment resource and
+// simulate that replicas are ready
+// Example usage:
+//
+//	th.SimulateDeploymentAnyNumberReplicaReady(ironicNames.INAName, 0)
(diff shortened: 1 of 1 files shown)


modules/common/test/helpers/deployment.go
```diff
@@ -50,6 +50,24 @@ func (tc *TestHelper) GetDeployment(name types.NamespacedName) *appsv1.Deploymen
...
+// SimulateDeploymentAnyNumberReplicaReady function retrieves the Deployment resource and
+// simulate that replicas are ready
+// Example usage:
+//
+//	th.SimulateDeploymentAnyNumberReplicaReady(ironicNames.INAName, 0)
+func (tc *TestHelper) SimulateDeploymentAnyNumberReplicaReady(name types.NamespacedName, replica int32) {
+	gomega.Eventually(func(g gomega.Gomega) {
+		deployment := tc.GetDeployment(name)
+
+		deployment.Status.Replicas = replica
```
gibizer, line 62: could you please also change SimulateDeploymentReplicaReady() to set Status.Replicas and ReadyReplicas based on deployment.Replicas to allow the positive test cases to still use the SimulateDeploymentReplicaReady call an only the negative cases need to use the new SimulateDeploymentAnyNumberReplicaReady

mrkisaolamb: done, should be ok also with other operators

gibizer (approved):
Looks good.

PR #496, opened 2024-04-23
Do not add network annotations in SimulateDeploymentReadyWithPods

umago:
This patch modifies the SimulateDeploymentReadyWithPods test function to not add network annotations when an empty map is passed to it.

This patch is required because we are removing networkAttachments for ovnNorthd (it's not needed, see [0]) and tests needs to be updated.

[0] https://github.com/openstack-k8s-operators/ovn-operator/pull/269

### DIFF
--- a/modules/common/test/helpers/deployment.go
+++ b/modules/common/test/helpers/deployment.go
@@ -117,20 +117,22 @@ func (tc *TestHelper) SimulateDeploymentReadyWithPods(name types.NamespacedName,
 		for i := range pod.Spec.InitContainers {
 			pod.Spec.InitContainers[i].VolumeMounts = []corev1.VolumeMount{}
 		}
-
-		var netStatus []networkv1.NetworkStatus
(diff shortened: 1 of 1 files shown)


modules/common/test/helpers/deployment.go
```diff
@@ -117,20 +117,22 @@ func (tc *TestHelper) SimulateDeploymentReadyWithPods(name types.NamespacedName,
...
+					netStatus,
+					networkv1.NetworkStatus{
+						Name: network,
+						IPs:  IPs,
+					},
+				)
+			}
+			netStatusAnnotation, err := json.Marshal(netStatus)
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			pod.Annotations[networkv1.NetworkStatusAnnot] = string(netStatusAnnotation)
```
booxter, line 134: there's still a bug here where the code may attempt to set into the map that is nil. But I see with this, northd PR is not affected, so it's good direction.

PR #502, opened 2024-04-30
In SimulateReady methods we should set correct replica number

mrkisaolamb:


mrkisaolamb:
We need to wait with this merge until after the end of the beta freeze

gibizer (commented):
looks good

gibizer (approved):

PR #503, opened 2024-05-03
Add UnmarshalYAML to the ansible inventory module

fao89:


modules/ansible/inventory_test.go
```diff
@@ -83,3 +83,67 @@ func TestInventoryMarshallNestedChildren(t *testing.T) {
...
+	invObj, _ := UnmarshalYAML([]byte(inventory))
+
+	if invObj.Groups["all"].Hosts["testing"].Vars["ansible_host"] != expected.Groups["all"].Hosts["testing"].Vars["ansible_host"] {
+		t.Logf("error should be %+v but got %+v", expected.Groups["all"].Hosts, invObj.Groups["all"].Hosts)
+		t.Fail()
+	}
+	if invObj.Groups["all"].Children["child_test"].Hosts["child_testing"].Vars["ansible_host"] != expected.Groups["all"].Children["child_test"].Hosts["child_testing"].Vars["ansible_host"] {
+		t.Logf("error should be %+v but got %+v", expected.Groups["all"].Children, invObj.Groups["all"].Children)
+		t.Fail()
+	}
```
fao89, line 115: couldn't use [reflect.DeepEqual](https://golang.org/pkg/reflect/#DeepEqual) because:
-  it couldn't unmarshal the host/group names
- Unmarshall parses empty values as nil, while inventory [creates the struct](https://github.com/openstack-k8s-operators/lib-common/blob/09a614506ca6251c24474bba8fccfef224b72c95/modules/ansible/inventory.go#L54-L56) with [non-nil, empty slice ](https://stackoverflow.com/a/44305910)

slagle (approved):

PR #504, opened 2024-05-06
Add more details to the Job error message

fao89:


modules/common/condition/conditions.go
```diff
@@ -103,6 +103,10 @@ const (
 	// This is a warning because the reconciler will retry deletion.
 	ErrorReason = "Error"
 
+	// ErrorReason (Severity=Error) documents a condition not in Status=True because the underlying object reached the backoff limit.
```
stuggi, line 106: `ErrorReason` -> `JobReasonBackoffLimitExceeded`

modules/common/job/job.go
```diff
@@ -207,6 +208,22 @@ func (j *Job) GetHash() string {
 	return j.hash
 }
 
+// GetTotalAttempts func
+func (j *Job) GetTotalAttempts() int64 {
+	if j.actualJob == nil {
+		return 0
+	}
+	return int64(j.actualJob.Status.Failed)
```
stuggi, line 216: should this be `Status.Failed` + `Status.Succeeded` ? like the total attempts of a job would be e.g.  if you have 2 failed and last succeeded so in total you'd have 2 + 1.  Or should rename to `GetTotalFailedAttempts` ?

fao89: I wanted to register the retries, so renamed to `GetTotalFailedAttempts`

jpodivin (changes requested):

modules/common/job/job.go
```diff
@@ -207,6 +208,22 @@ func (j *Job) GetHash() string {
 	return j.hash
 }
 
+// GetTotalFailedAttempts func
+func (j *Job) GetTotalFailedAttempts() int64 {
+	if j.actualJob == nil {
+		return 0
+	}
+	return int64(j.actualJob.Status.Failed)
```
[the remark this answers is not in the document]
fao89: for passing it to `strconv.FormatInt`, but you are right, changing the `fmt` formatting is better

jpodivin (commented):
+2 From me. If nobody else has any suggestions we can merge this.

stuggi:
we are in merge freeze. lets not merge things until then

jpodivin:
> can we unhold and merge now?

I think we can.

jpodivin (approved):

PR #505, opened 2024-05-07
Add service Endpoint Validate() method

stuggi:
This method can e.g. be used in webhooks to validate if the configuration in the CR is correct.

### DIFF
--- a/modules/common/service/types.go
+++ b/modules/common/service/types.go
@@ -19,9 +19,13 @@ limitations under the License.
 package service
 
 import (
+	"fmt"
 	"time"
 
+	"golang.org/x/exp/slices"
+
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
 // Endpoint - typedef to enumerate Endpoint verbs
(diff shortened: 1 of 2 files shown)


modules/common/service/types.go
```diff
@@ -60,6 +64,30 @@ func (e *Endpoint) String() string {
...
+	return nil
+}
+
+// ValidateRoutedOverrides - validates map of RoutedOverrideSpec
+func ValidateRoutedOverrides(basePath *field.Path, overrides map[Endpoint]RoutedOverrideSpec) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	// validate the service override key is valid
+	for k := range overrides {
+		path := basePath.Child("override").Child("service").Key(k.String())
```
gibizer, line 81: this now assume that all the user of this validation stores the field under .Override.Service . I think is something only the caller knows so we should move this there and do not make assumption about the location of the field in the CR here.

stuggi: right, while we have this in sync across the operators,  we could make it more generic bu passing this in. I'll update it. thanks for the pointer

stuggi: done

abays (approved):
/lgtm

PR #507, opened 2024-05-20
Add method too ensure single secret

mrkisaolamb:


modules/common/secret/secret.go
```diff
@@ -445,3 +446,74 @@ func VerifySecret(
...
 }
+
+type ConditionUpdater interface {
+	Set(c *condition.Condition)
+	MarkTrue(t condition.Type, messageFormat string, messageArgs ...interface{})
+}
+
+// ensureSecret - ensures that the Secret object exists and the expected fields
+// are in the Secret. It returns a hash of the values of the expected fields.
+func EnsureSecret(
```
stuggi, line 457: thats the same functionality what  VerifySecret provide, no? check secret exist, contains expected fields and return the hash. it is just the condition handling which it provides, which I think could be done in the caller.

gibizer: Another difference is that EnsureSecret is setting the InputReady condition. I'm wondering if we could pull out the common code to a private function to avoid the duplication and keep this two public methods for now. Later on we can move all the callers of VerifySecret to EnsureSecret and remove VerifySecret but that requires individual changes in each caller due to the signature change.

stuggi: right missed the return of the secret. setting the condition is a big difference from what is usually done in all the other current lib-common code. from a quick grep, only the `ReconcileRbac` is setting conditions within the lib-common code (except of the conditions handling stuff itself). Is this a direction we want to move to? I am not sure. Right now the caller is setting the condition depending on the shared function return. if we keep the condition handling in the caller, VerifySecret only misses to return the secret, which we could add, or add a new func to deprecate the current one if we want to name it EnsureSecret.

If we want to also go with condition handling in lib-common funcs, I think the ConditionUpdater should not be part of the secret pkg, it should be in condition.

bogdando:
is this still needed @mrkisaolamb ?

PR #516, opened 2024-06-05
Add AdditionalSubjectNames and the related function to tls module

fmount:
`AdditionalSubjectNames` represents the key to access a comma separated list of `additionalSubjectNames` that should be passed to the `CertificateRequest`.

### DIFF
--- a/modules/common/tls/tls.go
+++ b/modules/common/tls/tls.go
@@ -67,6 +67,13 @@ const (
 
 	// TLSHashName - Name of the hash of hashes of all cert resources used to identify a change
 	TLSHashName = "certs"
+
+	// AdditionalSubjectNamesKey - Comma separated list of additionalSubjectNames
+	// that should be passed to the CertificateRequest
+	AdditionalSubjectNamesKey = "additionalSubjectNames"
+
+	// DefaultClusterInternalDomain - cluster internal dns domain
+	DefaultClusterInternalDomain = "cluster.local"
 )
 
 // SimpleService defines the observed state of TLS for a single service
--- a/modules/common/util/map.go
+++ b/modules/common/util/map.go
@@ -16,7 +16,10 @@ limitations under the License.
 
 package util
 
-import "sort"
+import (
+	"sort"
+	"strings"
+)
 
 // InitMap - Inititialise a map to an empty map if it is nil.
 func InitMap(m *map[string]string) {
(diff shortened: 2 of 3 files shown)


modules/common/tls/tls.go
```diff
@@ -67,6 +68,13 @@ const (
 
 	// TLSHashName - Name of the hash of hashes of all cert resources used to identify a change
 	TLSHashName = "certs"
+
+	// AdditionalSubjectNames - Comma separated list of additionalSubjectNames
+	// that should be passed to the CertificateRequest
+	AdditionalSubjectNames = "additionalSubjectNames"
```
stuggi, line 74: to align with the above, could you name it `AdditionalSubjectNamesKey`

modules/common/tls/tls.go
```diff
@@ -435,3 +443,17 @@ func (c *Ca) CreateVolume() corev1.Volume {
 
 	return volume
 }
+
+// GetAdditionalSubjectNames represents a utility function that takes a comma
+// separated list of additionalSubjectNames and returns a list of string that
+// can be easily handled by openstack-operator when a CertificateRequest is
+// generated.
+func GetAdditionalSubjectNames(svc corev1.Service) []string {
```
stuggi, line 451: how about we add a generic func to https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/util/map.go where we pass in a map and the key to lookup from the hash . then the bellow and return the list. I think it might be  useful for other cases

modules/common/tls/tls.go
fmount: ack

modules/common/tls/tls.go
fmount: ok, it's something I had in my initial version, but then I thought it was better to have a `Service` as input to hide details at the openstack-operator level.
I'll change this function to be something like:

```golang
// GetStringsFromMap - It returns a list of string based on a comma separated
// list of strings. This is usually set to normalize annotations
func GetStringsFromMap(hash map[string]string, key string) []string {
	res := []string{}
	if h, ok := hash[key]; ok {
		if h != "" {
			res = strings.Split(h, ",")
		}
	}
	return res
}
```

and add some unit tests to it.
Let me know if it's not something that matches with what you had in mind.

modules/common/tls/tls.go
```diff
@@ -67,6 +67,13 @@ const (
 
 	// TLSHashName - Name of the hash of hashes of all cert resources used to identify a change
 	TLSHashName = "certs"
+
+	// AdditionalSubjectNamesKey - Comma separated list of additionalSubjectNames
+	// that should be passed to the CertificateRequest
+	AdditionalSubjectNamesKey = "additionalSubjectNames"
```
fmount, line 73: is this consistent with the common approach?

modules/common/util/map.go
```diff
@@ -99,3 +102,20 @@ func MergeMaps[K comparable, V any](baseMap map[K]V, extraMaps ...map[K]V) map[K
...
 }
+
+// GetStringListFromMap - It returns a list of strings based on a comma
+// separated list assigned to the map key. This is usually invoked to normalize
+// annotation fields where a list of items is expressed with a comma separated
+// list of strings.
+// Example:
+// input: in["additionalSubjectNamesKey"] = "foo.bar,bar.svc,*.foo.bar"
+// output: [foo.bar bar.svc *.foo.bar]
+func GetStringListFromMap(in map[string]string, key string) []string {
```
fmount, line 113: This seems what we discussed in the previous iteration. Let's agree on this and then I update the other patches.

stuggi: yes, thats what I had in mind of doing a common func, instead of adding only for this particular use case func.

modules/common/tls/tls.go
stuggi: yes lgtm

stuggi (approved):
/lgtm

PR #517, opened 2024-06-12
Skip encodePath based on runtime GOOS

bshephar:
Qualify of life improvement that skips path encoding on systems where it's likely to cause test failures - Macs, it impacts Macs.

Look, I appreciate that this is a selfish change and only impacts a small subset of engineers. But it's annoying to edit this file and get tests passing locally each time.

### DIFF
--- a/modules/test/crd.go
+++ b/modules/test/crd.go
@@ -18,6 +18,7 @@ import (
 	"go/build"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"unicode/utf8"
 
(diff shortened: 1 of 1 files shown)


gibizer (approved):
I had no time to get back to this and look at how to make the encoding apply only to the go package path and not the whole absolute path. I suspect this change can only cause troubles for mac users with caputal letter in their github user. I dont know if we have one or not l
 But I know @bshephar is affected without this change. Sonlets merge it to unblock him.

karelyatin (commented):
/lgtm

PR #520, opened 2024-06-21
[svc] return error if no LoadBalancer IP

stuggi:
If there is an issue with the LoadBalancer service in the cluster, like the MetalLB instance is down, or the requested IP using the metallb.universe.tf/loadBalancerIPs annotation is outside the configured range of the IPAddressPool, the k8s LoadBalancert service External-IP is stuck in pending. This would not allow dataplane service to connect to the service.
This change returns an error in this case to reflect that there is an issue with the k8s service, which needs to be fixed and the consumer of the func can reconcile to wait for a LoadBalancer IP to be available.


stuggi:
need to update the test

gibizer (approved):
looks good.

PR #522, opened 2024-06-24
Add webhook package, to be used for common webhook checks

stuggi:
Adds webhook package, intended to be a collection of webhook funcs, which can be used by the service operators.

This initial one will add a `ValidateStorageRequest()` func which validates a storage request meets a provided min size. Depending on the err (bool) parameter, the check result is either a warning or an invalid error.


abays (approved):
/lgtm

PR #523, opened 2024-06-26
DNM Revert "[svc] return error if no LoadBalancer IP"

averdagu:
This reverts commit 38b791ecf590b82914fe720250823b54c9c45b46.
This pull-request is only to add it as depends-on to the commit https://review.rdoproject.org/r/c/testproject/+/53195
I suspect this patch is making the gate fail earlier, once confirmed I'll properly find a way to fix gate.


averdagu:
Did not work as expected, closing it.

PR #524, opened 2024-07-01
Adding general secret selector struct

jpodivin:
The same copy and pasted code is being used across our repos with minor variations for individual operators.[0] I propose we move it here. Gradually removing local implementations as we go. 

[0] https://github.com/search?q=org%3Aopenstack-k8s-operators+%22PasswordSelector+struct%22+language%3AGo&type=code


modules/common/secret/types.go
```diff
@@ -0,0 +1,27 @@
...
+package secret
+
+// PasswordSelector
+type PasswordSelector struct {
+	// +kubebuilder:validation:Optional
+	// Service - Selector to get the service password from the Secret
+	Service string `json:"service"`
+	// +kubebuilder:validation:Optional
+	// AuthEncryptionKey - Selector to get the auth encryption key from the Secret
+	AuthEncryptionKey string `json:"authEncryptionKey"`
```
stuggi, line 26: this one is only used in heat, right. we should not have it in all the other services, where there is no usage of the parameter

modules/common/secret/types.go
```diff
@@ -0,0 +1,27 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package secret
+
+// PasswordSelector
+type PasswordSelector struct {
+	// +kubebuilder:validation:Optional
```
stuggi, line 21: the defaulting kubebuilder annotation can not be used, like its done today, in the service operators if we move it here, especially if you have multiple parameters in the struct and you want to allow the user to only specify the ones they want to change from the default. otherwise you'd always have to specify all.

modules/common/secret/types.go
jpodivin: Nova operator uses `MetadataSecret` for similar purpose [1]. Selecting field from secret is pretty obvious extension of the selector that may be introduced in other operators at some point. And it would be much easier if the struct had support for it. 
On the other hand, one empty field won't tax resources in a noticeable way.

[1] https://github.com/openstack-k8s-operators/nova-operator/blob/9ea2124e10e9a748633e2edf5db3d7446f653be9/api/v1beta1/common_types.go#L79

modules/common/secret/types.go
jpodivin: I suppose that makes sense. They will have to set value for at least one selector anyway.

modules/common/secret/types.go
stuggi: not sure I get your comment. you say AuthEncryptionKey is expected to come in other operators?

modules/common/secret/types.go
stuggi: sorry, don't get your reply. can you explain it.

modules/common/secret/types.go
jpodivin: I'm saying that it isn't there now, but that does make sense to have capability to select field from retrieved secret. The use case is general enough that I can see it showing up in other operators. And if it does not, which is possible, the field just won't be used.

stuggi: but we should not have fields in operators which are not being used. so I am not convinced a shared Passwordselector should have e.g. `authEncryptionKey`

stuggi: also we have to keep in mind that in future, when there is the need for new field it will be added to all the operators. we do not have a versioning for api structs, like it is for the apis in the lib-common. at least not right now.

jpodivin: In that case the only option is to stick with a single selector ... it's not going to be as useful in the long run, and I'm not really sure it's that bad to have one extra field.

stuggi: es mentioned on the other parameter, you can also not default it to an per operator specific value.

jpodivin: I'm not, there is no default there. Just `Optional`.

stuggi: for authEncryptionKey yes, which I think should not be in operators which don't use it,  but we'd need a default for Service

jpodivin: Doing that with kubebuilder annotation is obviously not going to work, not when the struct is defined here. We can however do that with a defaulting webhook.

stuggi: yes, but wouldn't it be then much easier/simpler to keep the struct in the operator itself, instead of adding a defaulting logic?

jpodivin: If you see that is a bridge too far than I see no reason to keep this PR open.

PR #526, opened 2024-07-08
Empty commit to make place for the v0.4.0 tag

gibizer:
To be able to distinguish between pseudoversions from main and from 18.0.0-proposed we need to start tagging main with v0.4.0 while keeping 18.0.0-proposed using the v0.3.x tags. Naturally we would only need to start tagging after the first commit arrives to main. But to avoid forgetting the tagging at that point we add an empty commit and tag it instead.


stuggi (approved):
/lgtm

PR #527, opened 2024-07-08
Add function to check for OCP IPv6 cluster network

olliewalsh:
The RabbitMQ config differs for IPv6 so add a function to determine if IPv6 is enabled on the OCP cluster network

Related: OSPRH-8372

### DIFF
--- a/modules/common/ocp/ocp.go
+++ b/modules/common/ocp/ocp.go
@@ -18,7 +18,6 @@ package ocp
 
 import (
 	"context"
-	"strings"
 
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 
@@ -26,6 +25,7 @@ import (
 	"gopkg.in/yaml.v3"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+	k8s_utils "k8s.io/utils/net"
 )
 
 // IsFipsCluster - Check if OCP has fips enabled which is a day 1 operation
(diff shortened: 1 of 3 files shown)


modules/common/ocp/ocp.go
```diff
@@ -47,3 +56,29 @@ func IsFipsCluster(ctx context.Context, h *helper.Helper) (bool, error) {
...
+	installConfig, err := getInstallConfig(ctx, h)
+	if err != nil {
+		return false, err
+	}
+
+	networking, ok := installConfig["networking"].(map[string]interface{})
+	if !ok {
+		return false, nil
+	}
+	clusterNetworks, ok := networking["clusterNetwork"].([]interface{})
```
olliewalsh, line 71: TODO: not safe to assume the cm reflects the current state since networks can be added after initial deployment

modules/common/ocp/ocp.go
```diff
@@ -47,3 +49,19 @@ func IsFipsCluster(ctx context.Context, h *helper.Helper) (bool, error) {
...
+// HasIPv6ClusterNetwork - Check if OCP has an IPv6 cluster network
+func HasIPv6ClusterNetwork(ctx context.Context, h *helper.Helper) (bool, error) {
+	networkConfig := &ocp_config.Network{}
+	err := h.GetClient().Get(ctx, types.NamespacedName{Name: "cluster", Namespace: ""}, networkConfig)
+	if err != nil {
+		return false, err
+	}
+
+	for _, clusterNetwork := range networkConfig.Status.ClusterNetwork {
+		if strings.Count(clusterNetwork.CIDR, ":") >= 2 {
```
stuggi, line 62: could also use `IsIPv6CIDRString` from https://github.com/kubernetes/utils/blob/master/net/ipfamily.go#L154C6-L154C22

olliewalsh: Done

stuggi (approved):
/lgtm

PR #536, opened 2024-07-24
Add functions for creating AEE jobs

bshephar:
This changes moves the logic from the openstack-ansibleee-operator controller into lib-common so that we can remove the unnecessary abstraction between OpenStackAnsibleEE CR's and Kubernetes Jobs

Jira: https://issues.redhat.com/browse/OSPRH-8900


modules/ansible/ansibleee.go
```diff
@@ -0,0 +1,321 @@
...
+}
+
+// JobForOpenStackAnsibleEE returns a openstackansibleee Job object
+func (a *EEJob) JobForOpenStackAnsibleEE(h *helper.Helper) (*batchv1.Job, error) {
+	ls := labelsForOpenStackAnsibleEE(a.Name, a.Labels)
+
+	args := a.Args
+
+	playbook := a.Playbook
+	if len(args) == 0 {
```
fmount, line 85: Does having `args` > 0 implies that `len(playbook)` (that sets the default `playbook.yaml`) shouldn't be evaluated?

modules/ansible/ansibleee.go
```diff
@@ -0,0 +1,321 @@
...
+		job.Spec.Template.Spec.InitContainers = a.InitContainers
+	}
+	if len(a.ServiceAccountName) > 0 {
+		job.Spec.Template.Spec.ServiceAccountName = a.ServiceAccountName
+	}
+	// Set primary inventory if specified as string
+	existingInventoryMounts := ""
+	if len(a.Inventory) > 0 {
+		setRunnerEnvVar(h, "RUNNER_INVENTORY", a.Inventory, "inventory", job, hashes)
+		existingInventoryMounts = "/runner/inventory/inventory.yaml"
```
fmount, line 148: should `/runner/inventory/inventory.yaml`, as well as `playbook.yaml` on L87 be const?

modules/ansible/ansibleee.go
```diff
@@ -0,0 +1,321 @@
...
+	_, errorHash := hashOfInputHashes(hashes)
+	if errorHash != nil {
+		return nil, fmt.Errorf("error generating hash of input hashes: %w", errorHash)
+	}
+
+	return job, nil
+}
+
+// labelsForOpenStackAnsibleEE returns the labels for selecting the resources
+// belonging to the given openstackansibleee CR name.
```
fmount, line 219: I think we don't have an `openstackansibleee` CR anymore, right? I assume this comment was copy/paste from the AEE operator

modules/ansible/ansibleee.go
```diff
@@ -0,0 +1,321 @@
...
+func labelsForOpenStackAnsibleEE(name string, labels map[string]string) map[string]string {
+	ls := map[string]string{
+		"app":                   "openstackansibleee",
```
fmount, line 222: this label might be turned into a `const`

modules/ansible/ansibleee.go
```diff
@@ -0,0 +1,321 @@
...
+		return "", err
+	}
+	return hash, nil
+}
+
+func hashOfInputHashes(hashes map[string]string) (string, error) {
+	var err error
+	var hash string
+	var builder strings.Builder
+	for key, value := range hashes {
```
stuggi, line 307: is this a one time calc of hashOfInputHashes? otherwise it should be sorted by keys. as there is no sorted order when look over hashmap, the result can have different order when rerunning and therefore different hash sum.

modules/ansible/ansibleee.go
bshephar: Yeah, correct. If args are passed then `playbook` is ignored:
https://github.com/openstack-k8s-operators/lib-common/pull/536/files/ef2c08297fa11cf6c5b9cf71ebf1a86b48eb1204#diff-e150238f8cf75bd058263dacb9a03d71a789d1be75a9d0a45b0fcb3ba37bd8a6R44-R45

modules/ansible/ansibleee.go
bshephar: Ah, yes, true. This will need to be changed

modules/ansible/ansibleee.go
bshephar: Yeah, can probably make these const instead

modules/ansible/ansibleee.go
bshephar: I don't think we're currently doing anything with this. I think I'll just remove this function entirely.

bshephar:
Ok, I think we can just move all of this into the openstack-operator as well. Then it can be leveraged by:
https://github.com/openstack-k8s-operators/openstack-operator/pull/972

To avoid inflating the size and complexity of the existing 972 PR. I'll just add these functions first, where they will be unused, then update the 972 PR to use them.

bshephar:
@fmount @stuggi  Let's continue the discussion over on:
https://github.com/openstack-k8s-operators/openstack-operator/pull/1004

PR #537, opened 2024-07-25
Requeue when TLS secret isn't found

fao89:
closes OSPRH-7090

### DIFF
--- a/modules/certmanager/certificate.go
+++ b/modules/certmanager/certificate.go
@@ -103,7 +103,7 @@ func (c *Certificate) CreateOrPatch(
 	ctx context.Context,
 	h *helper.Helper,
 	owner client.Object,
-) (ctrl.Result, error) {
+) (ctrl.Result, controllerutil.OperationResult, error) {
 	var err error
 	cert := &certmgrv1.Certificate{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


modules/certmanager/certificate.go
```diff
@@ -243,6 +243,10 @@ func EnsureCert(
 	// get cert secret
 	certSecret, _, err := secret.GetSecret(ctx, helper, certSecretName, namespace)
 	if err != nil {
+		if k8s_errors.IsNotFound(err) && op == controllerutil.OperationResultCreated {
+			helper.GetLogger().Info(fmt.Sprintf("Secret %s not found, reconcile in %s", certSecretName, cert.timeout))
```
stuggi, line 247: I don't think this is a good approach. This would hide that the cert is missing just as a log entry in the controller manager pod log. I think we should still pass it as an error to the upper level and either handle there the IsNotFound or just reflect it as an error and reconcile based on the updated event on the cert. In general related to the Jira I think its a valid information, especially if there is really an issue with cert-manager and the deployment won't progress.

fao89: Requeueing when the cert was created makes sense (give more time to see if the secret gets created), in any other case we should raise an error. On another level, we can't know if the cert was recently created or not

stuggi: But this will prevent getting the error when there is an issue and the cert secret creation won't progress. The error should be visible. Why is it an issue if the upper level reports for a short time that the cert secret is not there. Its a valid situation.

fao89: it would prevent the first time, but after requeueing it would raise the error. In a similar situation, we had the feedback that displaying an error while it is just a failed attempt (that the operator would handle by itself) is confusing to the users

stuggi: ah right, sorry missed the `op == controllerutil.OperationResultCreated`

stuggi (approved):
/lgtm

PR #541, opened 2024-07-29
Fix for Mac

bshephar:
There's a second function that encodes the path and needs to also check for the darwin runtime before doing it. This enables tests of operators that import openshift CRD's to run on MacOS platforms.


stuggi:
/lgmt don't have a mac to test, but I am ok to land it to unblock users

stuggi (approved):
/lgtm don't have a system to validate, but looks good to me to unblock users

PR #546, opened 2024-08-05
[cert] Make sure DNSNames and IPAddresses are sorted

stuggi:
To reduce possibility of changing certs, make sure lists of DNSNames and IPAddresses are sorted.

### DIFF
--- a/modules/certmanager/certificate.go
+++ b/modules/certmanager/certificate.go
@@ -19,11 +19,13 @@ package certmanager
 import (
 	"context"
 	"fmt"
+	"sort"
 	"time"
 
 	certmgrv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	certmgrmetav1 "github.com/cert-manager/cert-manager/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


bshephar (approved):

modules/certmanager/certificate.go
```diff
@@ -66,10 +67,15 @@ func NewCertificate(
 	certificate *certmgrv1.Certificate,
 	timeout time.Duration,
 ) *Certificate {
-	return &Certificate{
+	crt := &Certificate{
 		certificate: certificate,
 		timeout:     timeout,
 	}
+
+	sort.Strings(crt.certificate.Spec.IPAddresses)
```
olliewalsh, line 75: Sorting as strings will not order by IP address correctly. Could we parse the IPs and then sort (as byte slices I guess)?

stuggi: right, let me fix that

stuggi: updated and added an ip sort func

olliewalsh (approved):
/lgtm

PR #547, opened 2024-08-06
Log 'not found' error instead of returning it alongside RequeueAfter

abays:
Returning a non-nil err with a RequeueAfter result causes k8s to discard the explicit requeue and wait much longer than desired.


stuggi (approved):
/lgtm

PR #548, opened 2024-08-07
Add message for waiting for TLS input resources

abays:
Given that we no longer return an error for k8s "not found" errors, we should add a message for TLSInput that can be used in `ctrlResult != ctrl.Result{}` situations (which indicate that a `Secret` is not found).


stuggi (approved):
/lgtm

PR #557, opened 2024-09-04
Fixes extra newline from confs

auniyal61:
As we are using templates to dynamicaly create confs on conditions, Often extra newlines are getting added in confs, we should remove them.

### DIFF
--- a/modules/common/util/template_util.go
+++ b/modules/common/util/template_util.go
@@ -57,6 +57,48 @@ type Template struct {
 	ConfigOptions      map[string]interface{} // map of parameters as input data to render the templates
 	SkipSetOwner       bool                   // skip setting ownership on the associated configmap
 	Version            string                 // optional version string to separate templates inside the InstanceType/Type directory. E.g. placementapi/config/18.0
+	PostProcessCleanup bool                   // optional boolean parameter to remove extra new lines from service confs, by default set to false
+}
+
+// This function removes extra space and new-lines from conf data
+func removeSubsequentNewLines(rawConf string) string {
+	lines := strings.Split(rawConf, "\n")
+	var result []string
+	var prevLineWasBlank bool
+
+	for _, line := range lines {
+		trimmedLine := strings.TrimSpace(line)
+
+		if trimmedLine == "" {
+			prevLineWasBlank = true
+			// if current line is blank, no need to add it in result
+		} else {
+			if strings.HasPrefix(trimmedLine, "[") && strings.HasSuffix(trimmedLine, "]") {
+				// section-header
+				if len(result) > 0 && !prevLineWasBlank {
+					result = append(result, "")
+				}
+				var sectionHeaderLine string
+				if len(result) > 0 {
+					// new section-hearder
+					sectionHeaderLine = "\n" + line
+				} else {
+					sectionHeaderLine = line
+				}
+				result = append(result, sectionHeaderLine)
+			} else {
+				// secion-body
+				result = append(result, line)
+			}
+			// reset flag
+			prevLineWasBlank = false
+		}
+	}
+
+	// add an extra line at EOF
+	result = append(result, "")
+
+	return strings.Join(result, "\n")
 }
 
 // GetTemplatesPath get path to templates, either running local or deployed as container
@@ -135,6 +177,7 @@ func ExecuteTemplate(templateFile string, data interface{}) (string, error) {
 	if err != nil {
 		return "", err
 	}
+
 	return renderedTemplate, nil
 }
 
@@ -242,6 +285,15 @@ func GetTemplateData(t Template) (map[string]string, error) {
 			return nil, err
 		}
 		data[filename] = renderedTemplate
+	}
+
+	if t.PostProcessCleanup {
+		for filename, content := range data {
+			if strings.HasSuffix(filename, ".conf") {
+				// as of now, only clean for confs
+				data[filename] = removeSubsequentNewLines(content)
+			}
+		}
 	}
 
 	return data, nil
--- a/modules/common/util/template_util.go
+++ b/modules/common/util/template_util.go
@@ -45,53 +45,43 @@ const (
 
 // Template - config map and secret details
 type Template struct {
-	Name               string                 // name of the cm/secret to create based of the Template. Check secret/configmap pkg on details how it is used.
-	Namespace          string                 // name of the nanmespace to create the cm/secret. Check secret/configmap pkg on details how it is used.
-	Type               TType                  // type of the templates, see TTtypes
-	InstanceType       string                 // the CRD name in lower case, to separate the templates for each CRD in /templates
-	SecretType         corev1.SecretType      // Secrets only, defaults to "Opaque"
-	AdditionalTemplate map[string]string      // templates which are common to multiple CRDs can be located in a shared folder and added via this type into the resulting CM/secret
-	CustomData         map[string]string      // custom data which won't get rendered as a template and just added to the resulting cm/secret
-	Labels             map[string]string      // labels to be set on the cm/secret
-	Annotations        map[string]string      // Annotations set on cm/secret
-	ConfigOptions      map[string]interface{} // map of parameters as input data to render the templates
-	SkipSetOwner       bool                   // skip setting ownership on the associated configmap
-	Version            string                 // optional version string to separate templates inside the InstanceType/Type directory. E.g. placementapi/config/18.0
-	PostProcessCleanup bool                   // optional boolean parameter to remove extra new lines from service confs, by default set to false
+	Name                        string                 // name of the cm/secret to create based of the Template. Check secret/configmap pkg on details how it is used.
+	Namespace                   string                 // name of the nanmespace to create the cm/secret. Check secret/configmap pkg on details how it is used.
+	Type                        TType                  // type of the templates, see TTtypes
+	InstanceType                string                 // the CRD name in lower case, to separate the templates for each CRD in /templates
+	SecretType                  corev1.SecretType      // Secrets only, defaults to "Opaque"
+	AdditionalTemplate          map[string]string      // templates which are common to multiple CRDs can be located in a shared folder and added via this type into the resulting CM/secret
+	CustomData                  map[string]string      // custom data which won't get rendered as a template and just added to the resulting cm/secret
+	Labels                      map[string]string      // labels to be set on the cm/secret
+	Annotations                 map[string]string      // Annotations set on cm/secret
+	ConfigOptions               map[string]interface{} // map of parameters as input data to render the templates
+	SkipSetOwner                bool                   // skip setting ownership on the associated configmap
+	Version                     string                 // optional version string to separate templates inside the InstanceType/Type directory. E.g. placementapi/config/18.0
+	PostProcessConfFilesCleanup bool                   // optional boolean parameter to remove extra new lines from service confs, by default set to false
 }
 
 // This function removes extra space and new-lines from conf data
 func removeSubsequentNewLines(rawConf string) string {
 	lines := strings.Split(rawConf, "\n")
 	var result []string
-	var prevLineWasBlank bool
 
 	for _, line := range lines {
 		trimmedLine := strings.TrimSpace(line)
 
-		if trimmedLine == "" {
-			prevLineWasBlank = true
-			// if current line is blank, no need to add it in result
-		} else {
+		if trimmedLine != "" {
 			if strings.HasPrefix(trimmedLine, "[") && strings.HasSuffix(trimmedLine, "]") {
 				// section-header
-				if len(result) > 0 && !prevLineWasBlank {
-					result = append(result, "")
-				}
-				var sectionHeaderLine string
 				if len(result) > 0 {
-					// new section-hearder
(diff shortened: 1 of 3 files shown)


stuggi:
This seems to be not correct for e.g. this case https://go.dev/play/p/5ms4rPxaX0l where you have 2 new lines:
```
line 1
line 2


line 3
```
it would result in 
```
line 1
line 2line 3
```
Can you explain what you try to achieve? Do you want to replace any empty line from the resulting rendered template? 

FYI, a commit to fix the govet was just merged. you'd have to rebase.

modules/common/util/template_util.go
```diff
@@ -157,6 +157,8 @@ func ExecuteTemplateData(templateData string, data interface{}) (string, error)
 		"add":   add,
 		"lower": lower,
 	}
+	// remove extra newlines
+	templateData = strings.ReplaceAll(templateData, "\n\n\n", "")
```
gibizer, line 161: yeah this potentially remove all the new lines between two lines of config making the config invalid. 

I think what we need is replacing every two or more consecutive new lines with a single one until we have no more two consecutive new lines.

gibizer (changes requested):
@stuggi How would you feel about adding a post processing hook (a function parameter) to the untils.Template so the service operators can implement any kind of post processing after the template is rendered? Would it be too much complexity?

stuggi:
could you explain the motivation of doing it as a separate task, instead of updating the template to suppress new lines https://pkg.go.dev/text/template#hdr-Text_and_spaces / https://docs.gomplate.ca/syntax/#suppressing-leading-newlines ?

auniyal61:
> it would result in
> 
> ```
> line 1
> line 2line 3
> ```
> 

so initially when I tested it with  nova-operator, I decided to go with below because we need atleast 2 lines for new section.
	`s = strings.ReplaceAll(s, "\n\n\n", "\n")`
missed '\n' here, as I tested with sample conf.file adding at https://github.com/openstack-k8s-operators/lib-common/tree/main/modules/common/util/testdata/templates/testservice/config

> could you explain the motivation of doing it as a separate task, instead of updating the template to suppress new lines https://pkg.go.dev/text/template#hdr-Text_and_spaces / https://docs.gomplate.ca/syntax/#suppressing-leading-newlines ?

we tried that here https://github.com/openstack-k8s-operators/nova-operator/pull/842

but it seems confusing, also conf did not rendered properly.
so added here, so this can reflect on all conf files.

SeanMooney:
> could you explain the motivation of doing it as a separate task, instead of updating the template to suppress new lines https://pkg.go.dev/text/template#hdr-Text_and_spaces / https://docs.gomplate.ca/syntax/#suppressing-leading-newlines ?

yes so we have looked at doing that twice
amit ahs a draft and it is very messy and makes the template harder to read.

https://github.com/openstack-k8s-operators/nova-operator/pull/855

I did a POC of that about 18 months ago when we were refactoring our templates
and abandoned it for the same reason 


the current apporch is not correct.

i would personnly do this in a streaming fashion.
scan the rendered template line by line and skip newlines if the previous line was a new line
so i would not use replace or similar regex approaches

i like the idea of being able to pass a object via an interface to the function but
if we take that approach i think a simple line striping functor object should be added to lib common since i suspect that will be what people normally want.

we could do this in the nova operator today but we wanted to do it in lib common to benefit all the operators

at the end of the day this is a cosmetic enhancement but its sill nice to have for readablity

auniyal61:
its still not looking good output confs
https://logserver.rdoproject.org/55/855/79661db75220177782ddf4f496cb24c008b1df0f/github-check/nova-operator-tempest-multinode/4209eeb/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/secrets/nova/

so need a better approach

stuggi:
ack, thanks for the explanation. doing this in the template can be confusing. I am ok to add it as a post processing, but I think it should be optional so that we keep the current behavior as the default and one needs to enable it if they like.

SeanMooney:
something like this i think would work 

https://go.dev/play/p/iOQ5UJ_RpvW?v=goprev.go?download=true.go?download=true.go?download=true.go?download=true

modules/common/util/template_util.go
gibizer: Also this only applies the transformation on the template not on the rendered text. The rendered text ends up in the buff variable after tmpl.Execute()

auniyal61: Updated outside execute.

auniyal61:
I am sure this should have worked this time, and printed like this: https://privatebin.corp.redhat.com/?9a9e2b43f5bf87bd#AyXQa1E1amv7vpGFbKeQGUKNu99ZNdN3t1JBHAb2Wppx

but its notcomming correctly
https://logserver.rdoproject.org/57/857/e3248dbf95dfb1878d852beef4475bfb68f4e811/github-check/nova-operator-tempest-multinode/616a00d/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/secrets/nova/nova-api-config-data.yaml-01-nova.conf

modules/common/util/template_util.go
```diff
@@ -135,6 +176,12 @@ func ExecuteTemplate(templateFile string, data interface{}) (string, error) {
 	if err != nil {
 		return "", err
 	}
+
```
gibizer, line 179: ExecuteTemplateData is called from different places to render templatest:
* here from ExecuteTemplate
* but also from secret.createOrUpdateSecret()

Moreover the actual template rendering happens via the text/template module by creating a template via template.New() and eventually calling Execute() on it. If you search for that pattern in this file then you will see that there are two functions that actually executes templates:
* ExecuteTemplateData 
* ExecuteTemplateFile

So both call path need to be handled. The nova.conf rendering goes through the ExecuteTemplateFile path this is why you don't see your cleaning applied to that config file.

gibizer: I would do the following:
* pull the template rendering part of the two paths to a common function
* pass down the template_utils.Template struct to that common function
* extend the template_utils.Template struct with a field that can configure if / what post processing is needed for the give template
* use that information in the common function to do the post processing.

auniyal61:
cleanup worked
cleanup worked
https://logserver.rdoproject.org/57/857/0b3a0d59ae6b1ae5c0bba96da969a76b5fabd90b/github-check/nova-operator-tempest-multinode/0cbc344/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/secrets/nova/nova-api-config-data.yaml-01-nova.conf

modules/common/util/template_util.go
auniyal61: Added bool in Template to make cleanup optional

gibizer:
I've pushed an extra commit to the branch with unit test for the cleaning logic and that shows some inconcistencies around the empty line handling between sections.

modules/common/util/template_util.go
```diff
@@ -244,5 +287,14 @@ func GetTemplateData(t Template) (map[string]string, error) {
 		data[filename] = renderedTemplate
 	}
 
+	if t.PostProcessCleanup {
+		for filename, content := range data {
+			if strings.HasSuffix(filename, ".conf") {
```
gibizer, line 292: As the cleaning is now optional I think the caller can make sure it is only enabled for the files necessary so we don't need to filter for extension (which also not super specific as the current cleaning is for ini like files, but a .conf file might not be ini like)

modules/common/util/template_util.go
```diff
@@ -57,6 +57,48 @@ type Template struct {
 	ConfigOptions      map[string]interface{} // map of parameters as input data to render the templates
 	SkipSetOwner       bool                   // skip setting ownership on the associated configmap
 	Version            string                 // optional version string to separate templates inside the InstanceType/Type directory. E.g. placementapi/config/18.0
+	PostProcessCleanup bool                   // optional boolean parameter to remove extra new lines from service confs, by default set to false
```
gibizer, line 60: I still prefer passing in the function references to run as post processing as that allows for a lot more flexibility in the future. 

But if the others wants a separate boolean for each future post processing then at least be super specific what this boolean enables. The code Sean provided not just remove extra empty lines from the rendered template but also insert missing empty lines between ini like sections.

gibizer (changes requested):

modules/common/util/template_util.go
gibizer: We had a call with @auniyal61 and I learned that a single Template instance can actually mean a list of rendered files. Both via AdditionalTemplate and the logic that picks up a set of files from a path + glob. This means that a single PostProcessCleanup bool is actually applied not to a single file but to a set of diverse files(potentially a mix of different file types like ini, kolla json, apache conf). This would also be true for the set of functions as well if we change this to that type. This makes the situation more complicated as post processing cannot be configured per template file. Changing the Template to allow per file post processing is seems to be a big work so I drop the idea out of the scope of this change.

@stuggi How do you feel about eventually refactoring Template? I see three possible directions we could go
1. keep a single Template struct to pick up multiple files and render them as today, and then extend the interface to make such rendering still configurable per template file level
2. break the coupling between template rendering and Secret / CM creation with the rendered template. Make it two separate, client callable, steps. Like below to allow post processing on the client side.
   ```
   renderedTemplates = utils.RenderTemplate(template utils.Template)
   // the service operator can do post processing here on the client side
   utils.CreateCM/Secret(renderedTemplates)
   ```
3. break the current Template struct to have a list of Template structs one for each template file making the template rendering configuration per template file struct.

modules/common/util/template_util.go
gibizer: as a single Template represents a set of diverse files instead of a single file. My suggestion that the client can use the PostProcessCleanup boolean instead of the static .conf suffix filtering does not work. So we need to keep the filter to avoid removing lines from other than ini lines.

modules/common/util/template_util.go
```diff
@@ -45,18 +45,50 @@ const (
...
+	SecretType                  corev1.SecretType      // Secrets only, defaults to "Opaque"
+	AdditionalTemplate          map[string]string      // templates which are common to multiple CRDs can be located in a shared folder and added via this type into the resulting CM/secret
+	CustomData                  map[string]string      // custom data which won't get rendered as a template and just added to the resulting cm/secret
+	Labels                      map[string]string      // labels to be set on the cm/secret
+	Annotations                 map[string]string      // Annotations set on cm/secret
+	ConfigOptions               map[string]interface{} // map of parameters as input data to render the templates
+	SkipSetOwner                bool                   // skip setting ownership on the associated configmap
+	Version                     string                 // optional version string to separate templates inside the InstanceType/Type directory. E.g. placementapi/config/18.0
+	PostProcessConfFilesCleanup bool                   // optional boolean parameter to remove extra new lines from service confs, by default set to false
+}
```
SeanMooney, line 61: so we could likely proceed with this version for now

however how i would approach this is as follows

1 defince a new interface with one function which is  takes a string and returns a string to post_process the rendered file.

instead of adding a bool here we would add a map of file exteion to interface object

then we would loop over all the files in the template and if there is registered post processor for that exteion in the map we can apply it to that specific file

we could also have a sentenal value for a default handleler if we want but i think that is likely not needed.

gibizer: I guess that is close to my suggestion/question 1 above. https://github.com/openstack-k8s-operators/lib-common/pull/557#discussion_r1762545498

SeanMooney: by hte way the key could be a regex instead of an extnetion allowing mapping the handler to either a file or set of files.

this would just be applied to the keys for the config map/serecte but that might be overkill.

its simple to do and reason about however so we might just want to do it now

SeanMooney: yep more or less

i think breaking up the templates and configmap logic ectra will be a lot more work

we can do that eventually but for now i think this would be a minimal change to the existing proposal that we can extend with minimal impact on other operators

and in the future if we want to do that larger refactor for 2/3 we still can

i would do that by introducing a new Struct by the way and keeping but deprecating the current one so we can avoid a hard switch over.

so for the short term  i would vote for 1 and then see if we need to do 2/3 later

modules/common/util/template_util.go
auniyal61: Hi @stuggi , can you please suggest on this.

stuggi: @auniyal61 sorry for the delay. I was out last week. I'll get back to this asap.

stuggi: @gibizer @SeanMooney soffy for the delay, I finally had some time today to look into this. What you think about this approach [1]. This adds new template functions , especially for our use case here `execTempl` and `removeNewLines`. With this you can define your template inline and then render the template using the `execTempl` and remove multiple empty lines using `removeNewLines`. An example on how you could use that in a template is shown in the description.

Specific for the nova.conf in this example PR [2]. The benefit is that you could control this just from inside the template and no other code change is required.

[1] https://github.com/openstack-k8s-operators/lib-common/pull/569
[2] https://github.com/openstack-k8s-operators/nova-operator/pull/874

auniyal61:
closing in favor of https://github.com/openstack-k8s-operators/lib-common/pull/569

PR #559, opened 2024-09-05
[tls] ValidateCertSecret to not return ctrlResult

stuggi:
ValidateCertSecrets miss information on which cert secret is missing. Lets not return a crtlResult in case of a cert secret is missing, instead return a NotFound error and let the caller decide what to do.

Jira: OSPRH-9991

### DIFF
--- a/modules/common/tls/tls.go
+++ b/modules/common/tls/tls.go
@@ -29,12 +29,13 @@ import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/service"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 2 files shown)


modules/common/tls/tls.go
```diff
@@ -283,12 +282,15 @@ func (s *Service) ValidateCertSecret(ctx context.Context, h *helper.Helper, name
 		h.GetClient(),
 		5*time.Second)
 	if err != nil {
-		return "", ctrlResult, err
+		return "", err
 	} else if (ctrlResult != ctrl.Result{}) {
-		return "", ctrlResult, nil
+		return "", k8s_errors.NewNotFound(
```
stuggi, line 287: this is basically the change to return a NotFound error in case VerifySecret returns a ctrlResult

mrkisaolamb (commented):
lgtm

modules/common/tls/tls.go
abays: This will remove the desired requeue time though (the one coming from `VerifySecret()` and then cause exponential requeue times via the error returned?

fao89 (approved):

modules/common/tls/tls.go
stuggi: depends on how the caller handles the error. for those who watch the secret they reconcile when its there for the event. for others they can return a requeue time

abays (approved):
/lgtm

PR #560, opened 2024-09-05
Fix ValidateCertSecret error message

stuggi:
right now err.Error() renders as
~~~
secrets \"%s not found in namespace %s\" not found
~~~

This changes the msg so that it returns

~~~
secrets \"%s in namespace %s\" not found
~~~

Jira: OSPRH-9991


abays (approved):
/lgtm

PR #569, opened 2024-10-11
Add additional template functions

stuggi:
Adds templat functions which:
- to exec an inline template
- indent a template
- remove n continuous empty lines from a template

An example would be a template like:
```
{{define "my-template"}}new template

content

with empty lines

to
remove
{{end}}
Some other template content and add the rendered from my-template
{{$var := execTempl "my-template" . | removeNewLines 1}}
{{$var}}
```

### DIFF
--- a/modules/common/util/template_util.go
+++ b/modules/common/util/template_util.go
@@ -17,6 +17,7 @@ limitations under the License.
 package util
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
 	"os"
@@ -138,6 +139,82 @@ func ExecuteTemplate(templateFile string, data interface{}) (string, error) {
 	return renderedTemplate, nil
 }
 
+// template functions
+var tmpl *template.Template
+
+// template function which allows to execute a template from within
+// a template file.
+// name - name of the template as defined with with `{{define "some-template"}}your template{{end}}
+// data - data to pass into to render the template for all can use `.`
+func execTempl(name string, data interface{}) (string, error) {
+	buf := &bytes.Buffer{}
+	err := tmpl.ExecuteTemplate(buf, name, data)
+	return buf.String(), err
+}
+
+// template function to indent the template with n tabs
+func indent(n int, in string) string {
+	var out string
+	s := bufio.NewScanner(bytes.NewReader([]byte(in)))
+	for s.Scan() {
+		line := strings.TrimSpace(s.Text())
+		for i := 0; i < n; i++ {
+			line = "\t" + line
+		}
+		out += line + "\n"
+	}
+	return out
+}
+
+// template function to remove empty lines if there are > n continuous empty lines
+func removeNewLines(n int, in string) string {
+	var out string
+	s := bufio.NewScanner(bytes.NewReader([]byte(in)))
+
+	// Variable to keep track of consecutive empty lines
+	emptyLineCount := 0
+	for s.Scan() {
+		line := s.Text()
+
+		if strings.TrimSpace(line) == "" {
+			emptyLineCount++
+			// If we have already seen more then n empty lines, skip this one
+			if emptyLineCount > n {
+				continue
+			}
+		} else {
+			// Reset the empty line counter when we encounter a non-empty line
+			emptyLineCount = 0
+		}
+
+		out += line + "\n"
+	}
+	return out
+}
+
+// This function removes extra space and new-lines from conf data.
+func removeNewLinesInSections(in string) string {
+	var out string
+	s := bufio.NewScanner(bytes.NewReader([]byte(in)))
+
+	for s.Scan() {
+		line := strings.TrimSpace(s.Text())
+
+		if line != "" {
+			if strings.HasPrefix(line, "[") && strings.HasSuffix(line, "]") {
+				// new section-header
+				if len(out) > 0 {
+					out += "\n"
+				}
+			}
+
+			out += line + "\n"
+		}
+	}
+
+	return out
+}
+
 // template function to increment an int
 func add(x, y int) int {
 	return x + y
@@ -153,11 +230,16 @@ func lower(s string) string {
 func ExecuteTemplateData(templateData string, data interface{}) (string, error) {
 
 	var buff bytes.Buffer
+	var err error
 	funcs := template.FuncMap{
-		"add":   add,
-		"lower": lower,
+		"add":                      add,
+		"execTempl":                execTempl,
+		"indent":                   indent,
+		"lower":                    lower,
+		"removeNewLines":           removeNewLines,
+		"removeNewLinesInSections": removeNewLinesInSections,
 	}
-	tmpl, err := template.New("tmp").Option("missingkey=error").Funcs(funcs).Parse(templateData)
+	tmpl, err = template.New("tmp").Option("missingkey=error").Funcs(funcs).Parse(templateData)
 	if err != nil {
 		return "", err
 	}
@@ -193,10 +275,14 @@ func ExecuteTemplateFile(filename string, data interface{}) (string, error) {
 	file := string(b)
 	var buff bytes.Buffer
 	funcs := template.FuncMap{
-		"add":   add,
-		"lower": lower,
+		"add":                      add,
+		"execTempl":                execTempl,
+		"indent":                   indent,
+		"lower":                    lower,
+		"removeNewLines":           removeNewLines,
+		"removeNewLinesInSections": removeNewLinesInSections,
 	}
-	tmpl, err := template.New("tmp").Option("missingkey=error").Funcs(funcs).Parse(file)
+	tmpl, err = template.New("tmp").Option("missingkey=error").Funcs(funcs).Parse(file)
 	if err != nil {
 		return "", err
 	}
--- a/modules/common/util/template_util_test.go
+++ b/modules/common/util/template_util_test.go
@@ -37,6 +37,243 @@ func TestLower(t *testing.T) {
 	})
 }
 
+func TestIndent(t *testing.T) {
+
+	t.Run("Indent string", func(t *testing.T) {
+		g := NewWithT(t)
+		const in = `foo
+bar`
+		// 5 tabs and line break
+		const expct = `					foo
+					bar
+`
+
+		s := indent(5, in)
+
+		g.Expect(s).To(BeIdenticalTo(expct))
+	})
+}
+
+func TestRemoveNewLines(t *testing.T) {
+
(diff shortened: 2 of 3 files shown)


stuggi:
an example for using it in nova for the nova.conf is in https://github.com/openstack-k8s-operators/nova-operator/pull/874

.pre-commit-config.yaml
```diff
@@ -40,4 +40,8 @@ repos:
       exclude: ^vendor
     - id: no-commit-to-branch
     - id: trailing-whitespace
-      exclude: ^vendor
+      exclude: |
+        (?x)(
+            ^vendor|
+            ^modules/common/util/template_util_test.go
```
stuggi, line 46: I had to add this because of the test data for `TestRemoveNewLines` where I test preceding spaces

SeanMooney:
we coudl do this but i guess we were trying to avoid defienign our own dsl
I'm concurred that if we do this this way we may end up building a complex templaating system
so we should be very intentional about taking this direction if we do.

stuggi:
> we coudl do this but i guess we were trying to avoid defienign our own dsl I'm concurred that if we do this this way we may end up building a complex templaating system so we should be very intentional about taking this direction if we do.

yes we should not introduce a lot of/further complexity. At least right now I am not aware of something else needed. I can remove the `indent` func as it was not requested. What I like with this approach is that it uses the default way on how you  extend the go templating using these funcs.

SeanMooney:
> > we coudl do this but i guess we were trying to avoid defienign our own dsl I'm concurred that if we do this this way we may end up building a complex templaating system so we should be very intentional about taking this direction if we do.
> 
> yes we should not introduce a lot of/further complexity. At least right now I am not aware of something else needed. I can remove the `indent` func as it was not requested. What I like with this approach is that it uses the default way on how you extend the go templating using these funcs.

yep that is a plus as it not a net new way of doing this.
and this filter/fucntion style approach should feel famialr to anyone who has used jinja as well.

modules/common/util/template_util.go
```diff
@@ -138,6 +139,57 @@ func ExecuteTemplate(templateFile string, data interface{}) (string, error) {
...
+		for i := 0; i < n; i++ {
+			line = "\t" + line
+		}
+		out += line + "\n"
+	}
+	return out
+}
+
+// template function to remove empty lines if there are > n continuous empty lines
+func removeNewLines(n int, in string) (out string) {
```
auniyal61, line 169: IIUTC, this allow people to have n number of lines in between.
I think we may not need this in case conf files, but may be for other templates!

auniyal61:
I think its better approach, as there are only template level changes at operator side.

right now, the removeNewLines function is not working as we want/expected, I had tested it locally.

[current_nova.conf render](https://privatebin.corp.redhat.com/?87d05bb8bff66009#2wjY8nDRDC4tRNsTWBQTmcoQffVReFxzrBQPHusQGCcv)

[removeNewLines(0, str)](https://privatebin.corp.redhat.com/?8ad6a5ecb2989b3d#5d1KP2LvAjr66ZpEeTfqFf5yApZhKK4vLqtP3g1n9EMm)

[removeNewLines(1, str)
](https://privatebin.corp.redhat.com/?5e35003c489a5510#GjsXjKHT1WFhiu5XhDjJkCw1BnDEYEuggXGtvAdSBjUg)

[removeNewLines(2, str)](https://privatebin.corp.redhat.com/?6bf8a9871eb0b92a#CrXiairiJJJBUn69R3KyjibzaAKYrXEecX6A7Ca3rZwE)

[removeNewLines(3, str)](https://privatebin.corp.redhat.com/?c9a0b9c30e054542#EmtG75reJBWHQAM4RFMdT9yx98pnDRwTywUe5y3oqSjb)


(but that may not be the focus of this patch)

stuggi:
> I think its better approach, as there are only template level changes at operator side.
> 
> right now, the removeNewLines function is not working as we want/expected, I had tested it locally.
> 
> [current_nova.conf render](https://privatebin.corp.redhat.com/?87d05bb8bff66009#2wjY8nDRDC4tRNsTWBQTmcoQffVReFxzrBQPHusQGCcv)
> 
> [removeNewLines(0, str)](https://privatebin.corp.redhat.com/?8ad6a5ecb2989b3d#5d1KP2LvAjr66ZpEeTfqFf5yApZhKK4vLqtP3g1n9EMm)
> 
> [removeNewLines(1, str) ](https://privatebin.corp.redhat.com/?5e35003c489a5510#GjsXjKHT1WFhiu5XhDjJkCw1BnDEYEuggXGtvAdSBjUg)


Isnt' this what you wanted? it will remove multiple continuous empty lines and just keep a single?
Not covering the special header case

> 
> [removeNewLines(2, str)](https://privatebin.corp.redhat.com/?6bf8a9871eb0b92a#CrXiairiJJJBUn69R3KyjibzaAKYrXEecX6A7Ca3rZwE)
> 
> [removeNewLines(3, str)](https://privatebin.corp.redhat.com/?c9a0b9c30e054542#EmtG75reJBWHQAM4RFMdT9yx98pnDRwTywUe5y3oqSjb)
> 
> (but that may not be the focus of this patch)

auniyal61:
> > I think its better approach, as there are only template level changes at operator side.
> > right now, the removeNewLines function is not working as we want/expected, I had tested it locally.
> > [current_nova.conf render](https://privatebin.corp.redhat.com/?87d05bb8bff66009#2wjY8nDRDC4tRNsTWBQTmcoQffVReFxzrBQPHusQGCcv)
> > [removeNewLines(0, str)](https://privatebin.corp.redhat.com/?8ad6a5ecb2989b3d#5d1KP2LvAjr66ZpEeTfqFf5yApZhKK4vLqtP3g1n9EMm)
> 
> Isnt' this what you wanted? it will remove multiple continuous empty lines and just keep a single? Not covering the special header case
> 



yes, after adding for section-header with n =0, it should work
we need [this](https://logserver.rdoproject.org/57/857/0b3a0d59ae6b1ae5c0bba96da969a76b5fabd90b/github-check/nova-operator-tempest-multinode/0cbc344/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/secrets/nova/nova-api-config-data.yaml-01-nova.conf).

stuggi:
> > > I think its better approach, as there are only template level changes at operator side.
> > > right now, the removeNewLines function is not working as we want/expected, I had tested it locally.
> > > [current_nova.conf render](https://privatebin.corp.redhat.com/?87d05bb8bff66009#2wjY8nDRDC4tRNsTWBQTmcoQffVReFxzrBQPHusQGCcv)
> > > [removeNewLines(0, str)](https://privatebin.corp.redhat.com/?8ad6a5ecb2989b3d#5d1KP2LvAjr66ZpEeTfqFf5yApZhKK4vLqtP3g1n9EMm)
> > 
> > 
> > Isnt' this what you wanted? it will remove multiple continuous empty lines and just keep a single? Not covering the special header case
> 
> yes, after adding for section-header with n =0, it should work we need [this](https://logserver.rdoproject.org/57/857/0b3a0d59ae6b1ae5c0bba96da969a76b5fabd90b/github-check/nova-operator-tempest-multinode/0cbc344/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/secrets/nova/nova-api-config-data.yaml-01-nova.conf).

I have added a form of your cleanup func from https://github.com/openstack-k8s-operators/lib-common/pull/557 and added you as co-author to the commit msg. I was just not sure if you want your full name listed there as it is not shown in your github account. I can change that, just let me know.

modules/common/util/template_util.go
```diff
@@ -153,11 +230,16 @@ func lower(s string) string {
...
+	var err error
 	funcs := template.FuncMap{
-		"add":   add,
-		"lower": lower,
+		"add":                      add,
+		"execTempl":                execTempl,
+		"indent":                   indent,
+		"lower":                    lower,
+		"removeNewLines":           removeNewLines,
+		"removeNewLinesInSections": removeNewLinesInSections,
```
SeanMooney, line 240: this is proably the only concern i have with this approach.

this is a nice approach however if we want to go this route
then i think we should also allow passing extra functions via the template class

basically if we go this route i think it would make esne to enable the operator to populate template.ExtaFuncMap with functions defined in the operator.

that just avoid having to always update libcommon first and bumping the libcommon version for functions that may only make sense for a single operator.

that could be a separate PR, to make it possible for client of libcommon to pass in a function map that will be combined with the default ones provided by libcommon.

SeanMooney:
reflecting on this over the weekend i think I'm satisfied that this is a better approach to achieve this usecase, i left one comment inline but I'm mostly happy with this as proposed.

auniyal61:
lgtm
thanks for including unit tests and bar.conf as well

modules/common/util/template_util.go
stuggi: yes we can enhance this and could be used then from within the operators. but if people see the need to have additional functions, we should move them into here and can be re-used by others and not end up each of the operators create its own functions. but for specific use cases it makes sense. Lets do this in a follow up

stuggi: I have created https://issues.redhat.com/browse/OSPRH-10672 to track it

SeanMooney: ack sound good to me

SeanMooney (approved):
this looks ok to proceed with from my perspective

.pre-commit-config.yaml
gibizer: this feels a bit scary. I would the test data to normal strings with \n and \t to avoid this. But that can be done in a follow up.

gibizer (approved):

.pre-commit-config.yaml
stuggi: I have removed the spaces from the test data and returned the config.

gibizer (approved):

auniyal61:
secrets from nova-operator gate job
https://logserver.rdoproject.org/74/874/da8db1ca9e6535b95fd9567fc99a477fb8770b01/github-check/nova-operator-tempest-multinode/265ba27/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/secrets/nova/nova-cell0-conductor-config-data.yaml-01-nova.conf

auniyal61 (approved):

abays (approved):
/lgtm

olliewalsh (approved):

PR #570, opened 2024-10-14
A custom slimmed down version of VolumeSource

dprince:
Slim down VolumeSource by removing deprecated and "removed" fields from the struct. This will slim down our nested use of ExtraMounts.

Using this version of the struct shaves 38k off the cinder-operator bundle size. Using this across the board would have a significant effect on reducing the OpenStackControlplane CRD size and little if any effect on utility of the nested ExtraMount struct.


dprince:
A preview of the effect on cinder-operator is here: https://github.com/dprince/cinder-operator/commit/95e834d6f50638bcdd0f88bf623b9bb667b196a8

dprince:
@olliewalsh This is similar to the StatefulSet idea you mentioned from Rabbit.

dprince:
I removed zz_generated.deepcopy.go as the Makefile here doesn't seem to regenerate it. (just stale). Did something get broken recently?

modules/storage/storage.go
```diff
@@ -40,6 +40,84 @@ const (
...
+	// PersistentVolumeClaim in the same namespace.
+	// More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes#persistentvolumeclaims
+	// +optional
+	PersistentVolumeClaim *corev1.PersistentVolumeClaimVolumeSource `json:"persistentVolumeClaim,omitempty" protobuf:"bytes,10,opt,name=persistentVolumeClaim"`
+	// cephFS represents a Ceph FS mount on the host that shares a pod's lifetime
+	// +optional
+	CephFS *corev1.CephFSVolumeSource `json:"cephfs,omitempty" protobuf:"bytes,14,opt,name=cephfs"`
+	// flocker represents a Flocker volume attached to a kubelet's host machine. This depends on the Flocker control service being running
+	// +optional
+	Flocker *corev1.FlockerVolumeSource `json:"flocker,omitempty" protobuf:"bytes,15,opt,name=flocker"`
```
bshephar, line 73: Can we take this even further by removing additional things like `flocker`? I don't think this is supported on OCP based on my preliminary checks. But there could be a few additional things we can strip down, if we're going to go to the trouble of copying the struct, we can probably removing anything that isn't supported by OCP or tested by us.

They don't claim to support OCP either by the looks of it:
https://flocker.readthedocs.io/en/latest/supported/index.html

modules/storage/storage.go
```diff
@@ -40,6 +40,84 @@ const (
...
+	DownwardAPI *corev1.DownwardAPIVolumeSource `json:"downwardAPI,omitempty" protobuf:"bytes,16,opt,name=downwardAPI"`
+	// fc represents a Fibre Channel resource that is attached to a kubelet's host machine and then exposed to the pod.
+	// +optional
+	FC *corev1.FCVolumeSource `json:"fc,omitempty" protobuf:"bytes,17,opt,name=fc"`
+	// configMap represents a configMap that should populate this volume
+	// +optional
+	ConfigMap *corev1.ConfigMapVolumeSource `json:"configMap,omitempty" protobuf:"bytes,19,opt,name=configMap"`
+	// quobyte represents a Quobyte mount on the host that shares a pod's lifetime
+	// +optional
+	Quobyte *corev1.QuobyteVolumeSource `json:"quobyte,omitempty" protobuf:"bytes,21,opt,name=quobyte"`
```
bshephar, line 85: Same for `quobyte` I think.

bshephar (commented):
We might want to leave additional things to allow the most customer possibilities. But just floating the idea of removing a few more things if they aren't explicitly supported by OpenShift.

dprince:
@bshephar the more we can agree to remove the better.

modules/storage/storage.go
```diff
@@ -40,6 +40,84 @@ const (
...
+	StorageOS *corev1.StorageOSVolumeSource `json:"storageos,omitempty" protobuf:"bytes,27,opt,name=storageos"`
+	// csi (Container Storage Interface) represents ephemeral storage that is handled by certain external CSI drivers (Beta feature).
+	// +optional
+	CSI *corev1.CSIVolumeSource `json:"csi,omitempty" protobuf:"bytes,28,opt,name=csi"`
+	// ephemeral represents a volume that is handled by a cluster storage driver.
+	// The volume's lifecycle is tied to the pod that defines it - it will be created before the pod starts,
+	// and deleted when the pod is removed.
+	//
+	// +optional
+	Ephemeral *corev1.EphemeralVolumeSource `json:"ephemeral,omitempty" protobuf:"bytes,29,opt,name=ephemeral"`
```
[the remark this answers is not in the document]
fmount: I'm not sure if it makes sense, our use cases are currently limited to configMaps, Secrets, NFS and a few amount of other things. In theory we don't need ephemeral because I don't see a use case atm. but unless it makes the difference in terms of size, I would keep it to stay closer w/ the k8s struct.

fmount:
> I removed zz_generated.deepcopy.go as the Makefile here doesn't seem to regenerate it. (just stale). Did something get broken recently?

Not sure, but [1] should solve your problem here.

[1]  https://paste.opendev.org/show/bUqTCs84vk01M2PefMn0/

fmount:
@dprince an additional change we need to consider across all the operators that implement `extraMounts` is the way we access `corev1.Volumes`.

We usually `append( ... ` a `corev1.Volume` to an array of the same type, and we need to patch functions like [1] to work with the `Volumes` struct defined in this module (but `Make` will tell you more about this comment).

With the goal in mind of reducing the bundle size, I think we're going in the right direction, and I can follow up w/ a `glance-operator` change (because it's simple) that is based on this patch.
If everyone agrees on that we can then prepare changes for all the operators implementing this new version. 

[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/pkg/cindervolume/volumes.go#L30

modules/storage/storage.go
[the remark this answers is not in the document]
dprince: I don't have an exact case in mind but this one wasn't deprecated. Is it possible this might be useful for some partner interface/integration? The extraMounts struct is used in many of our operators, not just storage.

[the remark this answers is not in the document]
stuggi: right, there might be use cased for it. I am not aware of the need right now, but thinking of it again, it can be useful.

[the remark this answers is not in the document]
stuggi: for reference https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/1698-generic-ephemeral-volumes/README.md

dprince:
> @dprince an additional change we need to consider across all the operators that implement `extraMounts` is the way we access `corev1.Volumes`.
> 
> We usually `append( ... ` a `corev1.Volume` to an array of the same type, and we need to patch functions like [1] to work with the `Volumes` struct defined in this module (but `Make` will tell you more about this comment).
> 
> With the goal in mind of reducing the bundle size, I think we're going in the right direction, and I can follow up w/ a `glance-operator` change (because it's simple) that is based on this patch. If everyone agrees on that we can then prepare changes for all the operators implementing this new version.
> 
> [1] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/pkg/cindervolume/volumes.go#L30

I updated my cinder PR. Wanted to get general feedback on this before going too far but I think it should work out fine.

modules/storage/storage.go
dprince: flocker is gone

modules/storage/storage.go
```diff
@@ -82,3 +141,22 @@ func (v *VolMounts) Propagate(svc []PropagationType) []VolMounts {
 
 	return vl
 }
+
+// ConvertVolumeSource function to convert from a VolumeSource to a corev1.VolumeSource
+func ConvertVolumeSource(v *VolumeSource) corev1.VolumeSource {
```
stuggi, line 146: an alternative would be to do it like I did it for convert corev1.ServiceSpec to OverrideServiceSpec here https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/service/service.go#L215 . the only benefit would be that you'd not have to update this func when there is a new VolumeSource

fmount: I agree, this would be nice and a potential (future) update would be reduced to an additional field in the `VolumeSource` struct.

dprince: My latest commit to the PR adds the new function 061cc8a

stuggi (approved):
/lgtm

modules/storage/storage.go
fmount: awesome! thank you @dprince!

fmount (approved):
/lgtm

PR #572, opened 2024-10-24
[condition] Add CreateServiceReadyCondition

stuggi:
Jira: OSPRH-4164

### DIFF
--- a/modules/common/condition/conditions.go
+++ b/modules/common/condition/conditions.go
@@ -35,6 +35,9 @@ const (
 	// DBSyncReadyCondition Status=True condition when dbsync job completed ok
 	DBSyncReadyCondition Type = "DBSyncReady"
 
+	// CreateServiceReadyCondition Status=True condition when k8s service for the service created ok
+	CreateServiceReadyCondition Type = "CreateServiceReady"
+
 	// ExposeServiceReadyCondition Status=True condition when service/routes to expose the service created ok
 	ExposeServiceReadyCondition Type = "ExposeServiceReady"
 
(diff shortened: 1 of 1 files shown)


modules/common/condition/conditions.go
```diff
@@ -35,6 +35,9 @@ const (
 	// DBSyncReadyCondition Status=True condition when dbsync job completed ok
 	DBSyncReadyCondition Type = "DBSyncReady"
 
+	// CreateServiceReadyCondition Status=True condition when k8s service for the service created ok
+	CreateServiceReadyCondition Type = "CreateServiceReady"
+
 	// ExposeServiceReadyCondition Status=True condition when service/routes to expose the service created ok
 	ExposeServiceReadyCondition Type = "ExposeServiceReady"
```
gibizer, line 42: Does it mean that the the end this condition will be unused and will be deleted in a follow up lib-common PR? Or do we want to eventually have a way to indicate the the route is created for the service and the openstack service therefore really exposed?

gibizer (approved):

modules/common/condition/conditions.go
stuggi: as discussed on slack, we have a condition for the routes in the openstack-operator for this, like this
```
    - lastTransitionTime: "2024-10-22T15:42:52Z"
      message: OpenStackControlPlane nova service exposed
      reason: Ready
      status: "True"
      type: OpenStackControlPlaneExposeNovaReady
...
    - lastTransitionTime: "2024-10-26T14:33:28Z"
      message: OpenStackControlPlane Nova completed
      reason: Ready
      status: "True"
      type: OpenStackControlPlaneNovaReady
```
when everything uses this new condition, the current one will be removed.

olliewalsh (approved):

modules/common/condition/conditions.go
stuggi: actually we'll keep the current condition. in the ovn-operator,  network-attachments are used to attach the sb DB to the dataplane network. since it also involves registering those into dns  I think ExposeService is correct/better and not CreateService.

gibizer: Fine by me.

PR #576, opened 2024-10-28
Simplify and fix name issue with ToCoreVolume

dprince:


modules/storage/storage.go
```diff
@@ -167,17 +167,11 @@ func (s *VolumeSource) ToCoreVolumeSource() (*corev1.VolumeSource, error) {
...
-		return nil, fmt.Errorf("error marshalling Volume: %w", err)
-	}
-
-	err = json.Unmarshal(coreVolumeBytes, coreVolume)
-	if err != nil {
-		return nil, fmt.Errorf("error unmarshalling Volume: %w", err)
+	volSource, err := s.VolumeSource.ToCoreVolumeSource()
+	coreVolume := &corev1.Volume{
+		Name:         s.Name,
+		VolumeSource: *volSource,
```
stuggi, line 173: in case of error when *volSource is nil this will hit a nil ptr dereference error. should check for err condition before.

dprince: good call, pushed a fix for that

stuggi (approved):
/lgtm

PR #578, opened 2024-11-06
Use internal Volume struct for tls

bshephar:
This change switches the tls package over to using the lib-common copy of the Volume struct. This will allow us to append TLS volumes to the volume struct in each service Operator as we update the version of lib-common in each of them.


dprince:
seems good to me

stuggi:
I am not sure I understand the intend for this switch. the tls volumes are appended to the deployment volumes in the operators, so doesn't this change just required another conversion to the k8s volume?

fmount:
-1 to switch to `storage.Volume` instead of simply using `corev1.Volume`. We don't want to be in the business of doing a lot of conversions, and we already "convert" to `corev1.Volume` what we have with `extraMounts`.
Is there a particular reason that led to this change?

bshephar:
Yeah, my bad. This isnt' necessary

PR #580, opened 2024-11-12
Create getDNSClusterDomain

averdagu:
Openshift DNS creates it's records using the value in clusterDomain. This information could be customizable in the futre.

Currently on most of the operators they're using a hardcoded value, this function allows to all operators get this information from lib-common, and in case in the future this value is customizable we'd only need to change it here.

### DIFF
--- a/modules/common/clusterdns/clusterdns.go
+++ b/modules/common/clusterdns/clusterdns.go
@@ -1,0 +1,25 @@
+/*
+Copyright 2024 Red Hat
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package clusterdns
+
+// GetDNSClusterDomain - Return openshift dns Cluster Domain name
+func GetDNSClusterDomain() string {
+	// Currently this is hardcoded on the openshift dns operator, but
+	// this could change in the future, allowing the space to retrieve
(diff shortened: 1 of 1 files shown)


mtomaska:
LGTM

modules/common/clusterdns/clusterdns.go
```diff
@@ -0,0 +1,25 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package clusterdns
+
+// GetDNSClusterDomain - Return openshift dns Cluster Domain name
+func GetDNSClusterDomain() (string, error) {
```
stuggi, line 20: should we keep returning an error ? I guess when there is the way to check if from the cluster we also have to change the parameters which get passed into the func

averdagu: Hmmm, I guess we should return error (in order to not need to modify all operators using this function) when we modify logic and try to get value from cluster.
And I don't think we ever need to use any parameter, as openshift dns operator only accepts one object in the cluster, so there's no need to parameterizice it. [0]

I just realize that I return "value, error" but is it possible that the convention we're using is "error, value"?

[0] https://github.com/openshift/cluster-dns-operator/blob/208d50c1a5e0aaeb991366daa749abdffa803224/pkg/operator/controller/controller.go#L180-L184

stuggi: but how would you query the dns object from the cluster? we'd need to either pass in that object, or pass in  a client to be able to get it

averdagu: Oh true. So since if this code changes we'll need to update the operators you suggest to remove the error returned?

stuggi: since we have to change it anyways, yes lets remove it

averdagu: Done

stuggi (approved):
/lgtm

PR #582, opened 2024-11-19
DNM - POC: Expose Affinity and let the service operators override the default policy

fmount:


fmount:
https://github.com/openstack-k8s-operators/lib-common/pull/587 is the new approach.

PR #584, opened 2024-12-03
Add missing tags to []conditions which got reported by crd checker

stuggi:
Adds the task which got reported by the crd checker recently added as a pre-commit validation.

Also syncs the validation of the conditon parameters with [1].

In a follow up, we should integrate the new ObservedGeneration for a condition [1]. The observedGeneration represents the .metadata.generation that the condition was set based upon.

[1] https://github.com/kubernetes/apimachinery/blob/release-1.29/pkg/apis/meta/v1/types.go#L1497
[1] https://github.com/kubernetes/apimachinery/blob/release-1.29/pkg/apis/meta/v1/types.go#L1518

### DIFF
--- a/modules/common/condition/types.go
+++ b/modules/common/condition/types.go
@@ -41,17 +41,50 @@ const (
 )
 
 // Type - A summarizing name for a given condition
+// ---
+// Many .condition.type values are consistent across resources like Available, but because arbitrary conditions can be
+// useful (see .node.status.conditions), the ability to deconflict is important.
+// The regex it matches is (dns1123SubdomainFmt/)?(qualifiedNameFmt)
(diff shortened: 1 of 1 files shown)


modules/common/condition/types.go
```diff
@@ -41,17 +41,50 @@ const (
...
+//	}
 type Condition struct {
 	// Type of condition in CamelCase.
+	// +required
+	// +kubebuilder:validation:Required
 	Type Type `json:"type"`
 
-	// Status of the condition, one of True, False, Unknown.
+	// status of the condition, one of True, False, Unknown.
+	// +required
```
bogdando, line 85: as Conditions got included into CRDs, we cannot change the optional field to become required, do we?
Also, how that change passes the crd checker? Let's try to include this into some operator as a dependency and see if crd checker allows that?

stuggi:
closing in favor of https://github.com/openstack-k8s-operators/lib-common/pull/585

PR #585, opened 2024-12-03
Add missing tags to []conditions which got reported by crd checker

stuggi:
Adds the task which got reported by the crd checker recently added as a pre-commit validation.

In a follow up, we should integrate the new ObservedGeneration for a condition [1]. The observedGeneration represents the .metadata.generation that the condition was set based upon.

[1] https://github.com/kubernetes/apimachinery/blob/release-1.29/pkg/apis/meta/v1/types.go#L1518


dprince:
I think some of these validations will go into the CRD yamls thus increasing the size. want to be sure we are actually getting value out of these vs. just making a tool like crd-schema-checker happy

bogdando (commented):
this minimal scoped change looks better to me than the 584th pr

PR #587, opened 2024-12-11
Expose PodAffinity and PodAntiAffinity struct and build overrides

fmount:
This patch introduces a very basic struct to expose Pod `Affinity`/ `Antiaffinity` interfaces as part of the `Topology CR` spec provided by `infra-operator` and allows to patch the (opinionated) default affinity that is currently applied to the services.
The existing `DistributePods` function is still present in the code to ensure backward compatibility and not block lib-common bumps, but is marked as deprecated.


modules/common/affinity/affinity.go
```diff
@@ -53,4 +58,46 @@ func DistributePods(
 			},
 		},
 	}
+	// patch the default affinity Object with the data passed as input
+	if overrides != nil {
+		patchedAffinity, _ := toCoreAffinity(defaultAffinity, overrides)
```
stuggi, line 63: can we ignore the error?

fmount: mmm we could backpropagate the resulting error, I agree

fmount: Done

modules/common/affinity/affinity_test.go
```diff
@@ -51,9 +51,7 @@ func TestDistributePods(t *testing.T) {
 
 	t.Run("Default pod distribution", func(t *testing.T) {
 		g := NewWithT(t)
-
-		d := DistributePods("ThisSelector", []string{"selectorValue1", "selectorValue2"}, "ThisTopologyKey")
-
+		d, _ := DistributePods("ThisSelector", []string{"selectorValue1", "selectorValue2"}, "ThisTopologyKey", nil)
```
stuggi, line 54: can we add another test case to use the override? other then that this looks good to me. lets get also feedback from @olliewalsh

stuggi: thanks!

modules/common/affinity/affinity.go
```diff
@@ -53,4 +58,44 @@ func DistributePods(
...
+		origAffinit, err := json.Marshal(affinity)
+		if err != nil {
+			return aff, fmt.Errorf("error marshalling Affinity Spec: %w", err)
+		}
+		patch, err := json.Marshal(override)
+		if err != nil {
+			return aff, fmt.Errorf("error marshalling Affinity Spec: %w", err)
+		}
+
+		patchedJSON, err := strategicpatch.StrategicMergePatch(origAffinit, patch, v1.Affinity{})
```
fultonj, line 88: This is the main point of the patch: take the default one and merge it with the override.

modules/common/affinity/affinity_test.go
```diff
@@ -47,13 +47,82 @@ var affinityObj = &corev1.Affinity{
...
+			// is applied, while PodAffinity is the result of the override passed
+			// as input
+			PodAntiAffinity: affinityObj.PodAntiAffinity,
+			NodeAffinity:    nil,
+			PodAffinity: &corev1.PodAffinity{
+				PreferredDuringSchedulingIgnoredDuringExecution: weightedPodAffinityTermOverride,
+			},
+		}
+		affinityOverride := &OverrideSpec{
+			PodAntiAffinity: nil,
```
fultonj, line 118: Here we see that even if you set Anti to nil, that the default behavior (of adding anti) results in it still getting added.

fultonj (commented):
Thank you. Looks good to me.

stuggi:
looks good to me, @fmount just wondering if it would be easier to introduce a new func for the current `DistributePods` and basically deprecate the current one. This would allow the teams to still bump lib-common in the operators without the need to update the func call. When we then switch the operators to use the new topology way we move them to the new func and afterwards to a cleanup removing `DistributePods` when all done.

fmount:
> looks good to me, @fmount just wondering if it would be easier to introduce a new func for the current `DistributePods` and basically deprecate the current one. This would allow the teams to still bump lib-common in the operators without the need to update the func call. When we then switch the operators to use the new topology way we move them to the new func and afterwards to a cleanup removing `DistributePods` when all done.

We could do that. If it helps to not make this work the bottleneck of the entire bump process, I think we can change the `function name`, and still keep the old definition until we explicitly do the switch. We can then do some cleanup. I'll work on this one today so we can land this patch.

fmount:
> looks good to me, @fmount just wondering if it would be easier to introduce a new func for the current `DistributePods` and basically deprecate the current one. This would allow the teams to still bump lib-common in the operators without the need to update the func call. When we then switch the operators to use the new topology way we move them to the new func and afterwards to a cleanup removing `DistributePods` when all done.

Hey @stuggi I made the update we discussed.
- `DistributePods` (and the related tests) is still present, but marked as deprecated
- `DistributePodsOverride` reprensents the new content we already reviewed, and it can be used by the operators that introduce the new behavior in their Deployments/StatefulSets

This way we do not block any `lib-common` bump, and we can provide a dedicated patch that switches to the new function. In addition, we can do some cleanup at the very end of the process, which is an easy process.
Let me know your thoughts and if we're ready to land this.

fmount:
@stuggi I updated both [1] and [2] to use this new version of `lib-common`.
 I'll do a manual bump there to start using the new function after this patch lands.
(@fultonj FYI ^)

[1] https://github.com/openstack-k8s-operators/infra-operator/pull/325
[2] https://github.com/openstack-k8s-operators/glance-operator/pull/670/

modules/common/affinity/types.go
```diff
@@ -0,0 +1,36 @@
...
+// +kubebuilder:object:generate:=true
+
+package affinity
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+// OverrideSpec -
+type OverrideSpec struct {
```
ASBishop, line 26: What's the purpose of creating this `OverrideSpec` type when it's (nearly) identical to corev1 `Affinity` structure, https://pkg.go.dev/k8s.io/api/core/v1#Affinity?

fmount: I used it to distinguish in between the Affinity struct that is present in corev1 and the Override we build through the Topology [1] and we pass to:

```golang
func toCoreAffinity(
	affinity *v1.Affinity,
	override *OverrideSpec,
```
It simply helped to make the reading easier and avoid creating aliasing w/ the `corev1` struct. If it helps to reduce the amount of code involved in this patch, I can experiment to rely on the existing struct and see how it goes.

modules/common/affinity/affinity.go
```diff
@@ -17,12 +17,20 @@ limitations under the License.
...
+	"fmt"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
 )
 
 // DistributePods - returns rule to ensure that two replicas of the same selector
 // should not run if possible on the same worker node
+// Deprecated: This function is still maintained to ensure backward compatibility
```
ASBishop, line 30: Does this really need to be deprecated? That would imply every caller will be upgraded to use the new function that supports overrides, but are overrides mandatory? This function would still be useful as long as there are callers that don't need overrides.

Regardless of whether the function is deprecated, the code can be consolidated by having one function wrap the other (it works both ways):
- DistributePods() could be a wrapper that calls DistributePodsOverride() with nil for the overrides parameter
- Or DistributePodsOverride()  could call the existing DistributePods() function to obtain the default affinity, then patch in the overrides

modules/common/affinity/affinity.go
```diff
@@ -54,3 +62,78 @@ func DistributePods(
 		},
 	}
 }
+
+// DistributePodsOverride - returns rule to ensure that two replicas of the same selector
+// should not run if possible on the same worker node
+func DistributePodsOverride(
```
ASBishop, line 68: What do you think about naming this function "DistributePodsWithOverrides"? This feels more intuitive to my english speaker ear.

modules/common/affinity/affinity.go
```diff
@@ -54,3 +62,78 @@ func DistributePods(
...
+	}
+	// patch the default affinity Object with the data passed as input
+	if overrides != nil {
+		patchedAffinity, err := toCoreAffinity(defaultAffinity, overrides)
+		return patchedAffinity, err
+	}
+	return defaultAffinity, nil
+}
+
+func toCoreAffinity(
```
ASBishop, line 107: This name isn't intuitive to me. My understanding is the function basically patches one affinity with "overrides" (another affinity) and returns the result. Would using a verb like "merge" be accurate? That would lead me to a name like MergeAffinityOverrides().

modules/common/affinity/affinity.go
```diff
@@ -54,3 +62,78 @@ func DistributePods(
...
+	affinity *v1.Affinity,
+	override *OverrideSpec,
```
ASBishop, line 109: I'm still not clear why this arg needs to be an OverrideSpec rather than just another v1.Affinity structure. Maybe my eyes simply aren't detecting the difference between the two structures?

modules/common/affinity/types.go
ASBishop: I think the 'override' parameter name is sufficient to distinguish the parameters. However, I am (partially) persuaded by your argument that having our own structure definition makes it easier to modify the code if our structure needs to diverge from the v1.Affinity structure.

Thinking about it some more "OverrideSpec" is quite a generic name. How do you feel about  "AffinityOverride" for the name?

Also, as long as the two structures are essentially the same (no divergence yet) you could 'inline' the v1.Affinity structure and not have to essentially copy it.

modules/common/affinity/affinity.go
ASBishop: Replying to myself, this is discussed in another comment so no need to discuss it again, here.

modules/common/affinity/types.go
olliewalsh: while they are identical then can it just be `type OverrideSpec corev1.Affinity`

modules/common/affinity/affinity.go
fmount: > Does this really need to be deprecated? That would imply every caller will be upgraded to use the new function that supports overrides, but are overrides mandatory? This function would still be useful as long as there are callers that don't need overrides.
> 
Overrides are not mandatory, the first version of this patch had a new parameter in the signature of the `DistributePods` function, but if you simply pass `nil` it results fully backward compatible.
However, as per a conversation I had w/ @stuggi , in order to avoid blocking lib-common bumps, we can create another function, and change it to all the operators when we add support for `Topology` (see the glance-operator patch).
`DistributePods` is marked as deprecated because we want to move all the services to the new signature, but without establishing an hard timeline (so who is using `DistributePods` can still do that).

> Regardless of whether the function is deprecated, the code can be consolidated by having one function wrap the other (it works both ways):
> 
>     * DistributePods() could be a wrapper that calls DistributePodsOverride() with nil for the overrides parameter
> 
>     * Or DistributePodsOverride()  could call the existing DistributePods() function to obtain the default affinity, then patch in the overrides

We definitely can: I can use `DistributePods` to get `defaultAffinity` and not deprecate it, if it helps.
However, my long term goal/idea is to consolidate everything into a single function, and pass `nil` from the service operator in its signature in case you don't need to handle `Topology` or `Affinity` overrides.
This is an intermediate step that requires a follow up and a cleanup when we update Service operators to switch to the updated signature.

modules/common/affinity/affinity.go
fmount: +1, I'm going to update it

modules/common/affinity/affinity.go
fmount: As mentioned in my previous comment, the redundancy introduced by the struct is mainly related to make sure we control future updates to what we support in the `Topology` Spec (it includes `affinity.OverrideSpec`). We can switch to `corev1.Affinity` (there is no difference between the two structs atm), but we need to be aware that we can get updates w/o our control. 
If we're good with that implication, I can `s/OverrideSpec/corev1.Affinity/g`.

modules/common/affinity/affinity.go
ASBishop: I guess I still don't see a benefit to forcing service operators to update their code to call the new function with nil overrides. There's no real cost for lib-common to offer two functions, with and without overrides, especially when one can be a trivial wrapper around the other. But this is just my own preference, so I'm curious how other feel about this.

modules/common/affinity/types.go
ASBishop: even better!

fmount: > I think the 'override' parameter name is sufficient to distinguish the parameters. However, I am (partially) persuaded by your argument that having our own structure definition makes it easier to modify the code if our structure needs to diverge from the v1.Affinity structure.
> 
> Thinking about it some more "OverrideSpec" is quite a generic name. How do you feel about "AffinityOverride" for the name?
> 
Originally it was `AffinityOverride`, but `golint` complains when you import it elsewhere because it would be `affinity.AffinityOverride`. For this reason I turned it into `OverrideSpec`.

> Also, as long as the two structures are essentially the same (no divergence yet) you could 'inline' the v1.Affinity structure and not have to essentially copy it.

fmount: > while they are identical then can it just be `type OverrideSpec corev1.Affinity`

If we agree to rely on `corev1.Affinity` I'm inclined to remove `OverrideSpec` struct entirely. Are we in agreement with this?

modules/common/affinity/affinity.go
fmount: I'm not saying I'm going to remove `DistributePods` and force all the Service operators to pass `nil` to the new function. Topology (and affinity overrides) should be supported everywhere, and at some point every service needs to bump to the DistributePods version with the updated signature.
Not sure I'm missing your point though.

modules/common/affinity/affinity.go
fmount: > is that really necessary? Would be a lot cleaner I expect to just default to the DistributePod affinity when no topology cr is referenced, otherwise use the topology cr (which may have no affinity constraints)

It is if we want to keep the `defaultAffinity` even when it's not explicitly added in the `Topology` CR, but using the override in case it is explicitly added by patching the `defaultAffinity`. 
To be clear, this would avoid any restart for the existing controlplane (because the statefulset would be unchanged. 
If we don't like this approach, we can simply drop this entire patch, pass the `corev1.Affinity` from a `Topology` CR, and do a simple assignment. 

@olliewalsh @ASBishop  wdyt

ASBishop: I don't understand the details yet, but I think it comes down to whether the "overrides" is meant to entirely replace the defaultAffinity, or only override a portion of it.

modules/common/affinity/affinity.go
ASBishop: Oh, OK. To me the word "deprecated" suggests it might be removed in the future. Perhaps the comment isn't really necessary if the intention is to allow existing operators to keep using the function when they don't need to switch to the new one because they require overrides.

modules/common/affinity/types.go
ASBishop: Ollie's suggestion looks like a reasonable compromise. It leverages the existing v1.Affinity (no need to repeat the details in our own structure) but establishes our own type name in case we need to diverge in the future.

And argh, I hate that tools like golint downvote using names that make better sense to us humans.

modules/common/affinity/affinity.go
olliewalsh: > It is if we want to keep the defaultAffinity even when it's not explicitly added in the Topology CR

yeah, I'm asking do we want that? Doesn't seem very intuitive to the end user that the resulting podspec is a combination of what they are explicitly providing and some hardcoded internal struct. When topologyref is used I think we can assume they know what they want/what they are doing

fmount: Good, this is an important bit, which bring us in a position where we could entirely avoid this change.
If we agree that we expect to trust the user, then we can abandon this change and include `corev1.Affinity` to [1].
Given I have to push an update to [1] w/o the dependency of this change, are we (@fultonj @stuggi @ASBishop @olliewalsh) ok to expect:
1. `DistributePods` (the regular one) if no Topology is referenced in a service operator
2.  no call to DistributePods and simply a `statefulset.Spec,Template.Spec.Affinity` assignement?

If so I'd like to move the attention to the patch [1], and land it during the week, so we can move to the service operator (glance to start) patch.

[1] https://github.com/openstack-k8s-operators/infra-operator/pull/325

stuggi: > Good, this is an important bit, which bring us in a position where we could entirely avoid this change. If we agree that we expect to trust the user, then we can abandon this change and include `corev1.Affinity` to [1]. Given I have to push an update to [1] w/o the dependency of this change, are we (@fultonj @stuggi @ASBishop @olliewalsh) ok to expect:
> 
> 1. `DistributePods` (the regular one) if no Topology is referenced in a service operator

I'd say yes, especially for existing deployments, or we have a default topology which gets created per default and set to all services when the topologyref gets introduced. it should be the same as DistributePods to not result in not wanted restarts.

> 2. no call to DistributePods and simply a `statefulset.Spec,Template.Spec.Affinity` assignement?

I think it sounds good to me. if a user wants to use a specific topology they can configure the whole thing and reference it.
 
> If so I'd like to move the attention to the patch [1], and land it during the week, so we can move to the service operator (glance to start) patch.
> 
> [1] [openstack-k8s-operators/infra-operator#325](https://github.com/openstack-k8s-operators/infra-operator/pull/325)

fmount:
Summary: we reached a point where we agreed on a few points:
1. The `affinity.OverrideSpec` struct is pretty much the same of `corev1.Affinity` provided by k8s
2. We do not want to `always` set a `PodAntiAffinity` by default: if a human operator creates a `Topology` CR that should be applied to a given component, he's responsible to provide any affinity related rule
3. we continue to apply, through the service operators, the same `antiaffinity` rule if `no Topology is referenced`.

As per the above, I'm closing this PR and move the conversation to [1].

[1] https://github.com/openstack-k8s-operators/infra-operator/pull/325

PR #589, opened 2024-12-11
[nad] Add GetJsonPathFromConfig

stuggi:
Adds function to return the result of the jsonPath as string from the NetworkAttachmentDefinition config. If the NAD has no config, an empty string is returned.
The jsonPath must be in the format e.g. ".ipam"


abays (approved):
/lgtm

PR #590, opened 2024-12-13
[conditions] Reset conditions on Init()

stuggi:
Recently the operators changed to use CreateServiceReadyCondition instead of ExposeServiceReadyCondition. An update of an existing environment would result in the ExposeServiceReadyCondition to be never removed from the conditions list of a CR because the list is not reset on Init().

This adds a Reset() method and a call right at the start of Init().

Related: https://issues.redhat.com//browse/OSPRH-4164


abays (approved):
/lgtm

PR #592, opened 2025-01-16
Add Topology common condition messages

fmount:


modules/common/condition/conditions.go
```diff
@@ -82,6 +82,10 @@ const (
 
 	// TLSInputReadyCondition Status=True condition when required TLS sources are ready
 	TLSInputReadyCondition Type = "TLSInputReady"
+
+	// TopologyReadyCondition Status=True condition that indicates a CR
+	// exists and is referenced by the Service
+	TopologyReadyCondition Type = "TopologyReady"
```
fultonj, line 88: Seems consistent with other conditions and I see it getting used in glance-operator patch here:

https://github.com/openstack-k8s-operators/glance-operator/pull/670/files#diff-fe88ad2f79de701bd18c4a83352465de46564ebf02e97c1f6e6c1e0bd3f66554R842-R849

fmount:
@olliewalsh @stuggi I think this can be landed anyway, so I can reduce the dependency chain on the service operator side (glance patch).

stuggi (approved):
/lgtm

PR #593, opened 2025-01-16
[const] Adds TemplateParameters const

stuggi:
Adds TemplateParameters const, which reflects the name of the key when storing a dump of the template parameters in the secret.

Jira: https://issues.redhat.com/browse/OSPRH-13100


olliewalsh (approved):
/lgtm

PR #596, opened 2025-01-23
Add function to check if first OCP cluster network is IPv6

olliewalsh:


modules/common/ocp/ocp.go
```diff
@@ -65,3 +65,17 @@ func HasIPv6ClusterNetwork(ctx context.Context, h *helper.Helper) (bool, error)
...
 	return false, nil
 }
+
+// FirstClusterNetworkIsIPv6 - Check if first OCP cluster network is IPv6
+func FirstClusterNetworkIsIPv6(ctx context.Context, h *helper.Helper) (bool, error) {
+	networkConfig := &ocp_config.Network{}
+	err := h.GetClient().Get(ctx, types.NamespacedName{Name: "cluster", Namespace: ""}, networkConfig)
+	if err != nil {
+		return false, err
+	}
```
stuggi, line 75: since both funcs now use a the same to get the `networkConfig`, could make a dedicated func for it and re-use it. but can do that in a follow up

PR #597, opened 2025-02-04
[topology] Add optional default for topologyConstraint LabelSelector

olliewalsh:
Replicate the defaulting behaviour used in cluster-level default constraints https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#cluster-level-default-constraints

### DIFF
--- a/modules/common/labels/labels.go
+++ b/modules/common/labels/labels.go
@@ -17,6 +17,7 @@ limitations under the License.
 package labels
 
 import (
+	"github.com/openstack-k8s-operators/lib-common/modules/common"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -59,4 +60,29 @@ func GetLabels(
 	}
 
 	return util.MergeStringMaps(labelSelector, custom)
+}
+
+// GetSingleLabelSelector - create a simple label selector for a single key/value label
+func GetSingleLabelSelector(
+	key string,
+	value string,
+) metav1.LabelSelector {
+	return metav1.LabelSelector{
+		MatchExpressions: []metav1.LabelSelectorRequirement{
+			{
+				Key:      key,
+				Operator: metav1.LabelSelectorOpIn,
+				Values: []string{
+					value,
+				},
+			},
+		},
+	}
+}
+
+// GetAppLabelSelector - create a simple label selector for the default AppSelector "service" label
+func GetAppLabelSelector(
+	name string,
+) metav1.LabelSelector {
+	return GetSingleLabelSelector(common.AppSelector, name)
 }
--- a/modules/common/topology/topology.go
+++ b/modules/common/topology/topology.go
@@ -19,10 +19,12 @@ package topology
 import (
 	"context"
 	"fmt"
+
 	topologyv1 "github.com/openstack-k8s-operators/infra-operator/apis/topology/v1beta1"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
@@ -33,6 +35,7 @@ func EnsureTopologyRef(
 	h *helper.Helper,
 	topologyRef *TopoRef,
 	finalizer string,
+	defaultLabelSelector *metav1.LabelSelector,
 ) (*topologyv1.Topology, string, error) {
 
 	var err error
@@ -43,7 +46,7 @@ func EnsureTopologyRef(
 		return nil, "", fmt.Errorf("No valid TopologyRef input passed")
 	}
 
-	topology, hash, err := topologyv1.GetTopologyByName(
+	topology, _, err := topologyv1.GetTopologyByName(
 		ctx,
 		h,
 		topologyRef.Name,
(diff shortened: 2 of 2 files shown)


modules/common/topology/topology.go
```diff
@@ -52,12 +55,33 @@ func EnsureTopologyRef(
...
+	// Set default LabelSelector on topologyConstraints if not set, similar to cluster level default:
+	// https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#cluster-level-default-constraints
+	topology = topology.DeepCopy()
+
+	topologyConstraints := topology.Spec.TopologySpreadConstraints
+	if topologyConstraints != nil {
+		for i := 0; i < len(*topologyConstraints); i++ {
+			current := &(*topologyConstraints)[i]
+			if current.LabelSelector == nil {
+				current.LabelSelector = defaultLabelSelector
```
fmount, line 75: if `defaultLabelSelector` is nil (because I pass nil to this function) L75 will segfault the operator calling this function.
Similarly to what we do with `topologyConstraints` on L71, we should check in advance (likely L65) if `defaultLabelSelector` is nil, and eventually skip this block.

fmount (changes requested):

modules/common/topology/topology.go
olliewalsh: operator need to handle nil LabelSelector either way

fmount: I still think, if that parameter is `nil`, that it might be more efficient to not iterate on a `topologyConstraint` list and check every `.LabelSelector`, but hopefully we do not expect big lists, which may downgrade my previous comment to a nit if you tested and everything looks ok.

fmount: After landing https://github.com/openstack-k8s-operators/glance-operator/pull/670 we can rebase `glance-operator` to incorporate this change

fmount: I tested this locally (simply passing `nil` to the function to meet the signature) and I don't see big problems other the inefficient iteration that assigns `nil` to an already `nil` value in case a `defaultLabelSelector` is not provided.
I'll do a couple of more tests tomorrow with fresh eyes, but I'm planning to land this tomorrow if you think that we don't have to handle that case and skip the iteration.

fmount:
@olliewalsh tested and works as expected (no regressions in the previous code that we tested). If you can take a look at https://github.com/openstack-k8s-operators/glance-operator/pull/670#issuecomment-2635842202 I was wondering if similarly to what we did for `DistributePods` [1], we can make the following:

```golang
// GetGlanceAPILabelSelector - Returns a metav1.LabelSelector for a given GlanceAPI
func GetGlanceAPILabelSelector(instance *glancev1.GlanceAPI) metav1.LabelSelector {
	return metav1.LabelSelector{
		MatchExpressions: []metav1.LabelSelectorRequirement{
			{
				Key:      glance.GlanceAPIName,
				Operator: metav1.LabelSelectorOpIn,
				Values: []string{
					fmt.Sprintf("%s-%s-%s", glance.ServiceName, instance.APIName(), instance.Spec.APIType),
				},
			},
		},
	}
}
``` 

a generic one, that can be called from this module.
This way we can avoid duplicating it across the operators, and we can provide 
```
common.AppSelector, []string{<service>.ServiceName`}
```
as input for that function.
You could then call it with something like `topology.GetDefaultTopologySelector(common.AppSelector, []string{<service>.ServiceName`}) and pass it down to `topology.EnsureTopologyRef(`.

Do you want to push an update for this before we land it?

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/pkg/glanceapi/statefulset.go#L296
[2] https://github.com/olliewalsh/glance-operator/commit/8386bff9a356bb8ff88a7cfa0b274b400c26b3b5

modules/common/topology/topology.go
```diff
@@ -52,12 +55,33 @@ func EnsureTopologyRef(
...
 			return topology, hash, err
 		}
 	}
+
+	// Set default LabelSelector on topologyConstraints if not set, similar to cluster level default:
+	// https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#cluster-level-default-constraints
+	topology = topology.DeepCopy()
+
+	topologyConstraints := topology.Spec.TopologySpreadConstraints
+	if topologyConstraints != nil {
```
stuggi, line 71: could be `if topologyConstraints != nil && defaultLabelSelector != nil` ?

modules/common/topology/topology.go
```diff
@@ -52,12 +55,33 @@ func EnsureTopologyRef(
...
+		for i := 0; i < len(*topologyConstraints); i++ {
+			current := &(*topologyConstraints)[i]
+			if current.LabelSelector == nil {
+				current.LabelSelector = defaultLabelSelector
+			}
+		}
+	}
+
+	hash, err = util.ObjectHash(topology.Spec)
```
stuggi, line 80: its probably not really expensive, but we can already get the hash on the `topologyv1.GetTopologyByName` (as it was), so we could move this inside the condition to only update the hash and calc it again when we iterate over it and there is an update.

olliewalsh: or could just change GetTopologyByName to not calculate the hash

modules/common/topology/topology.go
olliewalsh: sure, not at all worried about performance here though. Typically will have one or two topologyConstraints. Much bigger performance concern I expect is how slow k8s scheduler gets if there are many.

modules/common/topology/topology.go
stuggi: either way would work if callers do not expect GetTopologyByName to return a hash

modules/common/labels/labels.go
```diff
@@ -60,3 +61,28 @@ func GetLabels(
...
+				Operator: metav1.LabelSelectorOpIn,
+				Values: []string{
+					value,
+				},
+			},
+		},
+	}
+}
+
+// GetAppLabelSelector - create a simple label selector for the default AppSelector "service" label
```
fmount, line 83: @olliewalsh I don't fully understand this: why we can't simply call `GetSingleLabelSelector(common.AppSelector, value)` from an operator? `common.AppSelector` is something imported everywhere, and already available when building labels. I think we don't need `GetAppSelector` as wrapper for a function where one of the parameters is a const available in the service operators.

olliewalsh: Aren't both just convenience functions?
Almost all consumers can just call `GetAppLabelSelector(instance.name)`, special cases that don't use the service label can use `GetSingleLabelSelector("foo", "bar")`

fmount: ack sounds good to have something like this

fmount (approved):
/lgtm

PR #606, opened 2025-02-26
Add GetLabelSelector utility function

fmount:
This function is used as utility to get the list of `matchLabels` used as `selector` for `Deployments` and `StatefulSet`. Most of the service operators use to pass a `serviceLabels map[string]string`, which is available in the controller(s). This function translates the data structure into a `metav1.LabelSelector` that we can easily plug to the existing `ensureTopology` logic provided by infra-operator [1].

[1] https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/topology/v1beta1/topology_types.go#L126

### DIFF
--- a/modules/common/labels/labels.go
+++ b/modules/common/labels/labels.go
@@ -86,3 +86,14 @@ func GetAppLabelSelector(
 ) metav1.LabelSelector {
 	return GetSingleLabelSelector(common.AppSelector, name)
 }
+
+// GetLabelSelector - utility function that returns a metav1.LabelSelector
+// based on the map[string]string that represents the k/v list passed to the
+// StatefulSet or Deployment as labelSelector
+func GetLabelSelector(
+	serviceLabels map[string]string,
+) metav1.LabelSelector {
+	return metav1.LabelSelector{
+		MatchLabels: serviceLabels,
+	}
+}


fmount:
@olliewalsh @stuggi  this is Glance [1] and it might really result in a quick update across the operators that is worth doing before the freeze. I analyzed the context a bit (see [2]) and we could do this and replace all the calls that get a Single label (service, component or anything else) with:

``
labels.GetLabelSelector(serviceLabels)
``
as I did in [3]. Let's discuss and have some feedback and decide what to do.

[1] https://paste.opendev.org/show/bc3Zd04FEd2E7rEUZPdn/
[2] https://paste.opendev.org/show/bpJJisuGag3C9GQMkFrX/
[3] https://github.com/openstack-k8s-operators/glance-operator/pull/693

modules/common/labels/labels.go
```diff
@@ -86,3 +86,25 @@ func GetAppLabelSelector(
...
+			Key:      key,
+			Operator: metav1.LabelSelectorOpIn,
+			Values: []string{
+				value,
+			},
+		}
+		lsr = append(lsr, item)
+	}
+	return metav1.LabelSelector{
+		MatchExpressions: lsr,
```
olliewalsh, line 108: I don't think this is necessary, can just do: 
`
metav1.LabelSelector {
    MatchLabels: serviceLabels
}
`

fmount: Ah thank you, that is definitely simpler and should have the same behavior. Updating the patch! Thank you!

fmount: Done

fmount: I can see the statefulSet being updated with:

```
...
      topologySpreadConstraints:
      - labelSelector:
          matchLabels:
            component: glance-api
            glanceAPI: glance-default-internal
            owner: glance-default-internal
            service: glance
```

stuggi (approved):
/lgtm

PR #609, opened 2025-03-07
Add check for ExtraMount duplicate checking

bshephar:
If an ExtraMount is provided with two mountPaths that are the same, the creation of Kubernetes objects will be blocked. This function will allow us to verify early before we try to create the Kubernetes objects. Such as in webhook validations.

### DIFF
--- a/modules/storage/storage.go
+++ b/modules/storage/storage.go
@@ -20,6 +20,7 @@ package storage
 import (
 	"encoding/json"
 	"fmt"
+	"slices"
 
 	corev1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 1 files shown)


fmount:
The point is: k8s would block you by default and it returns a pretty nice verbose message because you're mounting something twice. In other words, we rely on k8s to provide this kind of check.
At that point you can go back to your CR and fix it. Is there a particular use case I'm missing?

fmount:
however I think I see an interesting use case

> The point is: k8s would block you by default and it returns a pretty nice verbose message because you're mounting something twice. In other words, we rely on k8s to provide this kind of check. At that point you can go back to your CR and fix it. Is there a particular use case I'm missing?

Thinking a little bit more about this, I recognize that there might be an opportunity to improve webhooks and call this function in advance to analyze what has been passed through extraMounts. Is your idea going to that direction?

PR #611, opened 2025-03-20
Allow setting updateStrategy on DaemonSet

averdagu:
Currently on ovn-operators we may need to modify the updateStrategy for the daemonset, this is currently not supported, as if the updateStrategy is modified on the daemonset, this won't be reflected on the openshift cluster

Related: OSPRH-11636


stuggi (approved):
/lgtm

PR #617, opened 2025-04-06
Fix ListsMustHaveSSATags raised by crd-schema-checker

fmount:
In openstack-operator, where we bump all the sub components CRDs, we currently get an error thrown by crd-schema-checker:

```
> ERROR: "ListsMustHaveSSATags": crd/openstackcontrolplanes.core.openstack.org version/v1beta1 field/^.spec.keystone.template.extraMounts must set x-kubernetes-list-type
> ERROR: "ListsMustHaveSSATags": crd/openstackcontrolplanes.core.openstack.org version/v1beta1 field/^.spec.keystone.template.extraMounts[*].extraVol must set x-kubernetes-list-type
...
```

This patch adds the `listType` to the storage module extraMounts to make the check happy [1].

[1] https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/555-server-side-apply/README.md#lists

### DIFF
--- a/modules/storage/storage.go
+++ b/modules/storage/storage.go
@@ -108,13 +108,16 @@ type Volume struct {
 type VolMounts struct {
 	// +kubebuilder:validation:type={PropagationEverywhere}
 	// Propagation defines which pod should mount the volume
+	// +listType=atomic
 	Propagation []PropagationType `json:"propagation,omitempty"`
 	// Label associated to a given extraMount
 	// +kubebuilder:validation:Optional
 	ExtraVolType ExtraVolType `json:"extraVolType,omitempty"`
 	// +kubebuilder:validation:Required
+	// +listType=atomic
 	Volumes []Volume `json:"volumes"`
 	// +kubebuilder:validation:Required
+	// +listType=atomic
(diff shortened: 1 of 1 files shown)


fmount:
I still have to check/test this and see if there are missing annotations, but it's related to https://github.com/openstack-k8s-operators/openstack-operator/pull/1395

bshephar (approved):

bshephar:
> I still have to check/test this and see if there are missing annotations, but it's related to [openstack-k8s-operators/openstack-operator#1395](https://github.com/openstack-k8s-operators/openstack-operator/pull/1395)

The fix looks good, but I think we also need to add it to the top level `extraMounts` object too right? I started looking into it here before I saw that you had already proposed a PR:
https://github.com/openstack-k8s-operators/keystone-operator/pull/562

fmount:
I don't know if right now we could benefit from this. In addition this patch does not entirely solve the problem, as we might need to unpack the most inner struct and add the annotation at that level (see this [1] as an example).
The other thing I'm worried about is adding more size to an already bit CRD w/o having much benefit.

[1] https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L973

PR #618, opened 2025-04-11
[18.0-fr2][deployment/statefulset] add IsReady func

stuggi:
Adds IsRead() for deployment and statefulset which can be used to validate when a deployment for the current Generation is fully provisioned and e.g. no rollout in progress.

Jira: OSPRH-14472


abays (approved):
/lgtm

PR #621, opened 2025-04-16
[envTest] Add MountPath check in AssertVolumeMountExists

fmount:
We recently introduced `extraMounts` `envTests` in several operators, and they rely on the common `helpers/volumes` `Assert` function to check if the resulting `Deployment` has the expected `Volume/VolumeMount`. This patch enhances `AssertVolumeMountExists` to either match a `subPath` or a regular `MountPath`. It depends on what is passed as input, and it keeps backward compatibility to what has been tested, without requiring a signature bump to cover `extraMounts`.

Jira: https://issues.redhat.com/browse/OSPRH-15839


fmount:
@stuggi I'm using https://github.com/openstack-k8s-operators/manila-operator/pull/410 to test this change. 
As per my local testing, all the `TLS` related asserts based on `subPath` continue to work, and we expanded the match to `MountPath` (in case of extraMounts, where is less common to have a `subPath`).

fmount:
> instead of checking mount and sub path with one parameter, we probably should have parameters to check both, e.g. for when both are used?
> 
> ```
>     - mountPath: /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
>       name: combined-ca-bundle
>       readOnly: true
>       subPath: tls-ca-bundle.pem
> ```

This was my first idea, but it means to change the signature of the function or simply provide a new version of it. Do you agree with creating a new version of this function so we can eventually deprecate the previous one? If we change the signature of the existing function it will result in a massive change during the next `lib-common/modules/common` bump.

stuggi:
> > instead of checking mount and sub path with one parameter, we probably should have parameters to check both, e.g. for when both are used?
> > ```
> >     - mountPath: /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
> >       name: combined-ca-bundle
> >       readOnly: true
> >       subPath: tls-ca-bundle.pem
> > ```
> 
> This was my first idea, but it means we either need to change the signature of the existing function or simply provide a new version of it. Do you agree with creating a new version of this function so we can eventually deprecate the previous one? If we change the signature of the existing function it will result in a massive change during the next `lib-common/modules/common` bump.

yes, lets create a new one and users can migrate to the new one when they see the need to check for the mountPath.

fmount:
testing the new version of this function via: https://github.com/openstack-k8s-operators/manila-operator/pull/410

fmount:
Some early feedback from local testing:

1. Empty volume Name: `th.AssertVolumeMountPathExists("",ManilaCephExtraMountsPath, "", container.VolumeMounts)`

```
  [FAILED] Volume name cannot be empty
  Expected
      <bool>: false
  to be true
```

2. Empty `mountPath` and `subPath`: `th.AssertVolumeMountPathExists(ManilaCephExtraMountsSecretName, "", "", container.VolumeMounts)`

```
  [FAILED] Both mountPath and subPath are empty
  Expected
      <bool>: false
  to be true
```

3. Empty `[]corev1.VolumeMounts{}`: `th.AssertVolumeMountPathExists(ManilaCephExtraMountsSecretName, ManilaCephExtraMountsPath, "", []corev1.VolumeMount{})`

```
  [FAILED] Volume mounts list is empty
  Expected
      <bool>: false
  to be true
```

All the other existing envTests cover the rest of the logic (matching `subPath`, `mountPath`, or both).

modules/common/test/helpers/volumes.go
```diff
@@ -50,3 +50,80 @@ func (tc *TestHelper) AssertVolumeMountExists(name string, subPath string, volum
...
+	// Early return if both mountPath and subPath are empty strings
+	if mountPath == "" && subPath == "" {
+		gomega.Expect(false).To(gomega.BeTrue(), "Both mountPath and subPath are empty")
+		return false
+	}
+
+	// init two bool variables to represent the status of mountPath and
+	// subPath check: when the function reaches this point we have at least an
+	// input that should be checked
+	mountPathExists := false
```
stuggi, line 92: could simplify with the below condition to `mountPathExists := mountPath == ""` ?

modules/common/test/helpers/volumes.go
```diff
@@ -50,3 +50,80 @@ func (tc *TestHelper) AssertVolumeMountExists(name string, subPath string, volum
...
+		subPathExists = true
+	}
+
+	for _, v := range volumeMounts {
+		// only process the current volumeMount if the name
+		// mathes
+		if v.Name == name {
+			// mountPath is passed as input and it has not
+			// been evaluated
+			if mountPath != "" && !mountPathExists {
```
stuggi, line 113: do we again have to check `mountPath != ""` ? we already did that above.

modules/common/test/helpers/volumes.go
```diff
@@ -50,3 +50,80 @@ func (tc *TestHelper) AssertVolumeMountExists(name string, subPath string, volum
...
+				}
+			}
+			// subPath is passed as input and it has not
+			// been evaluated
+			if subPath != "" && !subPathExists {
+				if v.SubPath == subPath {
+					subPathExists = true
+				}
+			}
+		}
```
stuggi, line 125: for the call of this assert, do we expect that the mountPath and the subPath are for the same entry, or do we want to cover the situation in different entries, like `mountPath: /var/lib/config-data/default` and ` subPath: keystone-api-config.json`

```
    - mountPath: /var/lib/config-data/default
      name: config-data
    - mountPath: /var/lib/kolla/config_files/config.json
      name: config-data
      readOnly: true
      subPath: keystone-api-config.json
```

if the assert is for checking `mountPath` and `subPath` for the same entry in the list, we could probably break if one of them is true?

modules/common/test/helpers/volumes.go
fmount: right, we do not need this and the nested if can be merged with this one actually.

fmount: Done.

modules/common/test/helpers/volumes.go
fmount: Done

modules/common/test/helpers/volumes.go
fmount: ok so we start the iteration if the function sees the right input (and at least we execute one loop cycle).
For each iteration, if we that both bools are set, we break the loop.  
@stuggi I'm resolving the comment but feel free to reopen if I missed something.

stuggi (approved):
/lgtm

PR #624, opened 2025-04-21
Add GetDefaultMemcachedSpec test helper

ratailor:
This change adds a GetMemcachedSpec test helper which returns memcachedv1.MemcachedSpec. This reduces code duplication and can be used in most of the other openstack operators as well.

### DIFF
--- a/modules/common/test/helpers/common.go
+++ b/modules/common/test/helpers/common.go
@@ -29,7 +29,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
+	memcachedv1 "github.com/openstack-k8s-operators/infra-operator/apis/memcached/v1beta1"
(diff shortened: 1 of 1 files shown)


stuggi:
is it to reduce something like https://github.com/openstack-k8s-operators/keystone-operator/blob/main/tests/functional/keystoneapi_controller_test.go#L103 ?  memchached test helpers should go into https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/test/helpers/memcached.go

ratailor:
> is it to reduce something like https://github.com/openstack-k8s-operators/keystone-operator/blob/main/tests/functional/keystoneapi_controller_test.go#L103 ? memchached test helpers should go into https://github.com/openstack-k8s-operators/infra-operator/blob/main/apis/test/helpers/memcached.go

Yes. I will abandon this change and submit new PR for in infra-operator for the same.
Thanks for your suggestion.

stuggi:
closing this. a helper was merged in the infra-op

PR #627, opened 2025-05-02
Add operator package

hjensas:
Add operators package, implements function setOperatorOptions which takes a pointer to controller options parses env vars and set's options.

Jira: OSPRH-16335


stuggi (approved):
/lgtm

PR #628, opened 2025-05-08
Add more annotations related functions

fmount:
This patch improves the existing `annotations` module and provides a generic mechanism to get a particular `value` associated to a `key` from the annotations `map[string]string`. It also provides utilities to parse boolean based annotation values, used by service operators to set flags on their CR.

Jira: https://issues.redhat.com/browse/OSPRH-16261

### DIFF
--- a/modules/common/annotations/network.go
+++ b/modules/common/annotations/network.go
@@ -19,6 +19,7 @@ package annotations
 import (
 	"encoding/json"
 	"fmt"
+	"strconv"
 )
 
 // NetworkAttachmentAnnot pod annotation for network-attachment-definition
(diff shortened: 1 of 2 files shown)


modules/common/annotations/network.go
```diff
@@ -54,3 +55,44 @@ func GetNADAnnotation(namespace string, nads []string) (map[string]string, error
...
+	var err error
+	// Try to get the value associated to the annotation key
+	if value, exists = GetAnnotation(ann, key); exists {
+		result, err = strconv.ParseBool(value)
+		if err != nil {
+			// the annotation is not a valid boolean, return an error
+			return false, exists, err
+		}
+	}
+	return result, exists, nil
```
stuggi, line 97: I guess we could simplify to this, or am I missing a case?
```
value, exists := ann[key] 
if !exists {
    return false, false, nil
}

result, err := strconv.ParseBool(value)
if err != nil {
	return false, true, err
}

return result, true, nil
```

modules/common/annotations/network.go
```diff
@@ -54,3 +55,44 @@ func GetNADAnnotation(namespace string, nads []string) (map[string]string, error
...
+// take actions based on the existance of the entry in the annotation map[string]string
+func GetAnnotation(
+	ann map[string]string,
+	key string,
+) (string, bool) {
+	if value, ok := ann[key]; ok {
+		return value, ok
+	}
+	return "", false
+}
```
stuggi, line 72: do we need this func? isn't it the same as `return value, exists := ann[key]`

modules/common/annotations/network.go
fmount: I assume the current version is equivalent to what you have there. I simply don't do an explicit `!exists` check because the variable is already set by `GetAnnotation` that returns true or false (the last return covers the !exists part)

modules/common/annotations/network.go
fmount: I added this because someone might want to not use this and not the one that checks the boolean and get an additional `exists` information through a bool.
If you think we don't need it we can remove it, it is just a utility that might be helpful to be wrapped and build something else around it.

fmount: Removed

modules/common/annotations/network.go
fmount: Done

stuggi (approved):
/lgtm

PR #631, opened 2025-06-05
Add NotificationBusInstance condition messages

fmount:
Because we plan to consume notificationBusInstance parameters across the board, it might be a good idea to have common conditions in the shared module.

Jira: https://issues.redhat.com/browse/OSPRH-15845

### DIFF
--- a/modules/common/condition/conditions.go
+++ b/modules/common/condition/conditions.go
@@ -68,6 +68,9 @@ const (
 	// RabbitMqTransportURLReadyCondition Status=True condition which indicates if the RabbitMQ TransportURLUrl is ready
 	RabbitMqTransportURLReadyCondition Type = "RabbitMqTransportURLReady"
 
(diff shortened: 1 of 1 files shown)


modules/common/condition/conditions.go
```diff
@@ -328,6 +328,21 @@ const (
 	// RabbitMqTransportURLReadyErrorMessage
 	RabbitMqTransportURLReadyErrorMessage = "RabbitMqTransportURL error occured %s"
 
+	//
+	// NotificationBusInstanceReady condition messages
+	//
+	// NotificationBusInstanceReadyInitMessage
+	NotificationBusInstanceReadyInitMessage = "NotificationBusInstance not started"
```
stuggi, line 335: should we also have a dedicated condition for it, like the `RabbitMqTransportURLReadyCondition` ?

fmount: makes sense, I think I missed that one! good catch. Adding in my next PS.

PR #641, opened 2025-09-24
Add INI config utility to common/util module

fmount:
This patch introduces a new `INI` manipulation utility with the ability to process/extend service configurations (`customServiceConfig`) by adding `key/value` pairs to specific sections. It includes comprehensive test coverage for various scenarios, including section detection, key validation, and edge cases. This utility does **not** replace existing keys if they are already specified via user input, **nor** does it add a new section if one does not already exist in the specified configuration. This design preserves the original purpose of `customServiceConfig` as user input.

Related: https://issues.redhat.com/browse/OSPRH-14309


modules/common/util/ini.go
```diff
@@ -0,0 +1,103 @@
...
+			sectionIndex = idx
+			// increment the index (as an offset) only when a section is found
+			if sectionName == customServiceConfigExtend.Section {
+				index = idx + 1
+			}
+		}
+		// Check if key already exists in target section
+		if token == customServiceConfigExtend.Key && sectionIndex > -1 &&
+			sectionName == customServiceConfigExtend.Section {
+			errMsg := fmt.Errorf("%w: key %s in section %s", ErrKeyAlreadyExists, token, sectionName)
```
stuggi, line 86: the openstack config is not strictly speaking ini config where you always have only once a parameter (key) in a section. there are services where it is valid to have multiple keys with the same name. e.g. in nova for pci passthrough, like the pci_passthrough_alias parameter.

fmount:
@stuggi do you think this patch is good to go now? I think I enhanced the existing code to make sure we can decide to allow duplicated keys (and I added a couple of tests based on the Nova documentation). Let me know if you think we need more changes!

fmount:
@stuggi are we good with this? let me know if this is a good time to land it, so we can take advantage of the regular (weekend) bump to update the glance patch

stuggi:
yeah, lets get it in. just wondering if the same use case could be achieved by using an additional template snippet?

PR #648, opened 2025-10-06
Add RandomString function

Deydra71:
RandomString generated a random string of specified length


stuggi:
I think you could just use https://github.com/kubernetes/apimachinery/blob/master/pkg/util/rand/rand.go#L98 ?


```
import (
	"fmt"
	"k8s.io/apimachinery/pkg/util/rand"
)

func main() {
	// Generate a random string of length 10
	randomSuffix := rand.String(10)
```

Deydra71:
@stuggi  you're right ofc, that's perfectly sufficient :))

PR #678, opened 2026-03-26
Add helpers for StatefulSet container merging, raw ConfigMaps, and preserved Secrets

lmiccini:
Three new utilities to reduce boilerplate in operator controllers:

- statefulset.MergeContainersByName: merges container specs by name, preserving server-defaulted fields (TerminationMessagePath, ImagePullPolicy, etc.) to avoid unnecessary reconcile loops. StatefulSet.CreateOrPatch now uses this instead of full Template replacement.

- configmap.CreateOrPatchRawConfigMap: creates/patches a ConfigMap from raw map[string]string data without template rendering machinery.

- secret.CreateOrPatchSecretPreserve: creates a Secret on first call and preserves existing Data on subsequent reconciles, useful for generated credentials that should not be rotated.

### DIFF
--- a/modules/common/configmap/configmap.go
+++ b/modules/common/configmap/configmap.go
@@ -199,6 +199,50 @@ func EnsureConfigMaps(
 	return nil
 }
 
+// CreateOrPatchRawConfigMap creates or patches a ConfigMap from raw data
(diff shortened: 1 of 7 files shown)


stuggi (approved):
/lgtm

PR #682, opened 2026-04-09
util: Add MultiTemplateDir to support multi-segment path in template

amartyasinha:
A new parameter MultiTemplateDir is introduced in Template struct. It takes the multi-segment path which we have inside templates dir and then continues to execute the templates. Earlier, only single- segment path (kind) was supported through InstanceType.

Follow-up nova-operator PR with the usage of MultiTemplateDir: https://github.com/openstack-k8s-operators/nova-operator/pull/1096

### DIFF
--- a/modules/common/util/template_util.go
+++ b/modules/common/util/template_util.go
@@ -59,6 +59,7 @@ type Template struct {
 	ConfigOptions      map[string]interface{} // map of parameters as input data to render the templates
 	SkipSetOwner       bool                   // skip setting ownership on the associated configmap
 	Version            string                 // optional version string to separate templates inside the InstanceType/Type directory. E.g. placementapi/config/18.0
+	MultiTemplateDir   string                 // templates dir for multi-group operators, e.g. nova/api; requires InstanceType to be set
 }
 
 // GetTemplatesPath get path to templates, either running local or deployed as container
@@ -83,20 +84,21 @@ func GetTemplatesPath() (string, error) {
 
 // GetAllTemplates - get all template files
 //
-// The structur of the folder is, base path, the kind (CRD in lower case),
+// The structure of the folder is, base path, subdir, templateType, version
 //   - path - base path of the templates folder
-//   - kind - sub folder for the CRDs templates
+//   - subdir - directory under that root: one segment (legacy InstanceType, e.g. NovaAPI) or
+//     multi-segment (e.g. nova/api from MultiTemplateDir)
 //   - templateType - TType of the templates. When the templates got rendered and added to a CM
 //     this information is e.g. used for the permissions they get mounted into the pod
 //   - version - if there need to be templates for different versions, they can be stored in a version subdir
 //
 // Sub directories inside the specified directory with the above parameters get ignored.
-func GetAllTemplates(path string, kind string, templateType string, version string) []string {
+func GetAllTemplates(path string, subdir string, templateType string, version string) []string {
 
-	templatePath := filepath.Join(path, strings.ToLower(kind), templateType, "*")
+	templatePath := filepath.Join(path, subdir, templateType, "*")
 
 	if version != "" {
-		templatePath = filepath.Join(path, strings.ToLower(kind), templateType, version, "*")
+		templatePath = filepath.Join(path, subdir, templateType, version, "*")
 	}
 
 	templatesFiles, err := filepath.Glob(templatePath)
@@ -293,8 +295,24 @@ func GetTemplateData(t Template) (map[string]string, error) {
 	data := make(map[string]string)
 
 	if t.Type != TemplateTypeNone {
-		// get all scripts templates which are in ../templesPath/cr.Kind/CMType/<OSPVersion - optional>
-		templatesFiles := GetAllTemplates(templatesPath, t.InstanceType, string(t.Type), string(t.Version))
+		// If MultiTemplateDir is set but InstanceType is not, return an error
+		// though we do not use InstanceType here, but it is used in secret.go
+		if t.MultiTemplateDir != "" && t.InstanceType == "" {
+			return nil, ErrInstanceTypeUnsetWithMultiTemplateDir
+		}
+
+		var templateSubdir string
+		// if MultiTemplateDir is set, it will take precedence over InstanceType
+		// otherwise, InstanceType is used to create the template subdir based on CRD name
+		if t.MultiTemplateDir != "" {
+			templateSubdir = t.MultiTemplateDir
+		} else {
+			templateSubdir = strings.ToLower(t.InstanceType)
+		}
+		if templateSubdir == "" {
+			return nil, fmt.Errorf("%w: type is %q", ErrTemplateSubdirUnset, t.Type)
+		}
+		templatesFiles := GetAllTemplates(templatesPath, templateSubdir, string(t.Type), string(t.Version))
 
 		// render all template files
 		for _, file := range templatesFiles {
--- a/modules/common/util/template_util_test.go
+++ b/modules/common/util/template_util_test.go
@@ -300,14 +300,14 @@ func TestGetAllTemplates(t *testing.T) {
 
 	tests := []struct {
 		name     string
-		kind     string
+		subdir   string
 		tmplType TType
 		version  string
 		want     []string
 	}{
 		{
 			name:     "Get TemplateTypeConfig templates with no version",
-			kind:     "testservice",
+			subdir:   "testservice",
 			tmplType: TemplateTypeConfig,
 			version:  "",
 			want: []string{
@@ -318,7 +318,7 @@ func TestGetAllTemplates(t *testing.T) {
 		},
 		{
 			name:     "Get TemplateTypeScripts templates with version",
-			kind:     "testservice",
+			subdir:   "testservice",
 			tmplType: TemplateTypeScripts,
 			version:  "1.0",
 			want: []string{
@@ -334,7 +334,7 @@ func TestGetAllTemplates(t *testing.T) {
 			p, _ := GetTemplatesPath()
 			g.Expect(p).To(BeADirectory())
 
-			templatesFiles := GetAllTemplates(p, tt.kind, string(tt.tmplType), tt.version)
+			templatesFiles := GetAllTemplates(p, tt.subdir, string(tt.tmplType), tt.version)
 
 			g.Expect(templatesFiles).To(HaveLen(len(tt.want)))
 			g.Expect(templatesFiles).Should(HaveEach(BeARegularFile()))
@@ -366,7 +366,7 @@ func TestGetTemplateData(t *testing.T) {
 				Name:         "testservice",
 				Namespace:    "somenamespace",
 				Type:         TemplateTypeConfig,
-				InstanceType: "testservice",
+				InstanceType: "TestService",
 				Version:      "",
 				ConfigOptions: map[string]interface{}{
 					"ServiceUser": "foo",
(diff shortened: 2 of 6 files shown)


modules/common/util/template_util.go
```diff
@@ -81,22 +82,42 @@ func GetTemplatesPath() (string, error) {
...
+// groupSegmentAndKindSuffix returns the first dot-separated segment of apiGroup and
+// kind prefix to ensure the correct template dir is used.
+func groupSegmentAndKindSuffix(apiGroup, kind string) (groupSegment, kindSuffix string) {
+	kind = strings.ToLower(strings.TrimSpace(kind))
+	apiGroup = strings.TrimSpace(apiGroup)
+	if apiGroup == "" {
+		return "", kind
+	}
+	groupSegment = strings.ToLower(strings.Split(apiGroup, ".")[0])
+	return groupSegment, strings.TrimPrefix(kind, groupSegment)
```
bogdando: this does like ``nova.openstack.org + NovaAPI → nova/api``. I would avoid splitting the words boundaries based on its structure implications

amartyasinha: Couldn't find any value which would returns the core service name (nova, cinder, etc). The closest I could find was group. Is there any other value which I can use to avoid splitting based on structure implications?

modules/common/secret/secret.go
```diff
@@ -269,7 +269,13 @@ func createOrUpdateSecret(
...
 		} else {
 			// Set ownership labels that can be found by the respective controller kind
-			ownerLabel := fmt.Sprintf("%s.%s", strings.ToLower(st.InstanceType), obj.GetObjectKind().GroupVersionKind().Group)
+			var ownerSubdir string
+			if md := strings.TrimSpace(st.MultiTemplateDir); md != "" {
+				ownerSubdir = md
+			} else {
+				ownerSubdir = strings.TrimSpace(st.InstanceType)
+			}
+			ownerLabel := fmt.Sprintf("%s.%s", strings.ToLower(ownerSubdir), obj.GetObjectKind().GroupVersionKind().Group)
```
bogdando: ``MultiTemplateDir`` may contain `/` leading to unexpected labels, like `nova/api.nova.openstack.org`

modules/common/util/template_util_test.go
```diff
@@ -382,6 +382,40 @@ func TestGetTemplateData(t *testing.T) {
...
 			error: false,
 		},
+		{
+			name: "Render TemplateTypeConfig using MultiTemplateDir ignores InstanceType",
+			tmpl: Template{
+				Name:             "testservice",
+				Namespace:        "somenamespace",
+				Type:             TemplateTypeConfig,
+				InstanceType:     "wrong-unused",
+				MultiTemplateDir: "testservice",
```
bogdando, line 392: add a test case with values like `nova/api`

modules/common/util/template_util.go
bogdando: a new data field may be introduced for that purpose

modules/common/secret/secret.go
amartyasinha: agree! reverted to keep using InstanceType. Due to this, I've made InstanceType required whenever MultiTemplateDir is used to not break the existing logic.

modules/common/util/template_util_test.go
amartyasinha: done, added a generic `test/service` testcase.

modules/common/util/template_util.go
amartyasinha: looking how to do that

amartyasinha: @bogdando WDYT about adding new spec struct in nova-operator (and other operators too if they want to support multi-group structure). Or if you think it'd be overkill and we should just add const values.

```go
type OpenStackServiceIdentity struct {
	// +kubebuilder:validation:Optional
	// OpenstackService is the core OpenStack service name (e.g. nova, placement).
	OpenstackService string `json:"openstackService,omitempty"`
	// +kubebuilder:validation:Optional
	// ServiceComponent is the component within that service (e.g. api, compute).
	ServiceComponent string `json:"serviceComponent,omitempty"`
}
```

```yaml
apiVersion: nova.openstack.org/v1beta1
kind: NovaAPI
metadata:
  name: nova-api
  namespace: openstack
spec:
  openStackServiceIdentity:
    openstackService: nova
    serviceComponent: api
  # ... other spec fields (replicas, customServiceConfig, etc.)
```
@gibizer @SeanMooney @stuggi kind ping in case you want to chime in.

stuggi: I have some higher prio tasks I need to focus on right now before I can come back to this. Its on my list for later this week

amartyasinha: thanks @stuggi!

bogdando: I would support  extending the spec like this, any solution which stops us guessing the structure of service/component from values

amartyasinha: ack! I'll propose a PR in nova-operator to extend the spec.

amartyasinha: Drafted a PR: https://github.com/openstack-k8s-operators/nova-operator/pull/1099
PTAL whenever you've got some spare time. Thanks!

modules/common/util/template_util.go
```diff
@@ -59,6 +59,7 @@ type Template struct {
 	ConfigOptions      map[string]interface{} // map of parameters as input data to render the templates
 	SkipSetOwner       bool                   // skip setting ownership on the associated configmap
 	Version            string                 // optional version string to separate templates inside the InstanceType/Type directory. E.g. placementapi/config/18.0
+	MultiTemplateDir   string                 // templates dir for multi-group operators, e.g. nova/api; requires InstanceType to be set
```
gibizer, line 62: If this holds the whole path like nova/api then why the InstanceType needs to be still set on the Template?

modules/common/util/template_util.go
```diff
@@ -81,22 +82,42 @@ func GetTemplatesPath() (string, error) {
...
-//   - kind - sub folder for the CRDs templates
+//   - subdir - directory under that root: one segment (legacy InstanceType, e.g. NovaAPI) or
+//     multi-segment (e.g. nova/api from MultiTemplateDir)
 //   - templateType - TType of the templates. When the templates got rendered and added to a CM
 //     this information is e.g. used for the permissions they get mounted into the pod
 //   - version - if there need to be templates for different versions, they can be stored in a version subdir
 //
 // Sub directories inside the specified directory with the above parameters get ignored.
-func GetAllTemplates(path string, kind string, templateType string, version string) []string {
+func GetAllTemplates(path string, subdir string, templateType string, version string) []string {
```
gibizer, line 96: generalizing from kind to subdir make sense to me. The kind can suggest a specific subdir but there can be non kind related subdirs as well. Good.

modules/common/util/template_util.go
```diff
@@ -81,22 +82,42 @@ func GetTemplatesPath() (string, error) {
...
 
-	templatePath := filepath.Join(path, strings.ToLower(kind), templateType, "*")
+	templatePath := filepath.Join(path, strings.ToLower(subdir), templateType, "*")
```
gibizer: if we assume subdir is already a valid path, then we don't need the ToLowe call. That is needed in the place when kind is transformed to a path.

modules/common/util/template_util.go
```diff
@@ -81,22 +82,42 @@ func GetTemplatesPath() (string, error) {
...
 
 	if version != "" {
-		templatePath = filepath.Join(path, strings.ToLower(kind), templateType, version, "*")
+		templatePath = filepath.Join(path, strings.ToLower(subdir), templateType, version, "*")
```
gibizer: ditto

modules/common/util/template_util.go
```diff
@@ -293,8 +314,22 @@ func GetTemplateData(t Template) (map[string]string, error) {
...
+		// though we do not use InstanceType here, but it is used in secret.go
+		if t.MultiTemplateDir != "" && t.InstanceType == "" {
+			return nil, ErrInstanceTypeUnsetWithMultiTemplateDir
+		}
+
+		var templateSubdir string
+		if t.MultiTemplateDir != "" {
+			templateSubdir = t.MultiTemplateDir
+		} else {
+			templateSubdir = t.InstanceType
```
gibizer: I think here is the place where we want to call toLower on the InstanceType to convert the Kind to a path

modules/common/util/template_util.go
```diff
@@ -293,8 +314,22 @@ func GetTemplateData(t Template) (map[string]string, error) {
 	data := make(map[string]string)
 
 	if t.Type != TemplateTypeNone {
-		// get all scripts templates which are in ../templesPath/cr.Kind/CMType/<OSPVersion - optional>
-		templatesFiles := GetAllTemplates(templatesPath, t.InstanceType, string(t.Type), string(t.Version))
+		// If MultiTemplateDir is set but InstanceType is not, return an error
+		// though we do not use InstanceType here, but it is used in secret.go
```
gibizer, line 299: Yeah I see it in https://github.com/openstack-k8s-operators/lib-common/blame/3e1007da0cbd673047305e423866c8c50b1e3167/modules/common/secret/secret.go#L272
  
 I'm wondering why we are setting an owner label on a Secret this way instead of using the owner reference way. Also it is strange that we only do that on Secret but not on ConfigMap.
@stuggi Are we still supporting cross namespace Secrets? (That is the most I could extract from the code why we are labelling)
Also it is strange that we only enforce InstanceType if the template is not TemplateTypeNone. So I guess a template with TemplateTypeNone and without a InstanceType cannot be outside of our namespace as secret creation would fail due to the missing InstanceType.

* If we don't need to support Secrets outside of the namespace the I suggest to clean that code up and remove this dependency and checks
* if we do need to support Secrets outside of the namespace then I suggest
  * to make the InstanceType mandatory in all template types to avoid surprises
  * consider supporting the same thing for ConfigMaps as well.

modules/common/util/template_util.go
```diff
@@ -293,8 +314,22 @@ func GetTemplateData(t Template) (map[string]string, error) {
...
+		if t.MultiTemplateDir != "" && t.InstanceType == "" {
+			return nil, ErrInstanceTypeUnsetWithMultiTemplateDir
+		}
+
+		var templateSubdir string
+		if t.MultiTemplateDir != "" {
```
gibizer, line 307: I would add a comment here to explain that MultiTemplateDir if set has a percedence. If it is not set then the code assumes that the templates are in the directory named after the InstanceType.

modules/common/util/template_util.go
```diff
@@ -81,22 +82,42 @@ func GetTemplatesPath() (string, error) {
...
+	if apiGroup == "" {
+		return "", kind
+	}
+	groupSegment = strings.ToLower(strings.Split(apiGroup, ".")[0])
+	return groupSegment, strings.TrimPrefix(kind, groupSegment)
+}
+
+// GetTemplateSubpath returns a single relative path under the templates root for CRD apiGroup and
+// Kind (e.g. nova.openstack.org + NovaAPI → nova/api, Nova → nova).
+func GetTemplateSubpath(apiGroup, kind string) string {
```
gibizer: Where this is used outside of the unit test?

gibizer (changes requested):

modules/common/util/template_util.go
gibizer: OK I see it below. Lets discuss it there.

modules/common/util/template_util.go
gibizer: I don't see this code being used in the PR outside of GetTemplateSubpath that is also unused. What is the reason and expected usage of this? The linked nova-operator CRD change seems scary.

gibizer (dismissed):

modules/common/util/template_util.go
amartyasinha: @gibizer GetTemplateSubpath is getting used in https://github.com/openstack-k8s-operators/nova-operator/pull/1096 which we can introduce after lib-common has got the support for handling MultiTemplateDir through this PR.

Regarding CRD change in nova-operator, it was to find a better way to extract values (`NovaAPI` to `nova` and `api`) instead of using `groupSegmentAndKindSuffix` function. Please see comment on CRD PR openstack-k8s-operators/nova-operator#1099

modules/common/util/template_util.go
amartyasinha: In https://github.com/openstack-k8s-operators/nova-operator/pull/1096

More detail in comments on https://github.com/openstack-k8s-operators/nova-operator/pull/1099

modules/common/util/template_util.go
amartyasinha: agree! I'll convert `subdir` (when it's `kind` value) to lower before passing the argument to `GetAllTemplates` func.

modules/common/util/template_util.go
amartyasinha: ack!

modules/common/util/template_util.go
amartyasinha: ack! adding

modules/common/util/template_util.go
amartyasinha: agree

modules/common/util/template_util.go
amartyasinha: added

modules/common/util/template_util.go
amartyasinha: done

modules/common/util/template_util.go
amartyasinha: done

modules/common/util/template_util.go
amartyasinha: done

modules/common/util/template_util.go
gibizer: I responded in https://github.com/openstack-k8s-operators/nova-operator/pull/1096

gibizer (commented):
It is cleaner now and the nova-operator patchs shows the intended usage of it. I wait for @stuggi about the wide question of instanceType usage and namespace external secret support. But other than that I'm OK to land this. Thanks @amartyasinha !

amartyasinha:
> It is cleaner now and the nova-operator patchs shows the intended usage of it. I wait for @stuggi about the wide question of instanceType usage and namespace external secret support. But other than that I'm OK to land this. Thanks @amartyasinha !

Thanks @gibizer for the review. IMHO, since secrets using `InstanceType` is not related to this change, maybe we can unblock this PR, and continue the discussion at other place. WDYT?

modules/common/util/template_util.go
amartyasinha: Now that we no longer have the functions to fetch path, this conversation can be marked resolved.

modules/common/util/template_util.go
amartyasinha: function along with unit test removed, thus marking resolved.

modules/common/util/template_util.go
stuggi: yes, the owner label is coming from creating/tracking object in different namespaces. There is one use case I know of where we create secret in a different namespace and this is for BMH provisioning, where userdata and networkdata secret needs to be created in the bmh namespaces , https://github.com/openstack-k8s-operators/openstack-baremetal-operator/blob/main/internal/openstackbaremetalset/baremetalhost.go#L83 , 

Just from a quick search it is not using the labels, for delete its using name https://github.com/openstack-k8s-operators/openstack-baremetal-operator/blob/main/internal/openstackbaremetalset/baremetalhost.go#L318, but I think its still good to see from the labels who the owner is, if someone checks the env and wonders who created those secrets in the bmh namespace. Therefore I don't think we should remove it. We can consider supporting it for CMs as well in a follow up. I wanted to spend more time on lib-common to cleanup/align things, but got other prios. Maybe after FR6 I can spend some time on this.

modules/common/util/template_util.go
```diff
@@ -293,8 +295,24 @@ func GetTemplateData(t Template) (map[string]string, error) {
...
+			return nil, ErrInstanceTypeUnsetWithMultiTemplateDir
+		}
+
+		var templateSubdir string
+		// if MultiTemplateDir is set, it will take precedence over InstanceType
+		// otherwise, InstanceType is used to create the template subdir based on CRD name
+		if t.MultiTemplateDir != "" {
+			templateSubdir = t.MultiTemplateDir
+		} else {
+			templateSubdir = strings.ToLower(t.InstanceType)
```
stuggi, line 310: good, we are doing the ToLower here, was wondering above

modules/common/util/template_util_test.go
```diff
@@ -300,14 +300,14 @@ func TestGetAllTemplates(t *testing.T) {
...
-		kind     string
+		subdir   string
 		tmplType TType
 		version  string
 		want     []string
 	}{
 		{
 			name:     "Get TemplateTypeConfig templates with no version",
-			kind:     "testservice",
+			subdir:   "testservice",
```
stuggi, line 310: maybe we should use `TestService` in one of them to check the ToLower situation? something we did not have right now.

stuggi:
> > It is cleaner now and the nova-operator patchs shows the intended usage of it. I wait for @stuggi about the wide question of instanceType usage and namespace external secret support. But other than that I'm OK to land this. Thanks @amartyasinha !
> 
> Thanks @gibizer for the review. IMHO, since secrets using `InstanceType` is not related to this change, maybe we can unblock this PR, and continue the discussion at other place. WDYT?

In general looks good to me. maybe we can adjust one of the tests I commented inline. I can land it later today.

modules/common/util/template_util_test.go
amartyasinha: Used `TestService` at L369.

amartyasinha:
> > > It is cleaner now and the nova-operator patchs shows the intended usage of it. I wait for @stuggi about the wide question of instanceType usage and namespace external secret support. But other than that I'm OK to land this. Thanks @amartyasinha !
> > 
> > 
> > Thanks @gibizer for the review. IMHO, since secrets using `InstanceType` is not related to this change, maybe we can unblock this PR, and continue the discussion at other place. WDYT?
> 
> In general looks good to me. maybe we can adjust one of the tests I commented inline. I can land it later today.

Thanks @stuggi! Addressed the change.

stuggi (approved):
/lgtm

stuggi:
@gibizer I can not merge it as it seems there is still a change request from your review. could you approve it so that it gets cleared?

modules/common/util/template_util.go
bogdando: thanks you for addressing this, LGTM

PR #683, opened 2026-04-10
Create a mechanism for analyzing certificates for PQC-readiness

subecho:
This patch adds functionality to the tls module where we can now generate an analysis for a given certificate and determine its readiness for PQC and compliance for TLS 1.3. Tests were also added to ensure this all works.

We will need to go back and add more to this when the new algorithms become available in the crypto libraries for true, quantum-safe instead of quantum-resistant solutions.

Co-Author: Claude <claude@anthropic.com>

### DIFF
--- a/modules/common/tls/tls.go
+++ b/modules/common/tls/tls.go
@@ -21,7 +21,13 @@ package tls
 
 import (
 	"context"
+	"crypto/ecdsa"
+	"crypto/ed25519"
+	"crypto/rsa"
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
+	"encoding/pem"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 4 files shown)


modules/common/tls/tls.go
```diff
@@ -443,3 +449,233 @@ func (c *Ca) CreateVolume() corev1.Volume {
...
+	x509.SHA256WithRSAPSS: false, // Depends on key size
+	x509.SHA384WithRSAPSS: false, // Depends on key size
+	x509.SHA512WithRSAPSS: false, // Depends on key size
+	x509.PureEd25519:      false, // Ed25519 is not PQC-safe
+}
+
+// Minimum key sizes required for transitional PQC safety (see NIST SP 800-57)
+const (
+	minPQCSafeRSAKeySize   = 3072
+	minPQCSafeECDSAKeySize = 384 // P-384 curve
```
abays, line 488: My understanding is that RSA and ECDSA are never PQC "safe" (as in, immune to breakage), but are rather PQC "resistant".  I'm not a PQC SME, however.

modules/common/tls/tls.go
```diff
@@ -443,3 +449,233 @@ func (c *Ca) CreateVolume() corev1.Volume {
...
+			return pubKey.Curve.Params().BitSize >= minPQCSafeECDSAKeySize
+		}
+	}
+
+	return baseSafe
+}
+
+// isTLS13Compatible checks if a certificate is compatible with TLS 1.3
+func isTLS13Compatible(cert *x509.Certificate) bool {
+	// TLS 1.3 removed support for RSA-PSS and requires specific signature algorithms.
```
abays, line 567: I'm not sure this is true with regards to TLS 1.3.  I think TLS 1.3 mandates support for RSA-PSS but instead dropped PKCS#1 v1.5 signatures.

modules/common/tls/tls.go
```diff
@@ -443,3 +449,233 @@ func (c *Ca) CreateVolume() corev1.Volume {
...
+	tls13Suites := []string{
+		"TLS_AES_128_GCM_SHA256",
+		"TLS_AES_256_GCM_SHA384",
+		"TLS_CHACHA20_POLY1305_SHA256",
+	}
+
+	if pqcSafe {
+		// For PQC-safe configs, prefer stronger ciphers
+		return append([]string{
+			"TLS_AES_256_GCM_SHA384",
```
abays, line 596: This results in a duplicate entry, given L589.

PR #684, opened 2026-04-16
[b/r] Default cert secrets to restore=false in EnsureCert

stuggi:
Service cert secrets are reissued by cert-manager from the restored CA after restore, so they don't need to be restored from backup. Set restore=false as the default in EnsureCert so callers don't need to pass it explicitly. Callers that need restore=true (e.g., CA cert secrets) override via request.Labels.

Jira: OSPRH-22912
Jira: OSPRH-22913


abays (approved):
/lgtm

PR #685, opened 2026-04-16
[b/r] Apply backup annotation overrides in EnsureCert

stuggi:
Check for backup annotation overrides on existing cert secrets in EnsureCert. This ensures cert-manager's SecretTemplate reflects user-set annotations (e.g., backup.openstack.org/restore=true) so they are not overwritten on the next Certificate reconcile.


abays (approved):
/lgtm

PR #687, opened 2026-04-30
[template]Make execTempl non-global

gibizer:
There was a package global variable in template_utils to implement execting templating within the template file itself. This global made it possible that two such templating function within the same process interfere via that global leading to intermittent errors like

```
  "error": "template: tmp:365:12: executing \"tmp\" at <execTempl \"nova-template\" .>: error calling execTempl: template: no template \"nova-template\" associated with template \"tmp\"
```
during parallel reconciliation with the same manager.

This makes the shared variable a local variable for each template execution calls.

### DIFF
--- a/modules/common/util/template_util.go
+++ b/modules/common/util/template_util.go
@@ -142,19 +142,6 @@ func ExecuteTemplate(templateFile string, data interface{}) (string, error) {
 	return renderedTemplate, nil
 }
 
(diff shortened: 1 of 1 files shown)


mrkisaolamb (approved):
lgtm

abays (approved):
/lgtm

PR #692, opened 2026-05-12
Add edpm module with unstructured NodeSet secret hash check

lmiccini:
Add a new modules/edpm package that checks whether OpenStackDataPlaneNodeSet deployed secret hashes are in sync with current cluster secrets. Uses an unstructured client to avoid importing dataplane API types, allowing consumers like infra-operator to drop the heavyweight openstack-operator/api dependency.


modules/edpm/nodeset.go
```diff
@@ -0,0 +1,135 @@
...
+		Group:   NodeSetGVK.Group,
+		Version: NodeSetGVK.Version,
+		Kind:    NodeSetGVK.Kind + "List",
+	})
+
+	if err := c.List(ctx, nodesetList, client.InNamespace(namespace)); err != nil {
+		if meta.IsNoMatchError(err) {
+			Log.Info("OpenStackDataPlaneNodeSet CRD not installed, skipping hash check")
+			return true, "", nil
+		}
```
abays, line 78: Maybe we should add a test for this path?

abays (approved):
This seems good to me, as it avoids the dreaded circular dependency problem

/lgtm

modules/test/crd.go
```diff
@@ -133,6 +133,28 @@ func GetCRDDirFromModule(moduleName string, goModPath string, relativeCRDPath st
...
+// the lib-common test module, similar to GetOpenShiftCRDDir.
+func GetDataPlaneCRDDir(goModPath string) (string, error) {
+	libCommon := "github.com/openstack-k8s-operators/lib-common/modules/test"
+	libCommon, version, err := getDependencyVersion(libCommon, goModPath)
+	if err != nil {
+		return "", err
+	}
+	versionedModule := fmt.Sprintf("%s@%s", libCommon, version)
+	path := filepath.Join(build.Default.GOPATH, "pkg", "mod", versionedModule, "dataplane_crds")
+
```
stuggi, line 156: using replace in go.mod will not work for this, test CRD, should we allow it?

```
        var path string
        if version == "" && strings.HasPrefix(libCommon, ".") {
                goModDir := filepath.Dir(goModPath)
                path = filepath.Join(goModDir, libCommon, "dataplane_crds")                                                                                                                  
        } else {                                                                                                                                                                             
                versionedModule := fmt.Sprintf("%s@%s", libCommon, version)                                                                                                                  
                path = filepath.Join(build.Default.GOPATH, "pkg", "mod", versionedModule, "dataplane_crds")                                                                                  
        }  
```

modules/test/crd.go
```diff
@@ -133,6 +133,28 @@ func GetCRDDirFromModule(moduleName string, goModPath string, relativeCRDPath st
 	return path, nil
 }
 
+// GetDataPlaneCRDDir returns the absolute path of the directory holding the
+// dataplane custom resource definitions. It will look the CRD path up from
+// the lib-common test module, similar to GetOpenShiftCRDDir.
```
stuggi: wondering if we should mention in the func desc that this should only be used if the CRD from the openstack-operator can not be used and its a minimal CRD?

stuggi (approved):
/lgtm

PR #693, opened 2026-05-13
Add HaveNodeSets helper and move edpm funcs to unstructured sub-package

lmiccini:
Add a lightweight function to check whether any NodeSets with
status.secretHashes exist in a namespace, for use by controllers
that need to skip NodeSet-dependent logic when no computes are
deployed.

Move all NodeSet helpers into edpm/unstructured to make the
dependency boundary explicit and prevent future PRs from introducing
typed dataplane API imports in this package.


stuggi (approved):
/lgtm

PR #694, opened 2026-05-14
Dedup some httpd related templates

fmount:
The `ssl.conf` `httpd` configuration is duplicated across service operators. Any TLS policy change (cipher suites,
protocol versions) requires coordinated patches across all repositories.
This patch embeds `ssl.conf` in `lib-common` via `//go:embed` and exposes a `GetCommonTemplates()` function that renders the embedded templates and returns the rendered content.
Service operators can consume this before rendering their config files.
This approach requires no changes in the existing functions/structs and no new abstractions.

Jira: [OSPRH-30239](https://redhat.atlassian.net/browse/OSPRH-30239)

### DIFF
--- a/modules/common/util/template_util.go
+++ b/modules/common/util/template_util.go
@@ -19,7 +19,9 @@ package util
 import (
 	"bufio"
 	"bytes"
+	"embed"
 	"fmt"
+	"io/fs"
 	"os"
 	"path"
 	"path/filepath"
@@ -28,6 +30,9 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 )
+
+//go:embed templates/common/config/*
+var commonTemplates embed.FS
 
 // TType - TemplateType
 type TType string
@@ -343,4 +348,35 @@ func GetTemplateData(t Template) (map[string]string, error) {
 	}
 
 	return data, nil
+}
+
+// GetCommonTemplates renders the common config templates (ssl.conf, etc.)
+// shipped with lib-common using the provided configOptions, and returns
+// map[filename]renderedContent.
+// Callers merge the result into their Template.CustomData before calling
+// EnsureSecrets/EnsureConfigMaps so that common templates are included in the
+// generated Secret/ConfigMaps without each operator duplicating the files.
+func GetCommonTemplates(configOptions map[string]any) (map[string]string, error) {
+	dir := "templates/common/config"
+	entries, err := fs.ReadDir(commonTemplates, dir)
+	if err != nil {
+		return nil, err
+	}
+
+	result := make(map[string]string, len(entries))
+	for _, e := range entries {
+		if e.IsDir() {
+			continue
+		}
+		data, err := fs.ReadFile(commonTemplates, path.Join(dir, e.Name()))
+		if err != nil {
+			return nil, err
+		}
+		rendered, err := ExecuteTemplateData(string(data), configOptions)
+		if err != nil {
+			return nil, err
+		}
+		result[e.Name()] = rendered
+	}
+	return result, nil
 }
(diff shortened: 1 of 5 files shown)


fultonj:
I think this patch and how it's consumed in the following is a good de-duplication

- https://github.com/openstack-k8s-operators/glance-operator/pull/917
- https://github.com/openstack-k8s-operators/cinder-operator/pull/648

modules/common/util/template_util.go
```diff
@@ -293,6 +298,19 @@ func GetTemplateData(t Template) (map[string]string, error) {
 
 	data := make(map[string]string)
 
+	// Render common embedded templates as fallback defaults.
+	// On-disk operator templates rendered below overwrite these
+	// if they share the same filename, giving operators precedence.
+	if t.Type == TemplateTypeConfig {
+		commonData, err := GetCommonTemplates(opts)
```
stuggi, line 307: wouldn't this inject the ssl.conf, or any other future config into all services, even if they don't use httpd?

stuggi: maybe the Template struct should have something like CommonTemplates []string where the consumer can pass a list of the ones which should be included?

fmount: Yes, in theory the idea is to inject common templates that are valid for all the operators. I didn't think about a scenario where we want to explicitly ignore them or make a selective import, and afaik all services use `httpd` right?
In addition, I wanted to extend this function to not have to change anything in the service operators (except removing the common files).
However I can investigate the change you have in mind and pass a `CommonTemplate []string` so that services can explicitly tell what to get from the common templates.

stuggi: I was thinking about  cinder-scheduler or volume, similar nova componentes. they won't use httpd

fmount: ah right, it makes sense to not get everything when you have multiple components. That would be exactly the purpose of `CommonTemplate []string`. I will update the patches accordingly!

fmount: Done, and I see good results in both glance and cinder. I think we can spread the pattern across the board and extend the common templates with more files/templates.
In storage, for example, it might be interesting to dedup `httpd.conf` and add a few `templateParams` to make sure we can do a per-service customization.
However I'd like to do this work in multiple iterations, and focus this iteration on `ssl.conf` which was part of the initial conversation. This way when `pqc` comes, we just need to update the single, lib-common, instance (1 single PR), and wait for the bump across the operators (which happens automatically on a weekly basis).

fmount:
@stuggi if you're good with this implementation (see glance and cinder as examples of consuming templates from a common location), I have patches ready for all operators I can push (cc @abays in case you want to take a look at these patches too).

abays:
@fmount I actually looked last week and forgot to say...

/lgtm

modules/common/util/template_util.go
```diff
@@ -293,6 +299,22 @@ func GetTemplateData(t Template) (map[string]string, error) {
...
+	// Note that local operator templates rendered below overwrite the common
+	// ones: if they share the same filename, service operators definition take
+	// precedence
+	if len(t.CommonTemplates) > 0 {
+		commonData, err := GetCommonTemplates(opts)
+		if err != nil {
+			return nil, err
+		}
+		for _, name := range t.CommonTemplates {
+			if v, ok := commonData[name]; ok {
```
stuggi: if the template requested was not found, should it result in an error? like if `CommonTemplates: []string{"sls.conf"}` it would be silent ignored. it probably should?

```
      v, ok := commonData[name]
      if !ok {
          return nil, fmt.Errorf("common template %q not found", name)
      }
      data[name] = v
```

fmount: I wanted to silently ignore to avoid blocking the operator in case one template of the list is not there.
However, because we control this interface from the service operators (because we pass that list), it's probably ok to return if the template does not exist and notify it to the caller instead of progressing in the reconciliation loop.
Let me push an update to this logic.

fmount: Done, and added a test to verify we return the error when we pass `[]string{"sls.conf"},`.
Thanks @stuggi, this is a good improvement overall.

stuggi (approved):
/lgtm

stuggi (approved):
/lgtm

PR #696, opened 2026-05-18
Add exec probe handler support to probes API

zzzeek:
Add ProbeHandlerType (HTTP/Exec) and Command/Port/Scheme fields to ProbeConf. Introduce SetProbeConfV2 and CreateProbeSetV2 with a self-contained ProbeConf signature that switches on handler type. Export Merge for downstream use. Existing SetProbeConf/CreateProbeSet delegate to V2 for full backward compatibility.

References: [OSPRH-30190](https://redhat.atlassian.net/browse/OSPRH-30190)


modules/common/probes/probes.go
```diff
@@ -125,72 +127,103 @@ func CreateProbeSet(
...
 		return errorList
 	}
-	// Path validation: fail is explicitly set as an empty string
-	// or the endpoint does't start with "/"
-	if config.Path != "" && !strings.HasPrefix(config.Path, "/") {
-		err := field.Invalid(basePath.Child("path"), config.Path,
-			"path must start with '/' if specified")
-		errorList = append(errorList, err)
+
+	switch config.Type {
```
fmount: @zzzeek not sure about your thoughts but I'd like to consolidate these two switch-case(s) into a single one.
e.g.
```golang
switch config.Type {
case ProbeHandlerExec:
    if len(config.Command) == 0 {
        errorList = append(errorList, field.Required(basePath.Child("command"),
            "command is required for exec probe type"))
    }
case "", ProbeHandlerHTTP:
    if config.Path != "" && !strings.HasPrefix(config.Path, "/") {
        errorList = append(errorList, field.Invalid(basePath.Child("path"), config.Path,
            "path must start with '/' if specified"))
    }
default:
    errorList = append(errorList, field.Invalid(basePath.Child("type"), config.Type,
        "type must be one of: HTTP, Exec, or unset"))
}
```
This `default` should address the invalid type which is the entrypoint for our check, while the "valid" use cases can check other parameters and eventually raise a specific error.

modules/common/probes/probes.go
```diff
@@ -125,72 +127,103 @@ func CreateProbeSet(
...
 		InitialDelaySeconds: config.InitialDelaySeconds,
 		TimeoutSeconds:      config.TimeoutSeconds,
 		PeriodSeconds:       config.PeriodSeconds,
 		FailureThreshold:    config.FailureThreshold,
 	}
-	if scheme != nil {
-		probe.HTTPGet.Scheme = *scheme
+
+	switch config.Type {
+	case ProbeHandlerExec:
```
fmount, line 144: I'm wondering if we really need  a `config.Type`. While I understand the purpose and makes things readable, technically you can infer the type from the `len(config.Command)`: if it is > 0 it implies you asked for an Exec Handler type, otherwise http, no other potential choices or extension are allowed AFAIK in k8s.
@zzzeek wdyt?

modules/common/probes/types.go
```diff
@@ -24,15 +24,29 @@ import (
...
+
 // ProbeConf - the configuration for liveness and readiness probes
-// LivenessPath - Endpoint path for the liveness probe
-// ReadinessPath - Endpoint path for the readiness probe
-// InitialDelaySeconds - Number of seconds after the container starts before liveness/readiness probes are initiated
-// TimeoutSeconds - Number of seconds after which the probe times out
-// PeriodSeconds - How often (in seconds) to perform the probe
 type ProbeConf struct {
+	// +kubebuilder:validation:Optional
+	Type ProbeHandlerType `json:"type,omitempty"`
```
fmount, line 41: This is connected to my other comment. In other words I'm wondering if we should infer the `ProbeHandlerType` from the `Command` (it must be passed in that case) and default to `HTTP` (the only other choice) when a `command` is not passed. The other checks might remain unchanged or enhanced.

modules/common/probes/probes.go
zzzeek: still a python dev I think of "explicit is better than implicit".    We discussed supporting other kinds of probes in the future (e.g. expanding the enum to include TCP, GRPC), which means for validation we need a way to validate the mutual exclusiveness of the different handler types and up front we need to know which `ProbeHandler` sub-field we are looking at.  If we support Exec, HTTP, TCP and GRPC all at the same time, it seems like it would be uncomfortable defining a hierarchy of which field sets we look at first to determine the intent, which we then validate on once we've decided on which handler type we think this is.   plus as you mention it makes things readable.

in the `Probe` struct, this mutual exclusiveness is determined by which of the sub-actions inside of `ProbeHandler` are set.   so they dont have quite the same problem since we're working with a flat namespace.

fmount: fair enough, it's eventually just a string so it shouldn't impact on the CRD size. let's keep it explicit if the idea is to extend it more over the time.

modules/common/probes/probes.go
zzzeek: yeah sorry looks like the robot was getting creative here, I spent most of my time fixing up the tests it generated.

modules/common/probes/types.go
zzzeek: per the other comment we want to leave room for TCP / others so leaving this as explicit should save on some headaches

modules/common/probes/probes.go
```diff
@@ -125,72 +127,99 @@ func CreateProbeSet(
 	}, nil
 }
 
-// SetProbeConf configures and returns liveness and readiness probes based on
-// the provided settings
+// SetProbeConf configures and returns an HTTP GET probe based on the provided
+// settings. For exec-based probes, use SetProbeConfV2 instead.
 func SetProbeConf(port int32, scheme *v1.URIScheme, config ProbeConf) (*v1.Probe, error) {
```
fmount: Thanks to the `CreateProbeSetV2` abstraction, I think `SetProbeConf` is not used anymore (except for tests). @zzzeek I'm wondering if we want to remove it since the entrypoint for the operators is `CreateProbeSet` that calls its `V2` variation. From my research all operators use `probes.CreateProbeSet(...)`, which internally woudl call at this point `SetProbeConfV2`.

fmount (commented):
Looks like we're going to the right direction here. I think we're almost good but I have an additional question on code cleanup after I checked the existing operators using this feature. It can go on a follow up though, but would be good to address while we're designing the V2 here.

modules/common/probes/probes.go
zzzeek: OK, I just removed SetProbeConf in the latest

fmount: thank you!

fmount (approved):

fmount:
@zzzeek I would have renamed `SetProbeConfV2` to `SetProbeConf` in the last commit but this is really irrelevant as it represents a hidden implementation detail. +1 from me as this extension is backward compatible with the previous version, and we do not need to rush any change in the operators that adopted this interface already.

PR #698, opened 2026-06-01
Tighten common ssl.conf defaults

rabi:
Prefer forward-secret GCM ciphers and disable TLSv1.1 in the shared Apache SSL template so operators consuming lib-common inherit stronger defaults.

jira: OSPRH-29016


stuggi:
looks good to me, maybe we wait for FR6 before landing it? iirc this is a 19 target, right?

fmount:
Right, as long as FR6 lands this might be a good candidate for the main branch that will track 19 work.

stuggi (approved):
/lgtm

PR #700, opened 2026-06-08
Add IsSecretHashInSync for scoped single-secret hash checking

lmiccini:
AreSecretHashesInSync checks all secrets across all NodeSets, which makes it unsuitable for controllers that need to track one specific secret without being affected by unrelated hash mismatches.

IsSecretHashInSync checks a single named secret against the secretHashes in all NodeSets. Returns inSync=true when the hash matches everywhere the secret is tracked, or when no NodeSet tracks it.


stuggi (approved):
/lgtm

PR #701, opened 2026-06-09
Cache secret hashes in nodeset sync checks to avoid redundant API calls

lmiccini:
AreSecretHashesInSync and IsSecretHashInSync previously fetched and hashed each secret once per NodeSet that referenced it. In environments with N NodeSets sharing the same secrets, this caused N redundant Get and Hash calls per secret per reconcile cycle.

Add a hash cache (map for AreSecretHashesInSync, single variable for IsSecretHashInSync) so each secret is fetched and hashed at most once per function invocation. No signature or behavioral change for callers.

### DIFF
--- a/modules/edpm/unstructured/nodeset.go
+++ b/modules/edpm/unstructured/nodeset.go
@@ -85,6 +85,8 @@ func AreSecretHashesInSync(
 		return true, "", nil
 	}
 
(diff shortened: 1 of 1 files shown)


stuggi (approved):
/lgtm

PR #702, opened 2026-06-10
Add annotation for skipping webhook validation

abays:
Adds a general annotation that can be consumed by operators to indicate that webhook validation should be skipped for a given resource, if they so choose to implement that functionality.


stuggi (approved):
/lgtm

PR #704, opened 2026-06-17
Require all pods updated before reporting IsReady

lmiccini:
During rolling updates, ReadyReplicas can match the desired count while UpdatedReplicas lags behind — old pods are still ready but running stale config. Add UpdatedReplicas == Spec.Replicas to both deployment.IsReady() and statefulset.IsReady() so consumers (e.g. transport secret rotation guards) don't act on stale readiness.


stuggi (approved):
/lgtm

PR #705, opened 2026-06-18
Default ImagePullPolicy to IfNotPresent for all workload helpers

abays:
When `ImagePullPolicy` is not explicitly set on a container and the image tag is "latest" or unset, Kubernetes defaults the policy to `Always`, causing unnecessary image pulls on every pod restart. This affected ~120 container definitions across all service operators.

Add `pod.SetPullPolicyDefaults()` which sets `PullIfNotPresent` on any container that lacks an explicit policy, and call it from the `Deployment`, `DaemonSet`, `StatefulSet`, `Job`, and `CronJob` `CreateOrPatch` paths. Containers that already set an explicit policy (including `PullAlways`) are left unchanged.

Relates-To: https://redhat.atlassian.net/browse/OSPRH-26843

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/modules/common/job/job.go
+++ b/modules/common/job/job.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/pod"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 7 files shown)


modules/common/job/job.go
```diff
@@ -129,6 +130,8 @@ func (j *Job) DoJob(
 	var ctrlResult ctrl.Result
 	var err error
 
+	pod.SetPullPolicyDefaults(&j.expectedJob.Spec.Template.Spec)
```
stuggi, line 133: this might retrigger jobs if that results in the hash to change, but I guess thats ok/needed

stuggi (approved):
/lgtm

PR #706, opened 2026-06-22
Add credential rotation helpers for secret rotation

lmiccini:
Add shared helpers for transport URL secret rotation across all
openstack-k8s-operators consumer operators:

- object.ManageSecretConsumerFinalizer: adds a consumer finalizer
  to a secret so the provider knows consumers still depend on it 
- object.RemoveSecretConsumerFinalizer: removes the finalizer
- object.FinalizeSecretRotation: the rotation guard — if old != new
  and guard is ready, removes finalizer from old and returns new;
  otherwise returns old to preserve the reference
- object.ManageRotationGracePeriod: time-based grace period that
  gives sub-CRs time to detect config changes, roll pods, and
  update conditions before the guard evaluates readiness
- condition.CredentialRotationGuardReady: returns true when all
  sub-CR specs are stable and all mirrored conditions are True
- condition.ServiceInstanceIsReady: generation/observedGeneration
  guard with replica count and DeploymentReadyCondition check
- statefulset.IsReady: checks ReadyReplicas, UpdatedReplicas,
  ObservedGeneration, and CurrentRevision == UpdateRevision so 
  DeploymentReady is only True when all pods have rolled
- deployment.IsReady: checks ReadyReplicas, UpdatedReplicas,
  Status.Replicas, and ObservedGeneration

Note: statefulset.IsReady now additionally requires
CurrentRevision == UpdateRevision compared to the previous
implementation. This tightens the readiness check to prevent
declaring ready during in-progress rolling updates.

### DIFF
--- a/modules/common/condition/funcs.go
+++ b/modules/common/condition/funcs.go
@@ -166,6 +166,37 @@ func (conditions *Conditions) IsUnknown(t Type) bool {
 	return true
 }
 
+// ServiceInstanceIsReady reports whether a service sub-CR has finished
+// reconciling its current spec and its deployment is ready to serve
+// requests. It implements the generation/observedGeneration guard used
+// by OpenStack service operators during credential rotation rollouts.
+func ServiceInstanceIsReady(generation, observedGeneration int64, readyCount, replicas int32, conditions *Conditions) bool {
+	if conditions == nil {
+		return false
+	}
+	if generation != observedGeneration || readyCount != replicas {
+		return false
+	}
+	if replicas == 0 {
+		return true
+	}
+	return conditions.IsTrue(DeploymentReadyCondition)
+}
+
+// CredentialRotationGuardReady reports whether it is safe to remove a
+// consumer finalizer from an old transport URL or application credential
+// secret. The guard requires all sub-CR specs to be stable (no pending
+// CreateOrPatch updates) and all parent sub-conditions to be True.
+// Note: ReadyCondition is NOT checked because operators reset conditions
+// via Init() at the top of each Reconcile, so ReadyCondition is Unknown
+// until the very end. AllSubConditionIsTrue covers the same safety.
+func CredentialRotationGuardReady(allSubCRsStable bool, conditions *Conditions) bool {
+	if conditions == nil {
+		return false
+	}
+	return allSubCRsStable && conditions.AllSubConditionIsTrue()
+}
+
 // AllSubConditionIsTrue validates if all subconditions are True
 // It assumes that all conditions report success via the True status
 func (conditions *Conditions) AllSubConditionIsTrue() bool {
--- a/modules/common/object/metadata.go
+++ b/modules/common/object/metadata.go
@@ -22,15 +22,18 @@ import (
 	"encoding/json"
 	"fmt"
 	"slices"
+	"time"
 
+	commonannotations "github.com/openstack-k8s-operators/lib-common/modules/common/annotations"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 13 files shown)


modules/common/object/metadata.go
```diff
@@ -181,3 +183,168 @@ func ManageConsumerFinalizer(
...
+	if graceUntilStr == "" {
+		before := obj.DeepCopyObject().(client.Object)
+		graceUntil := time.Now().Add(gracePeriod)
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
+		annotations[RotationGraceAnnotation] = graceUntil.Format(time.RFC3339)
+		obj.SetAnnotations(annotations)
+		if err := c.Patch(ctx, obj, client.MergeFrom(before)); err != nil {
+			return ctrl.Result{}, true, err
```
stuggi: If this Patch fails, the function returns (ctrl.Result{}, true, err). The true means `graceActive=true` even though the grace period was not set. The documented caller pattern checks err first so it's currently harmless, but should it return false since the grace period was not activated?

modules/common/condition/funcs.go
```diff
@@ -166,6 +166,34 @@ func (conditions *Conditions) IsUnknown(t Type) bool {
...
+// requests. It implements the generation/observedGeneration guard used
+// by OpenStack service operators during credential rotation rollouts.
+func ServiceInstanceIsReady(generation, observedGeneration int64, readyCount, replicas int32, conditions *Conditions) bool {
+	if conditions == nil {
+		return false
+	}
+	return generation == observedGeneration &&
+		readyCount == replicas &&
+		(conditions.IsTrue(DeploymentReadyCondition) ||
+			(conditions.IsFalse(DeploymentReadyCondition) && replicas == 0))
```
stuggi: Wondering if the condition state matter if replicas==0 set for the service?

modules/common/statefulset/statefulset.go
```diff
@@ -157,9 +167,11 @@ func (s *StatefulSet) Delete(
...
 // - all pods run the current spec (UpdatedReplicas == requested replicas)
 // - both when the Generatation of the object matches the ObservedGeneration of the Status
+// - the rollout is complete (CurrentRevision == UpdateRevision)
 func IsReady(deployment appsv1.StatefulSet) bool {
 	return deployment.Spec.Replicas != nil &&
 		*deployment.Spec.Replicas == deployment.Status.ReadyReplicas &&
 		*deployment.Spec.Replicas == deployment.Status.UpdatedReplicas &&
-		deployment.Generation == deployment.Status.ObservedGeneration
+		deployment.Generation == deployment.Status.ObservedGeneration &&
+		deployment.Status.CurrentRevision == deployment.Status.UpdateRevision
```
stuggi, line 176: modules/common/test/helpers/statefulset.go:47 — Test helpers don't set CurrentRevision/UpdateRevision

SimulateStatefulSetReplicaReady and SimulateStatefulSetReplicaReadyWithPods set ReadyReplicas, UpdatedReplicas, and ObservedGeneration but never set CurrentRevision or UpdateRevision. The new IsReady check passes accidentally because "" == "". We may want to update the helpers to set them to something?

modules/common/object/metadata.go
```diff
@@ -181,3 +183,168 @@ func ManageConsumerFinalizer(
...
+	); err != nil {
+		return statusSecretName, err
+	}
+	return currentSecretName, nil
+}
+
+const (
+	// RotationGraceAnnotation is set on an object to track when a
+	// credential rotation grace period started.
+	RotationGraceAnnotation = "openstack.org/rotation-grace-until"
```
stuggi: should we move it to https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/annotations/annotations.go

stuggi (approved):
/lgtm

PR #714, opened 2026-07-08
Add NewRule constructor to secret package

zzzeek:
we're looking to move this code to use lib-common:

https://github.com/openstack-k8s-operators/mariadb-operator/blob/66665ea51eed2025f566351a2aabd82b2f049fac/internal/controller/mariadbaccount_controller.go#L761-L785

however we need to be able to create a rule with that character whitelist.

### DIFF
--- a/modules/common/secret/password.go
+++ b/modules/common/secret/password.go
@@ -76,6 +76,13 @@ var (
 	ErrPasswordRequirements = errors.New("password does not meet the requirements")
 )
 
+// NewRule creates a Rule for use with PasswordValidator.Requirements or
+// PasswordValidator.Rejects. Callers outside this package must use this
+// constructor since Rule's fields are unexported.
+func NewRule(description string, pattern regexp.Regexp) Rule {
+	return Rule{description: description, pattern: pattern}
+}
+
 // Validate - implements the Validator interface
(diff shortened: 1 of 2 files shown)


zzzeek:
there's another way to go here if we prefer, and it's to add a built-in rule that does the character whitelist we have which is `allowedSpecialChars := "@#%^*-_=+:,.!~"`.   this is the character list we've decided is "safe" for a database password that would be in an ALTER USER statement.

fmount:
> there's another way to go here if we prefer, and it's to add a built-in rule that does the character whitelist we have which is `allowedSpecialChars := "@#%^*-_=+:,.!~"`. this is the character list we've decided is "safe" for a database password that would be in an ALTER USER statement.

Hi @zzzeek sorry for being late here. The original idea was to have a set of common rules that might have been applied to all the operators. If this pattern applies to all services, I'm good w/ a generic/static rule.
I also see we might have per-service patterns based on the purpose of each component, so I'm not against exposing a `NewRule` that customizes the behavior to meet a given component expectations.

Based on the above (and your tests) we can land this one or add the pattern to the built-in rule.

stuggi (approved):
/lgtm

PR #720, opened 2026-07-22
Add projected SA token volume helpers to serviceaccount package and default security pod context helper

lmiccini:
Add KubeAPIAccessVolume() and KubeAPIAccessVolumeMount() helpers that create a projected volume replacing the default automounted SA token with a time-limited one (1h expiry). Use together with AutomountServiceAccountToken=false on the PodSpec.

This enables operators to harden workloads by disabling the default long-lived automounted token while still providing K8s API access via a short-lived projected token at the standard mount path.

Add RestrictiveSecurityContext helper to pod package: provides a centralized helper for building hardened container
SecurityContext for unprivileged workloads: RunAsNonRoot, drops all capabilities, disables privilege escalation, and applies the RuntimeDefault seccomp profile.

### DIFF
--- a/modules/common/pod/security.go
+++ b/modules/common/pod/security.go
@@ -1,0 +1,53 @@
+/*
+Copyright 2026 Red Hat
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package pod
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/ptr"
+)
+
+// RestrictiveSecurityContext returns a hardened container SecurityContext
+// suitable for unprivileged workloads. It sets RunAsNonRoot, drops all
+// capabilities, disables privilege escalation, and applies the RuntimeDefault
+// seccomp profile. The provided uid is used for both RunAsUser and RunAsGroup.
+// Optional addCapabilities are added back after dropping ALL.
+func RestrictiveSecurityContext(uid int64, addCapabilities ...corev1.Capability) *corev1.SecurityContext {
+	return RestrictiveSecurityContextWithGID(uid, uid, addCapabilities...)
+}
+
+// RestrictiveSecurityContextWithGID is like RestrictiveSecurityContext but
+// allows specifying a different GID.
+func RestrictiveSecurityContextWithGID(uid, gid int64, addCapabilities ...corev1.Capability) *corev1.SecurityContext {
+	caps := &corev1.Capabilities{
+		Drop: []corev1.Capability{"ALL"},
+	}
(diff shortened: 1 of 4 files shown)


modules/common/serviceaccount/projected_token.go
```diff
@@ -0,0 +1,94 @@
...
+	// projected service account token (1 hour).
+	DefaultTokenExpirationSeconds int64 = 3600
+)
+
+// KubeAPIAccessVolume returns a projected Volume that provides a
+// time-limited service account token, the cluster CA certificate, and
+// the pod namespace.  Use it together with AutomountServiceAccountToken=false
+// on the PodSpec to replace the default long-lived automounted token with a
+// short-lived one.
+func KubeAPIAccessVolume() corev1.Volume {
```
stuggi: might be better to make allow ExpirationSeconds to be configurable and just default to   DefaultTokenExpirationSeconds if not provided?

modules/common/pod/security.go
```diff
@@ -0,0 +1,52 @@
...
+// RestrictiveSecurityContextWithGID is like RestrictiveSecurityContext but
+// allows specifying a different GID.
+func RestrictiveSecurityContextWithGID(uid, gid int64, addCapabilities ...corev1.Capability) *corev1.SecurityContext {
+	caps := &corev1.Capabilities{
+		Drop: []corev1.Capability{"ALL"},
+	}
+	if len(addCapabilities) > 0 {
+		caps.Add = addCapabilities
+	}
+	return &corev1.SecurityContext{
```
stuggi, line 42: AFAIK, ReadOnlyRootFilesystem defaults to false. a  RestrictiveSecurityContextWithGID maybe should default  to true and have a parameter to override?

modules/common/pod/security_test.go
```diff
@@ -0,0 +1,74 @@
+package pod
```
stuggi, line 17: nit, should also have copyright header

modules/common/serviceaccount/projected_token.go
```diff
@@ -0,0 +1,99 @@
...
+func KubeAPIAccessVolume(expirationSeconds ...int64) corev1.Volume {
+	expiration := DefaultTokenExpirationSeconds
+	if len(expirationSeconds) > 0 && expirationSeconds[0] > 0 {
+		expiration = expirationSeconds[0]
+	}
+	return corev1.Volume{
+		Name: KubeAPIAccessVolumeName,
+		VolumeSource: corev1.VolumeSource{
+			Projected: &corev1.ProjectedVolumeSource{
+				DefaultMode: ptr.To[int32](0o444),
```
stuggi: lets make it `DefaultMode: ptr.To[int32](0444),` to make it consistent with how we use it in other places in lib-common as in https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/tls/tls.go#L438C6-L438C39

modules/common/serviceaccount/projected_token_test.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+func TestKubeAPIAccessVolume(t *testing.T) {
+	g := NewWithT(t)
+
+	vol := KubeAPIAccessVolume()
+
+	g.Expect(vol.Name).To(Equal(KubeAPIAccessVolumeName))
+	g.Expect(vol.Projected).NotTo(BeNil())
+	g.Expect(*vol.Projected.DefaultMode).To(Equal(int32(0o444)))
```
stuggi: same

stuggi (approved):
/lgtm

PR #728, opened 2026-08-06
Add helpers for removing kolla

stuggi:
#### Add modules/users UID/GID registry
    
Single source of truth for all container image user identities, matching TCIB's _SUPPORTED_USERS in uid_gid_manage.sh. Go operators import constants directly (e.g. serviceuser.KeystoneUID); non-Go consumers use zz_generated_uid_gid.yaml (regenerated by make generate-uids-gids).
    
Includes all 42 users from TCIB:
    - OpenStack services: aodh, barbican, ceilometer, cinder, cloudkitty, cyborg, designate, glance, heat, horizon, ironic, ironic-inspector, keystone, manila, neutron, nova, octavia, placement, swift, watcher
    - Infrastructure: memcached, mysql, rabbitmq, redis, valkey, haproxy
    - System/auxiliary: apache (48), qemu (107), tss (59), ansible (227), kolla (for now kept), libvirt, nfast, openvswitch, hugetlbfs
    - Networking: frr, frrvty, ovn-bgp
    - Testing: horizontest, rally, tempest, tobiko
    - Admin: cloud-admin
    
Groups match TCIB: barbican→nfast, nova→libvirt/qemu/tss, frr→frrvty, keystone→apache. Home directories match TCIB (memcached→/run/memcache, redis→/run/redis, etc.).
    
valkey (42460) intentionally shares redis's UID (drop-in replacement).

#### Add SecurityContext helpers for kolla removal
    
Pod and container-level SecurityContext builders that enforce the hardened baseline every operator adopts after removing kolla:
    
    - RestrictivePodSecurityContext(uid, supplementalGroups...): pod-level RunAsUser/RunAsGroup/FSGroup=uid, RunAsNonRoot=true, seccomp RuntimeDefault, optional supplemental groups (e.g. ApacheGID)
    - RestrictiveSecurityContext(uid, addCapabilities...): container-level RunAsUser/RunAsGroup=uid, RunAsNonRoot=true, drop ALL capabilities (then add back any explicitly requested), no privilege escalation, seccomp RuntimeDefault
    - RestrictiveSecurityContextWithGID(uid, gid, addCapabilities...): same but allows a different RunAsGroup
    - PrivilegedSecurityContext(uid): for workloads needing full host device access (Ceph RBD via os-brick in glance/cinder)

#### Add modules/common/volume package
    
Generic volume helpers for writable emptyDir paths needed by all service operators after the kolla entrypoint removal:

    - WritableDirVolume(name, sizeLimit...): creates an emptyDir Volume
    - WritableDirVolumeMount(name, mountPath): creates a VolumeMount
    - WritableDirSubPathMounts(name, baseDir, subdirs...): SubPath mounts
      for writable subdirectories of a base path (e.g. service home dir)
    - ConfigSecretVolumes(secretNames): config Secrets at
      /var/lib/config-data/secret-{idx} with DefaultMode 0440
    
Well-known constants for standardized volume names and mount paths:

      RunHttpdVolumeName/MountPath   — /run/httpd (httpd PID file)
      VarLogHttpdVolumeName/MountPath — /var/log/httpd (httpd log fallback)
      TmpVolumeName/MountPath        — /tmp
      HomeDirCacheSubdir             — .cache (RHEL pkg_resources cache)
    
Replaces the per-operator copy-pasted GetRunHttpdVolume, GetVarLogHttpdVolume, GetLogVolume, GetConfigSecretVolumes helpers with one generic pair + constants.


modules/users/registry.go
```diff
@@ -0,0 +1,178 @@
...
+)
+
+// Registry maps service names to their user definitions. This is the
+// source of truth — the YAML generator (gen_uid_gid_yaml.go) marshals
+// this to zz_generated_uid_gid.yaml for container image builds.
+// Read-only at runtime; do not modify.
+var Registry = map[string]ServiceUser{
+	"ansible":          {UID: AnsibleUID, GID: AnsibleGID, Home: "/var/lib/ansible", Groups: []string{"kolla"}},
+	"aodh":             {UID: AodhUID, GID: AodhGID, Home: "/var/lib/aodh", Groups: []string{"kolla"}},
+	"barbican":         {UID: BarbicanUID, GID: BarbicanGID, Home: "/var/lib/barbican", Groups: []string{"kolla", "nfast"}},
```
abays, line 138: Is an `apache` entry missing here?

modules/common/volume/volume.go
```diff
@@ -0,0 +1,115 @@
...
+	TmpMountPath = "/tmp"
+	// HomeDirCacheSubdir is the ".cache" subdirectory under a service's
+	// home directory. RHEL's python3-setuptools downstream patch caches
+	// iter_entry_points() scans under $HOME/.cache/python-entrypoints/
+	// on every process start. Needs a writable emptyDir mount when
+	// ReadOnlyRootFilesystem is enabled.
+	HomeDirCacheSubdir = ".cache"
+)
+
+var configSecretMode int32 = 0440
```
abays: Is this going to eventually be used by more than one function?  Just curious, because otherwise it seems like it could live inside the `ConfigSecretVolumes` func.

modules/common/pod/security.go
```diff
@@ -43,11 +41,54 @@ func RestrictiveSecurityContextWithGID(uid, gid int64, addCapabilities ...corev1
...
+// ReadOnlyRootFilesystem and capability dropping. RunAsUser is set to uid,
+// RunAsGroup to gid.
+func PrivilegedSecurityContext(uid, gid int64) *corev1.SecurityContext {
+	return &corev1.SecurityContext{
+		RunAsUser:    ptr.To(uid),
+		RunAsGroup:   ptr.To(gid),
+		RunAsNonRoot: ptr.To(true),
+		Privileged:   ptr.To(true),
+	}
+}
```
abays, line 97: Claude is saying that `LVM/iSCSI/multipath` images frequently ship with `USER root`.  "Those images will be rejected at admission because `RunAsNonRoot=true` fails when the effective `UID` is 0."  Not sure if that's a real concern or not.

modules/common/pod/security.go
```diff
@@ -22,17 +22,15 @@ import (
...
 // Optional addCapabilities are added back after dropping ALL.
-func RestrictiveSecurityContext(uid int64, addCapabilities ...corev1.Capability) *corev1.SecurityContext {
-	return RestrictiveSecurityContextWithGID(uid, uid, addCapabilities...)
-}
-
-// RestrictiveSecurityContextWithGID is like RestrictiveSecurityContext but
-// allows specifying a different GID.
-func RestrictiveSecurityContextWithGID(uid, gid int64, addCapabilities ...corev1.Capability) *corev1.SecurityContext {
+//
+// Does not set ReadOnlyRootFilesystem -- a future RestrictiveReadOnlySecurityContext
```
abays, line 30: Is there a reason we're not adding `RestrictiveReadOnlySecurityContext` func now?  Just wonder if anything exists that needs `ReadOnlyRootFilesystem` already.

modules/common/volume/volume.go
```diff
@@ -0,0 +1,115 @@
...
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName:  secretName,
+					DefaultMode: &configSecretMode,
+				},
+			},
+		})
+		mounts = append(mounts, corev1.VolumeMount{
+			Name:      secretName,
+			MountPath: "/var/lib/config-data/secret-" + strconv.Itoa(idx),
```
abays, line 108: Could using `idx` here cause a problem if somehow `secretNames` was reordered between calls?  Or do we assume the slice is always in the same order?

modules/common/pod/security.go
stuggi: I was testing ReadOnlyRootFilesystem and it needs more work to identify at which places we then need empty vol mounts so that the services work. Thats why I have not done it as part of this work as it would be a to big of a change. I think we should add it when we really consider using ReadOnlyRootFilesystem for the service deployments

modules/common/pod/security.go
stuggi: ack, right now I don't think we have this case, but  I think we could set `RunAsNonRoot` to `false if a consumer passes `0` as the `uid` so its generic and can also be used with `UID` 0

modules/common/volume/volume.go
stuggi: not sure and not today, I will inline it

modules/common/volume/volume.go
stuggi: I think this could only happen if the caller reoders things, like when one changed the ordering in ConfigSecretVolumes. this would change the pod template and a rolling restart happens, but I think thats expected and ok here in this func. I think its the callers response to order the secretNames if they come from a source like map[string]string

modules/users/registry.go
stuggi: it was not in TCIB, but since I also added `named` I think its more complete if I also add `apache`.

stuggi: done

modules/common/volume/volume.go
stuggi: done

modules/common/pod/security.go
stuggi: done

modules/common/pod/security.go
abays: Cool, makes sense

abays (approved):
/lgtm

PR #729, opened 2026-08-06
Remove unused volume types from extraMounts VolumeSource

fmount:
Remove `HostPath`, `ISCSI`, `DownwardAPI`, `FC`, `ScaleIO`, `StorageOS` and `CephFS` from the `VolumeSource` struct. None of these are used via `extraMounts` by any operator.


stuggi (approved):
/lgtm

PR #733, opened 2026-08-13
Add IsReadyForInput to statefulset and deployment packages

lmiccini:
Each package now exposes IsReadyForInput, which reads the workload
    directly from a client.Reader (typically an uncached API reader) and
    confirms that the rollout is complete AND the pod template carries a
    CONFIG_HASH environment variable matching the expected value.
    
    This lets child controllers verify their StatefulSet/Deployment is
    truly running the desired configuration before reporting readiness,
    eliminating the informer cache race where stale status could cause a
    parent to release credential finalizers prematurely.

### DIFF
--- a/modules/common/deployment/deployment.go
+++ b/modules/common/deployment/deployment.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"time"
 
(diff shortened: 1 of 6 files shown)


stuggi (approved):
/lgtm

PR #746, opened 2026-09-09
Add Aetos UID and GID to registry

vyzigold:
Add the 42404 as Aetos UID/GID to the user registry. Same as https://github.com/openstack-k8s-operators/s2i-openstack-containers/blob/main/containers/base/scripts/uid_gid_manage#L36C34-L36C39

### DIFF
--- a/modules/users/registry.go
+++ b/modules/users/registry.go
@@ -37,6 +37,8 @@ type ServiceUser struct {
 // UIDs on PersistentVolumes. Changing a UID would require a manual chown
 // migration on every PV in every deployment.
 const (
+	AetosUID           int64 = 42404
+	AetosGID           int64 = 42404
 	AnsibleUID         int64 = 227
 	AnsibleGID         int64 = 227
 	AodhUID            int64 = 42402


abays (approved):
/lgtm

modules/users/registry.go
```diff
@@ -37,6 +37,8 @@ type ServiceUser struct {
 // UIDs on PersistentVolumes. Changing a UID would require a manual chown
 // migration on every PV in every deployment.
 const (
+	AetosUID           int64 = 42404
```
stuggi, line 40: should there also be an equivalent in Registry bellow, followed by running `make generate-uids-gids verify-uids-gids` to update `modules/users/zz_generated_uid_gid.yaml`, which is planned to be the  single source for s2i images. its not integrated yet.

vyzigold: I am blind. You're right. I'm on it.

vyzigold: https://github.com/openstack-k8s-operators/lib-common/pull/747

PR #747, opened 2026-09-09
Finish adding Aetos to user registry

vyzigold:
Followup to https://github.com/openstack-k8s-operators/lib-common/pull/746 addressing stuggi's review.

### DIFF
--- a/modules/users/registry.go
+++ b/modules/users/registry.go
@@ -134,6 +134,7 @@ const (
 // this to zz_generated_uid_gid.yaml for container image builds.
 // Read-only at runtime; do not modify.
 var Registry = map[string]ServiceUser{
+	"aetos":            {UID: AetosUID, GID: AetosGID, Home: "/var/lib/aetos", Groups: []string{"kolla"}},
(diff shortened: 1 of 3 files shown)


modules/users/registry.go
```diff
@@ -134,6 +134,7 @@ const (
 // this to zz_generated_uid_gid.yaml for container image builds.
 // Read-only at runtime; do not modify.
 var Registry = map[string]ServiceUser{
+	"aetos":            {UID: AetosUID, GID: AetosGID, Home: "/var/lib/aetos", Groups: []string{"kolla"}},
```
stuggi, line 137: we'd have to update the tests for the new user https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/users/registry_test.go#L24

stuggi: damn coderabbit was faster :)

vyzigold: This has been the hardest "one line change" in a while :D . Thank you for the vigilant review, even if it was slower than coderabbit. Tests run successfully and coderabbit doesn't have any other suggestions, so I hope that's all.

stuggi: thanks, looks good!

stuggi (approved):
/lgtm

